PR #1, opened 2015-03-12
Transfer code to parent repository

eugenechertikhin:


README.md
```diff
@@ -1 +1,25 @@
...
+
+Exporter for MySQL server metrics http://prometheus.io/
+
+## Building and running
+
+    make
+    ./mysqld_exporter <flags>
+
+## Configuration
+The configuration is in JSON. An example with all possible options:
```
brian-brazil: I think it might be better to be configurable entirely via flags rather than having to mess with json, there's not many possible options.

mysqld_exporter.go
```diff
@@ -0,0 +1,191 @@
...
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	namespace = "mysql"
+)
+
+var (
+	listenAddress = flag.String("web.listen-address", ":9091", "Address to listen on for web interface and telemetry.")
```
brian-brazil: @juliusv what's the next free port number?

mysqld_exporter.go
```diff
@@ -0,0 +1,191 @@
...
+// return new empty exporter
+func NewMySQLExporter(config *Config) *Exporter {
+	return &Exporter{
+		config: *config,
+		up: prometheus.NewGauge(prometheus.GaugeOpts{
+			Namespace: namespace,
+			Name:      "up",
+			Help:      "Was the last scrape of mysqld successful.",
+		}),
+		totalScrapes: prometheus.NewCounter(prometheus.CounterOpts{
```
brian-brazil: mysql_up should cover this and the next one already.

I'd suggest a gauge for how long the scrape of mysql took in seconds

http://prometheus.io/docs/practices/instrumentation/#collectors

mysqld_exporter.go
```diff
@@ -0,0 +1,191 @@
...
+	e.up.Set(0)
+
+	db, err := sql.Open("mysql", e.config.Config["mysql_connection"])
+	if err != nil {
+		log.Printf("error open connection to db using %s", e.config.Config["mysql_connection"])
+		return
+	}
+	defer db.Close()
+
+	rows, err := db.Query("SHOW STATUS")
```
brian-brazil: I presume SHOW SLAVE STATUS etc. will be in future changes?

bjk-soundcloud: This needs to be `SHOW GLOBAL STATUS`.  The default[0] is for the connected session status only.

[0]: http://dev.mysql.com/doc/refman/5.6/en/show-status.html

mysqld_exporter.go
juliusv: 9104 please :)

mysqld_exporter.go
```diff
@@ -0,0 +1,191 @@
...
+	ch <- e.totalScrapes
+	ch <- e.errorScrapes
+	e.collectMetrics(ch)
+}
+
+func (e *Exporter) scrape(scrapes chan<- []string) {
+	defer close(scrapes)
+
+	e.totalScrapes.Inc()
+	e.up.Set(0)
```
matthiasr: I'm wondering if there is a better way to do this than to set it to 0 then 1 again. If a scrape were to happen in the meantime, wouldn't this flap?

mysqld_exporter.go
```diff
@@ -0,0 +1,191 @@
...
+			// not really important
+			// log.Printf("Error while parsing field value %s (%s): %v", row[0], row[1], err)
+			e.errorScrapes.Inc()
+			continue
+		}
+
+		if _, ok := e.metrics[name]; !ok {
+			e.metrics[name] = prometheus.NewGauge(prometheus.GaugeOpts{
+				Namespace: namespace,
+				Name:      name,
```
brian-brazil: You should lower case this

mysqld_exporter.go
```diff
@@ -0,0 +1,191 @@
...
+	return
+}
+
+func (e *Exporter) setMetrics(scrapes <-chan []string) {
+	for row := range scrapes {
+
+		name := row[0]
+		value, err := strconv.ParseInt(row[1], 10, 64)
+		if err != nil {
+			// not really important
```
brian-brazil: Some of the variables are strings per http://dev.mysql.com/doc/refman/5.1/en/server-status-variables.html
It might be best to blacklist those, in case any of them happen to look like a number

brian-brazil:
This generally looks good, it'll be great to get this in.

You should add an AUTHORS.md and CONTRIBUTING.md

bjk-soundcloud:
For reference, the ganglia[0] mysqld python module is a good example of what can/needs to be scraped.

[0]: https://github.com/ganglia/gmond_python_modules/blob/master/mysqld/python_modules/mysql.py

mysqld_exporter.go
brian-brazil: We should add a note then that this is for 5.0 onwards

bjk-soundcloud:
The MySQL documentation[0] has some good information on the variables in `SHOW GLOBAL STATUS`

[0]: http://dev.mysql.com/doc/refman/5.6/en/server-status-variables.html

mysqld_exporter.go
matthiasr: With 5.1 being EOL I think that's safe to assume.

NOTICE
```diff
@@ -0,0 +1,5 @@
+Exporter for mysql daemon
```
juliusv: For repos with the `prometheus` org we'd like to be consistent with NOTICE, AUTHORS.md, etc. files. Could you make this:

```
Exporter for MySQL daemon.
Copyright 2015 The Prometheus Authors
```

The best place for initial authorship attribution would be in the AUTHORS.md file, like in this: https://github.com/prometheus/node_exporter/blob/master/AUTHORS.md

Makefile
```diff
@@ -0,0 +1,17 @@
+# Copyright 2013 The Prometheus Authors
```
juliusv: 2015?

mysqld_exporter.go
```diff
@@ -0,0 +1,191 @@
...
+			Help:      "Was the last scrape of mysqld successful.",
+		}),
+		totalScrapes: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: namespace,
+			Name:      "exporter_total_scrapes",
+			Help:      "Current total mysqld scrapes.",
+		}),
+		errorScrapes: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: namespace,
+			Name:      "exporter_error_scrapes",
```
juliusv: To be closer to Prometheus's usual naming conventions, could you rename these metrics:

```
exporter_total_scrapes -> exporter_scrapes_total
exporter_error_scrapes -> exporter_scrape_errors_total
```

mysqld_exporter.go
bjk-soundcloud: 5.1 was in Debian Squeeze, 5.5 was in Wheezy.  Supporting older than 5.1 is not a target.

README.md
eugenechertikhin: IMHO this is bad idea show login/password in process list.
But if you think this is needed - I will.

juliusv: Hm, I'd have to agree for the credentials. Those should usually be in a config file or passed via an environment variable. @bjk-soundcloud @matthiasr any opinion on this?

mysqld_exporter.go
eugenechertikhin: fixed

README.md
brian-brazil: Yeah, password would be a problem then.

Can we make the JSON structure simpler? I think config can be removed

mysqld_exporter.go
brian-brazil: You'd be surprised how old software some people run, we should be clear about what we support in any case.

juliusv: :+1: about at least mentioning it

README.md
juliusv: Using env variables and then simply `os.Getenv(...)` could even be nicer. It doesn't require putting another file in the filesystem, finding it, etc. Your run script can simply set an env variable, no matter where the contents of that come from.

mysqld_exporter.go
eugenechertikhin: fixed

README.md
```diff
@@ -1 +1,25 @@
...
+```
+{
+   "config" : {
+      "mysql_connection" : "login:password@host/dbname"
+   }
+}
+```
+Name     | Description
```
brian-brazil: It's probably best to link off to https://github.com/go-sql-driver/mysql#dsn-data-source-name rather than duplicating their docs.

NOTICE
eugenechertikhin: done

mysqld_exporter.go
eugenechertikhin: yep. I plan todo this1

README.md
matthiasr: I agree, `config` is extraneous.

I think the configuration file is fine for now. It won't be a big deal to add other ways later if desired.

eugenechertikhin: I will drop config file and parse env vars

eugenechertikhin:
@brian-brazil show slave status may be added in future. we are don't use it in our environment, so do not implement it in first version

README.md
bjk-soundcloud: +1 to ENV that contains the DSN[0]

[0]: https://github.com/go-sql-driver/mysql#dsn-data-source-name

mysqld_exporter.go
```diff
@@ -0,0 +1,179 @@
...
+		log.Printf("error open connection to db")
+		e.errorScrapes.Inc()
+		e.duration.Set(float64(time.Now().UnixNano() - now))
+		return
+	}
+	defer db.Close()
+
+	rows, err := db.Query("SHOW STATUS")
+	if err != nil {
+		log.Printf("error running query on db")
```
juliusv: I'd also log the error message:

    log.Println("error running status query on database:", err)

mysqld_exporter.go
```diff
@@ -0,0 +1,179 @@
...
+		return
+	}
+	defer rows.Close()
+
+	var key, val []byte
+	for rows.Next() {
+		// get RawBytes from data
+		err = rows.Scan(&key, &val)
+		if err != nil {
+			log.Printf("error getting result set")
```
juliusv: dito here:

    log.Println("error getting result set:", err)

mysqld_exporter.go
```diff
@@ -0,0 +1,179 @@
...
+		var res []string = make([]string, 2)
+		res[0] = string(key)
+		res[1] = string(val)
+
+		scrapes <- res
+	}
+
+	e.duration.Set(float64(time.Now().UnixNano() - now))
+
+	return
```
juliusv: Unnecessary return.

mysqld_exporter.go
```diff
@@ -0,0 +1,179 @@
...
+func (e *Exporter) collectMetrics(metrics chan<- prometheus.Metric) {
+	for _, m := range e.metrics {
+		m.Collect(metrics)
+	}
+}
+
+func main() {
+	flag.Parse()
+
+	dsn := os.Getenv("DSN")
```
juliusv: Wondering if this should be `DATA_SOURCE_NAME` to be clearer to everyone. Opinions?

mysqld_exporter.go
```diff
@@ -0,0 +1,179 @@
...
+	if len(dsn) == 0 {
```
juliusv: It'd be more common to say `if dsn == ""` in Go for strings.

mysqld_exporter.go
juliusv: Guess this still needs to be changed to `SHOW GLOBAL STATUS`.

mysqld_exporter.go
eugenechertikhin: i'm discuss with my sysadmin and we think need return config file. its depend on extra queries, options for checking slave status and so on...

brian-brazil: I think we'd always try to fetch things like the slave status. Things should work without configuration as much as possible.

mysqld_exporter.go
eugenechertikhin: i'm googling best practice about this.. and assume len(x) more clear (for me). in any way imho this is doen't matter. will use if x == "" in future.

mysqld_exporter.go
eugenechertikhin: ok
if i need fetch some metrics from sql package or make some special queries for monitoring database application?

in oracle it will be absolutely required.

juliusv: Hmm, but this exporter is for MySQL only, right? I don't have any knowledge on MySQL options though, and whether it's ok to always fetch everything. @bjk-soundcloud will know a ton about it, but he's on vacation this week.

mysqld_exporter.go
juliusv: yeah, was more of a nit, doesn't really matter

mysqld_exporter.go
brian-brazil: It's a question as to whether it's instrumentation for the mysqld, or instrumentation for the data inside the mysqld. Care should be taken when mixing these notions, for the same reasons we avoid putting exporters into the node_exporter.

juliusv:
@eugenechertikhin Heya, what's the state on this? Are you interested in resolving the remaining bits, or should we take it from here?

mysqld_exporter.go
matthiasr: Depending on the server state, it will or will not return data, e.g. `show slave status;` is just empty on a non-slave. I agree that the exporter should just inspect the server and figure out what data is sensible to export, then export that. I.e. if it is a slave, export slave data, if not those time series don't exist. The same could be true for e.g. `performance_schema` which exists on some versions and not others or may have different formats.

eugenechertikhin:
Hey Julius,
I will resolve in 2 days, sorry for delay - its depend on my main work. thanks for understanding

juliusv:
@eugenechertikhin Cool, no problem.

mysqld_exporter.go
eugenechertikhin: fixed

mysqld_exporter.go
eugenechertikhin: ok

mysqld_exporter.go
eugenechertikhin: ok

mysqld_exporter.go
eugenechertikhin: true

mysqld_exporter.go
```diff
@@ -0,0 +1,217 @@
...
+}
+
+// return new empty exporter
+func NewMySQLExporter(dsn string) *Exporter {
+	return &Exporter{
+		dsn: dsn,
+		duration: prometheus.NewGauge(prometheus.GaugeOpts{
+			Namespace: namespace,
+			Name:      "exporter_last_scrape_duration_nanoseconds",
+			Help:      "Was the last scrape  duration.",
```
juliusv: Oopsie, this help string got broken in the process. I guess "Was the" -> "The"?

mysqld_exporter.go
```diff
@@ -0,0 +1,217 @@
...
+func (e *Exporter) scrape(scrapes chan<- []string) {
+	defer close(scrapes)
+
+	now := time.Now().UnixNano()
+
+	e.totalScrapes.Inc()
+
+	db, err := sql.Open("mysql", e.dsn)
+	if err != nil {
+		log.Printf("error open connection to database: ", err)
```
juliusv: s/open/opening/

README.md
```diff
@@ -1 +1,20 @@
...
+    ./mysqld_exporter <flags>
+
+Flags description
+
+Name               | Description
+-------------------|------------
+web.listen-address | Address to listen on for web interface and telemetry.
+web.telemetry-path | Path under which to expose metrics.
+
+Variable with database datasource must be set in DSN environment variable
```
juliusv: "DSN" is "DATA_SOURCE_NAME" now.

juliusv:
@matthiasr @bjk-soundcloud @brian-brazil If you guys think this does roughly what it should now, I think we should get it merged soon and do any further cleanups afterwards so that they can be done by other people as well.

bjk-soundcloud:
@juliusv I'm not sure how this handles the different outputs of SHOW GLOBAL STATUS. Values can be Int, Float, or String.

Otherwise, LGTM. :+1:

matthiasr:
:+1: let's get it rolling

mysqld_exporter.go
```diff
@@ -0,0 +1,217 @@
...
+	metrics                    map[string]prometheus.Gauge
+}
+
+// return new empty exporter
+func NewMySQLExporter(dsn string) *Exporter {
+	return &Exporter{
+		dsn: dsn,
+		duration: prometheus.NewGauge(prometheus.GaugeOpts{
+			Namespace: namespace,
+			Name:      "exporter_last_scrape_duration_nanoseconds",
```
brian-brazil: We have this as seconds in other exporters

mysqld_exporter.go
```diff
@@ -0,0 +1,217 @@
...
+			Help:      "The last scrape  duration.",
```
brian-brazil: There's an extra space after scrape

mysqld_exporter.go
brian-brazil: I mean that you need to lower case the string contained in name, as we usually use lowercase metric names in prometheus

mysqld_exporter.go
```diff
@@ -0,0 +1,217 @@
...
+		}),
+		totalScrapes: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: namespace,
+			Name:      "exporter_scrapes_total",
+			Help:      "Current total mysqld scrapes.",
+		}),
+		errorScrapes: prometheus.NewCounter(prometheus.CounterOpts{
```
brian-brazil: This sort of metric is difficult to work with, as you have to compare rates to figure out if the mysqld is working.

A gauge that is 1 or 0 depending on if the scrape suceeded is better

mysqld_exporter.go
eugenechertikhin: this is already done in setMetrics

name := strings.ToLower(row[0])

brian-brazil: Ah, missed that.

mysqld_exporter.go
eugenechertikhin: fixed

brian-brazil: That code doesn't look right, you want to make errorScrape a Gauge and use .Set() to set it to 0 or 1

juliusv:
Thanks! I'm merging this now.

eugenechertikhin:
thank you Julius!

PR #2, opened 2015-04-30
Added docker file

mlety2:


juliusv:
@discordianfish

discordianfish:
That Dockerfile requires the binary to exist already, so it can't be used with automated builds.
I'm not opposed to use binaries built outside Docker in general because that might reduce image size and we can reuse the same build process but until we have a process to build and distribute binaries, I would make the Dockerfiles build the binaries on their own.

@mlety2 Do you want to change the Dockerfile to use alpine linux and install + cleanup stuff in one RUN command like we do here? https://github.com/prometheus/prometheus/blob/master/Dockerfile

mlety2:
Ok sure. I originally had a script that would build the binary in a debian container. Would that play well with automated builds?

discordianfish:
@mlety2 Unfortunately not, the automated build service will basically just checkout the repo and run 'docker build', so everything needs to happen in the Dockerfile.

mlety2:
I added the build steps to the debian image. I couldn't get the make script to recognise go installed via the alpine package manager. I also couldn't get the go binary from https://golang.org/dl to run in alpine either, I kept getting `/bin/sh: /src/.build/go1.4.2/bin/go: not found` even though it was definitely there.

discordianfish:
Okay, I think this is good for now. Thanks! @juliusv LGTM

discordianfish:
@juliusv Can you add the docker webhook? Or give me permissions to do so. Or add the docker automated builds yourself :)

juliusv:
@discordianfish Added the "Docker" Service for now.

discordianfish:
FYI: The image is added to our docker hub account: https://registry.hub.docker.com/u/prom/mysqld-exporter/

PR #6, opened 2015-05-29
Fix slave status parsing

jfindley:
Fixes slave status parsing.  As it supports variable column widths, this should work on all 5.x versions of mysql/mariadb.  No promises about 4.x.

Also cleaned up log files - removing an extraneous space between description and error.

### DIFF
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -6,6 +6,7 @@ import (
 	_ "github.com/go-sql-driver/mysql"
 	"net/http"
 	"os"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync"
@@ -88,7 +89,7 @@ func (e *Exporter) scrape(scrapes chan<- []string) {
 
 	db, err := sql.Open("mysql", e.dsn)
 	if err != nil {
-		log.Printf("error opening connection to database: ", err)
+		log.Println("error opening connection to database:", err)
 		e.error.Set(1)
 		e.duration.Set(float64(time.Now().UnixNano()-now) / 1000000000)
 		return
@@ -98,7 +99,7 @@ func (e *Exporter) scrape(scrapes chan<- []string) {
 	// fetch database status
 	rows, err := db.Query("SHOW GLOBAL STATUS")
 	if err != nil {
-		log.Println("error running status query on database: ", err)
+		log.Println("error running status query on database:", err)
 		e.error.Set(1)
 		e.duration.Set(float64(time.Now().UnixNano()-now) / 1000000000)
 		return
@@ -106,11 +107,12 @@ func (e *Exporter) scrape(scrapes chan<- []string) {
 	defer rows.Close()
 
 	var key, val []byte
+
 	for rows.Next() {
 		// get RawBytes from data
 		err = rows.Scan(&key, &val)
 		if err != nil {
-			log.Printf("error getting result set: ", err)
+			log.Println("error getting result set:", err)
 			return
 		}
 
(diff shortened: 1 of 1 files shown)


mysqld_exporter.go
```diff
@@ -124,40 +125,66 @@ func (e *Exporter) scrape(scrapes chan<- []string) {
...
+	if err != nil {
+		log.Println("error retrieving column list:", err)
+		e.error.Set(1)
+		e.duration.Set(float64(time.Now().UnixNano()-now) / 1000000000)
+		return
+	}
+
+	var slaveData = make([]sql.RawBytes, len(slaveCols))
+
+	scanArgs := make([]interface{}, len(slaveCols))
```
brian-brazil: interface{} means any type (like Object in Java), so I think you should be able to use slaveData directly in the rows.Scan.

mysqld_exporter.go
```diff
@@ -124,40 +125,66 @@ func (e *Exporter) scrape(scrapes chan<- []string) {
...
-				res[1] = "0"
-			} else {
-				res[1] = string(val)
-			}
+
+			scrapes <- res
 		}
 
-		scrapes <- res
+		if col == "Seconds_Behind_Master" {
```
brian-brazil: The previous code tried to read all the values, there are other number values that may be of use so I'd suggest keeping with that unless you've got a reason not to

mysqld_exporter.go
jfindley: I know what interface{} means ;)
I guess this response means I should have added a comment here.  The reason I do this is because rows.Scan() needs a pointer to an object rather than the object itself.  &slaveData would give a pointer to the slice of byteslices, when what we need is for each element of slaveData to be a pointer.  Creating a slice of interfaces, with each element being a pointer to the underlying sql.NullByte is a simple and fairly standard way of doing this.

Does that clarify it?

mysqld_exporter.go
```diff
@@ -88,7 +88,7 @@ func (e *Exporter) scrape(scrapes chan<- []string) {
 
 	db, err := sql.Open("mysql", e.dsn)
 	if err != nil {
-		log.Printf("error opening connection to database: ", err)
+		log.Printf("error opening connection to database:", err)
```
juliusv: This should just be `log.Println`. Same for several other occurrences below.

mysqld_exporter.go
brian-brazil: Yes, would make([]*sql.RawBytes, len(slaveCols) do the same job then?

jfindley: Sadly not - []*sql.RawBytes doesn't satisfy []interface{}.

mysqld_exporter.go
jfindley: I skipped the others because they're generally either junk, strings that have no sensible integer conversion, or both.  I can put them in if you'd prefer, but they're almost entirely useless except for these three.

mysqld_exporter.go
jfindley: Thanks.  Fixed.

mysqld_exporter.go
brian-brazil: The Master_Log_Pos ones can be useful, and it doesn't hurt to have what we can integer wise.

matthiasr: +1 for master_log_pos + master_log_file (as good as possible – just strip non-integer bits?), but also +1 on discarding garbage.

mysqld_exporter.go
```diff
@@ -124,40 +126,85 @@ func (e *Exporter) scrape(scrapes chan<- []string) {
...
-			log.Printf("error getting result set: ", err)
+			log.Println("error retrieving result set:", err)
+			e.error.Set(1)
+			e.duration.Set(float64(time.Now().UnixNano()-now) / 1000000000)
 			return
 		}
 
+	}
+
+	threadRexp := regexp.MustCompile("^Slave_(.*)_Running$")
```
brian-brazil: You could combine all bar the log_file into one regexp, which'll reduce the copy&paste

discordianfish:
LGTM
@juliusv / @brian-brazil Can we merge this? It sits around for some time already.

brian-brazil:
:+1: Can you squash the commits?

jfindley:
Done. Thanks.

PR #10, opened 2015-07-13
Refactor metrics

SuperQ:
In order to make it easier to differentiate between `SHOW SLAVE STATUS`
and `SHOW GLOBAL STATUS` we need add a `Subsystem` to metric names.

Refactor `SHOW GLOBAL STATUS` pattern matching

This will make room in the metric namespace for MySQL metrics from the
`performance_schema` tables.

### DIFF
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -17,7 +17,9 @@ import (
 )
 
 const (
-	namespace = "mysql"
+	namespace    = "mysql"
+	slaveStatus  = "slave_status"
+	globalStatus = "global_status"
 )
 
 var (
@@ -59,21 +61,25 @@ func NewMySQLExporter(dsn string) *Exporter {
 		metrics: map[string]prometheus.Gauge{},
 		commands: prometheus.NewCounterVec(prometheus.CounterOpts{
 			Namespace: namespace,
+			Subsystem: globalStatus,
 			Name:      "commands_total",
 			Help:      "Number of executed mysql commands.",
 		}, []string{"command"}),
 		connectionErrors: prometheus.NewCounterVec(prometheus.CounterOpts{
 			Namespace: namespace,
+			Subsystem: globalStatus,
 			Name:      "connection_errors_total",
 			Help:      "Number of mysql connection errors.",
 		}, []string{"error"}),
 		innodbRows: prometheus.NewCounterVec(prometheus.CounterOpts{
 			Namespace: namespace,
+			Subsystem: globalStatus,
 			Name:      "innodb_rows_total",
 			Help:      "Mysql Innodb row operations.",
 		}, []string{"operation"}),
 		performanceSchema: prometheus.NewCounterVec(prometheus.CounterOpts{
 			Namespace: namespace,
+			Subsystem: globalStatus,
 			Name:      "performance_schema_total",
 			Help:      "Mysql instrumentations that could not be loaded or created due to memory constraints",
 		}, []string{"instrumentation"}),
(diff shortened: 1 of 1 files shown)


SuperQ:
@fabxc @matthiasr This will break all the existing metric names, but is necessary to avoid collisions between GLOBAL and SLAVE status variables.

fabxc:
:+1: To feel pretty one could move the source strings to constants.

SuperQ:
@fabxc Added `slave_status` and `global_status` constants.

mysqld_exporter.go
```diff
@@ -17,7 +17,9 @@ import (
 )
 
 const (
-	namespace = "mysql"
+	namespace     = "mysql"
+	slave_status  = "slave_status"
```
fabxc: camelCase in Go.

SuperQ:
@fabxc I agree on the common prefix approach in general, but in this case it would be confusing from the MySQL perspective.  For example the next PR will be called "performance" to represent the `performance_schema` subsystem.

fabxc:
I see. LGTM :+1:

SuperQ:
@beorn7 PTAL

beorn7:
Looking...

mysqld_exporter.go
```diff
@@ -225,36 +241,47 @@ func (e *Exporter) setGenericMetric(name string, value float64) {
...
+		e.commands.With(prometheus.Labels{"command": command[1]}).Set(value)
+	case len(connectionError) > 0:
+		e.connectionErrors.With(prometheus.Labels{"error": connectionError[1]}).Set(value)
+	case len(innodbRow) > 0:
+		e.innodbRows.With(prometheus.Labels{"operation": innodbRow[1]}).Set(value)
+	case len(performanceSchema) > 0:
+		e.performanceSchema.With(prometheus.Labels{"instrumentation": performanceSchema[1]}).Set(value)
+	default:
+		e.setGenericMetric(globalStatus, name, value)
+	}
```
beorn7: Flow is a bit clunky here. Also, all regexp's are evaluation although we only need one hit.

How about the following:

if command := comRegex.FindStringSubmatch(name); command != nil {
    e.commands.With(prometheus.Labels{"command": command[1]}).Set(value)
    return
}
if //...
// and finally:
e.setGenericMetric(globalStatus, name, value)

SuperQ: Yea, I didn't write this part.  I can refactor it as well.

SuperQ:
@beorn7 Ok, I refactored the regexp, much better now.

mysqld_exporter.go
```diff
@@ -225,36 +241,42 @@ func (e *Exporter) setGenericMetric(name string, value float64) {
 	e.metrics[name].Set(value)
 }
 
-func (e *Exporter) setMetrics(scrapes <-chan []string) {
-	var comRegex = regexp.MustCompile(`^com_(.*)$`)
-	var connectionErrorRegex = regexp.MustCompile(`^connection_errors_(.*)$`)
-	var innodbRowRegex = regexp.MustCompile(`^innodb_rows_(.*)$`)
-	var performanceSchemaRegex = regexp.MustCompile(`^performance_schema_(.*)$`)
+var (
+	globalStatusRegexp = regexp.MustCompile(`^(com|connection_errors|innodb_rows|performance_schema)_(.*)$`)
```
beorn7: Since this is a single var now, you should just write

    var globalStatusRegexp = regexp...

beorn7:
:+1: otherwise.

PR #12, opened 2015-07-22
Add support for performance_schema

SuperQ:
Add support for getting table metrics from MySQL
`performance_schema.table_io_waits_summary_by_table`

### DIFF
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -9,7 +9,6 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
-	"sync"
 	"time"
 
 	_ "github.com/go-sql-driver/mysql"
@@ -25,6 +24,10 @@ var (
 	metricPath = flag.String(
 		"web.telemetry-path", "/metrics",
 		"Path under which to expose metrics.",
+	)
+	perfTableIOWaits = flag.Bool(
+		"collect.perf_schema.tableiowaits", true,
+		"Collect metrics from performance_schema.table_io_waits_summary_by_table",
 	)
 )
 
@@ -75,7 +78,7 @@ var (
 	performanceSchemaTableWaitsDesc = prometheus.NewDesc(
 		prometheus.BuildFQName(namespace, performanceSchema, "table_io_waits_total"),
 		"The total number of table I/O wait events for each table and operation.",
-		[]string{"schema", "name", "operations"}, nil,
+		[]string{"schema", "name", "operation"}, nil,
 	)
 )
 
@@ -87,7 +90,6 @@ var (
 
 // Exporter collects MySQL metrics. It implements prometheus.Collector.
 type Exporter struct {
-	mutex           sync.Mutex
 	dsn             string
 	duration, error prometheus.Gauge
 	totalScrapes    prometheus.Counter
@@ -123,7 +125,7 @@ func (e *Exporter) Describe(ch chan<- *prometheus.Desc) {
 	// We cannot know in advance what metrics the exporter will generate
 	// from MySQL. So we use the poor man's describe method: Run a collect
 	// and send the descriptors of all the collected metrics. The problem
-	// here is that we need to connecte to the MySQL DB. If it is currently
+	// here is that we need to connect to the MySQL DB. If it is currently
 	// unavailable, the descriptors will be incomplete. Since this is a
 	// stand-alone exporter and not used as a library within other code
 	// implementing additional metrics, the worst that can happen is that we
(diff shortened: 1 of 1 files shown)


beorn7:
What's your suggested course of action now?

- Get this submitted ASAP and refactor it to a saner state later?
- Go through a detailed review now to improve things?
- Let me write additional commits to straighten thing (also the legacy ones) up with this PR?

SuperQ:
So I think some refactoring needs to happen in-place with regards to the labels.  I tested the code, it does not work as-is.

beorn7:
I'll have a closer look ASAP. It sounds we should first clean up the code because that will make debugging easier.

beorn7:
@SuperQ The latest commit brings things in line with sanity. However, I haven't even tried it out, and obviously, we should have a test reading from some example input. Unfortunately, I'm out of time. Perhaps I'll find some time during my vacation, but unlikely... You might want to try it out in the meantime...

mysqld_exporter.go
```diff
@@ -12,297 +12,326 @@ import (
...
-	}
-
-	e.commands.Describe(ch)
-	e.connectionErrors.Describe(ch)
-	e.innodbRows.Describe(ch)
-	e.performanceSchema.Describe(ch)
+	// We cannot know in advance what metrics the exporter will generate
+	// from MySQL. So we use the poor man's describe method: Run a collect
+	// and send the descriptors of all the collected metrics. The problem
+	// here is that we need to connecte to the MySQL DB. If it is currently
```
juliusv: s/connecte/connect/

mysqld_exporter.go
```diff
@@ -12,297 +12,326 @@ import (
...
-	dsn               string
-	mutex             sync.RWMutex
-	duration, error   prometheus.Gauge
-	totalScrapes      prometheus.Counter
-	metrics           map[string]prometheus.Gauge
-	commands          *prometheus.CounterVec
-	connectionErrors  *prometheus.CounterVec
-	innodbRows        *prometheus.CounterVec
-	performanceSchema *prometheus.CounterVec
+	mutex           sync.Mutex
```
grobie: Unused, can go away.

SuperQ:
@beorn7 I fixed a race condition with the help of @juliusv.  I tested it against a real mysqld server and got good results.

SuperQ:
Added a bool flag to make this optional

SuperQ:
Removed [WIP], this is ready for final review. :) :) :)

mysqld_exporter.go
```diff
@@ -1,311 +1,347 @@
...
+	// exported metrics during the runtime of the exporter.
+
+	metricCh := make(chan prometheus.Metric)
+	doneCh := make(chan struct{})
+
+	go func() {
+		for m := range metricCh {
+			ch <- m.Desc()
+		}
+		doneCh <- struct{}{}
```
fabxc: This works just fine but `close(doneCh)` is the more idiomatic solution.

mysqld_exporter.go
```diff
@@ -1,311 +1,347 @@
...
+var (
+	listenAddress = flag.String(
+		"web.listen-address", ":9104",
+		"Address to listen on for web interface and telemetry.",
+	)
+	metricPath = flag.String(
+		"web.telemetry-path", "/metrics",
+		"Path under which to expose metrics.",
+	)
+	perfTableIoWaits = flag.Bool(
```
fabxc: `perfTableIOWaits`

fabxc:
Looks good aside from nit comments :+1:

SuperQ:
@juliusv Any additional issues?

mysqld_exporter.go
```diff
@@ -1,311 +1,347 @@
...
+		}
+		doneCh <- struct{}{}
+	}()
 
-	ch <- e.duration.Desc()
-	ch <- e.totalScrapes.Desc()
-	ch <- e.error.Desc()
+	e.Collect(metricCh)
+	close(metricCh)
+	close(doneCh)
```
juliusv: I think you've misunerstood Fabian here. Line 143 should be `close(doneCh)`, while this line here should still be `<-doneCh`.

juliusv:
The rest looks roughly ok from a brief scan.

juliusv:
:+1: on the channel fix

fabxc:
:shipit: :ship:

beorn7:
I had a few things written for this while on vacation (especially for testability) but never pushed it because of my crappy connectivity.

I'll create a separate PR if it's worth it...

PR #13, opened 2015-08-13
Added HTTP basic auth support.

roman-vynar:
This code adds the support for HTTP basic auth similar to https://github.com/prometheus/node_exporter
It will allow to protect the metrics.

Prometheus config excerpt:
```
...
  - job_name: dbservers

    basic_auth:
       username: prometheus
       password: abc123
...
```

In this case, the command to run mysqld_exporter will look like:
```
DATA_SOURCE_NAME="root:root@(localhost:3306)/mysql" ./mysqld_exporter \
-web.listen-address=":9101" -auth.user=prometheus -auth.pass=abc123 > /dev/null 2>&1 &
```

Thanks for this awesome Prometheus exporter.

### DIFF
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -25,7 +25,26 @@ const (
 var (
 	listenAddress = flag.String("web.listen-address", ":9104", "Address to listen on for web interface and telemetry.")
 	metricPath    = flag.String("web.telemetry-path", "/metrics", "Path under which to expose metrics.")
+	authUser      = flag.String("auth.user", "", "Username for basic auth.")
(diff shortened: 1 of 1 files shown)


brian-brazil:
We believe that any authentication/authorization for the web frontends should be done in a reverse proxy. There's just too many possible configurations (SSL alone has 20+ options that we'd need to expose), and it's better than all be handled in something designed for such a role as it's not a essential component of a monitoring system. http://www.robustperception.io/adding-basic-auth-to-prometheus-with-nginx/ is one tutorial that explains how to do this.

See https://github.com/prometheus/prometheus/pull/972 for background.

PR #14, opened 2015-08-19
Add support for Percona userstats output.

wrouesnel:
This commit adds support for dumping Percona userstats. These metrics are per
user counters are are exported as such to enable detection of resets by
Prometheus.

The new functionality, being limited to Percona, is enabled by invoking with
-collector.userstats=true

### DIFF
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"database/sql"
 	"flag"
+	"fmt"
 	"net/http"
 	"os"
 	"regexp"
@@ -45,6 +46,8 @@ var (
 		"collect.perf_schema.indexiowaitstime", true,
 		"Collect time metrics from performance_schema.table_io_waits_summary_by_index_usage",
 	)
+	userStat = flag.Bool("collect.info_schema.userstats", false,
+		"If running with userstat=1, set to true to collect user statistics")
 )
 
 // Metric name parts.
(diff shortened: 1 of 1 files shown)


SuperQ:
Cool, this should be useful.  FYI we're doing some major refactoring of the exporter right now, see https://github.com/prometheus/mysqld_exporter/pull/12

SuperQ:
Ok, https://github.com/prometheus/mysqld_exporter/pull/12 is merged, please rebase.

wrouesnel:
Rebased and refactored to match the performance schema approach. Tested against a server with 8000 user accounts.

mysqld_exporter.go
```diff
@@ -340,6 +348,139 @@ func (e *Exporter) scrape(ch chan<- prometheus.Metric) {
 			)
 		}
 	}
+
+	if *userStat {
+		informationSchemaUserStatisticsRows, err := db.Query("SELECT * FROM information_schema.USER_STATISTICS")
```
SuperQ: I would prefer to have the columns listed instead of `*`.  This avoids ordering or feature changes between MySQL versions.

mysqld_exporter.go
```diff
@@ -80,6 +83,11 @@ var (
 		"The total number of table I/O wait events for each table and operation.",
 		[]string{"schema", "name", "operation"}, nil,
 	)
+	informationSchemaUserStatisticsDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, informationSchema, "user_statistics_total"),
+		"Per-user metrics counters for each database user",
+		[]string{"user", "counter"}, nil,
```
SuperQ: I'm not sure having the counter type as a label is correct use of prometheus metrics in this case.  Labels are meant to be aggregated over, like "all users".  It doesn't make sense to aggregate over `TOTAL_CONNECTIONS` and `BYTES_SENT`.  These need to be split out into separate metrics. 

Another good example is you can't have one metric name be both a counter and a gauge.  If `concurrent_connections` was enabled below, this would be invalid:

`mysql_info_schema_user_statistics_total{user="foo",counter="concurrent_connections"}` as a gauge 
and `mysql_info_schema_user_statistics_total{user="foo",counter="total_connections"}` as a counter.

wrouesnel:
It was originally done differently, but the changes to the layout of the exporter seemed to imply this would no longer fit. I'm going to refactor it to look more like my original code, since that approach wasn't as clean, but was more adaptable and kept the label names correctly.

SuperQ:
Yea, it's a lot of copy-n-paste code.  Maybe take a look at the way it's still doing `SHOW SLAVE STATUS`.  In looking over the way we're recording some of the io_waits, I'm possibly going to change that a bit as well.  There are two types of metrics for those, one for reads and writes, the other for SQL operations (fetch, insert, update, delete).

Thanks for adding this userstat support, should be very useful.  With this and some performance_schema, I'm probably going to make a major release, maybe a "1.0" for mysqld_exporter in mid September.

wrouesnel:
The thing I'm stuck on is whether it makes sense for the exporter to be breaking things down into labelled metric sets or not. Something like "bytes_sent" and "bytes_received" can be merged fairly logically, but "total_connections" necessarily stands separate from "denied_connections" and "lost_connections" if we don't want to fiddle with the data to make the labels make sense. Which is problematic because if the fields change meaning, we could be ignoring one we need to make everything add up.

SuperQ:
Normally for something like `bytes_sent` and `bytes_received` I would make them separate metrics.  It's not common to total them, and it's easy to combine multiple metrics streams in the rule language.

matthiasr:
Having total/lost/denied as separate metrics is also idiomatic in
Prometheus, because it allows easy calculation of e.g loss rates by simply
dividing; don't force it into labels.

Storage wise there is no difference anyway; two metrics is the same as two
label values in disk. The difference is purely logical in collection and
querying.

wrouesnel:
Missed a gauge value for concurrent connections, fixed now.

SuperQ:
Ok, looks much better now.  :+1: for general idea.  I'll leave it to others to comment on Go code nit-picking.

SuperQ:
@wrouesnel Hey, can you merge upstream one more time?  I'd like to get this review completed and merged.

mysqld_exporter.go
```diff
@@ -80,6 +84,34 @@ var (
 		"The total number of table I/O wait events for each table and operation.",
 		[]string{"schema", "name", "operation"}, nil,
 	)
+	// Map known user-statistics values to types. Unknown types will be mapped as
+	// untyped.
+	informationSchemaUserStatisticsTypes = map[string]struct {
+		vtype prometheus.ValueType
+		desc  *prometheus.Desc
+	}{
+		"TOTAL_CONNECTIONS":      {prometheus.CounterValue, prometheus.NewDesc(prometheus.BuildFQName(namespace, informationSchema, "user_statistics_total_connections"), "The number of connections created for this user.", []string{"user"}, nil)},
```
SuperQ: Maybe multi-line these, would make diffs easier to read.

wrouesnel:
Rebased and tested.

I agree the the statistics map is a little dense, but it's all semi-automatically generated code anyway that shouldn't change unless a field gets added. IMO spacing it out at the moment will just add clutter.

mysqld_exporter.go
fabxc: Yes, please.

mysqld_exporter.go
```diff
@@ -601,6 +632,55 @@ func (e *Exporter) scrape(ch chan<- prometheus.Metric) {
...
+			return
+		}
+
+		var user string // Holds the username, which should be in column 0
+		var userStatData = make([]float64, len(columnNames))
+		var userStatScanArgs = make([]interface{}, len(columnNames))
+		userStatScanArgs[0] = &user
+		for i := range userStatData[1:] {
+			userStatScanArgs[i+1] = &userStatData[i+1]
+		}
```
fabxc: I would add a comment above this block and explain what's happening  with those slices here and below.

wrouesnel:
Ok so I've spaced out the mapping table for stats and added a comment explaining what's going on with it's use (good idea too since while reading it I had a moment where I thought the map was wrong).

SuperQ:
@wrouesnel Awesome, I'll take a look with people in the morning and get it merged.

fabxc:
:+1:

PR #15, opened 2015-08-21
Add new metrics

SuperQ:
Add more metrics from performance_schema:
* Time metrics for table_io_waits_summary_by_table
* Counter metrics for table_io_waits_summary_by_index_usage
* Time metrics for table_io_waits_summary_by_index_usage

Time metrics are returned by MySQL as `BIGINT` picoseconds, convert to
Float64 seconds.

Index usage columns are special[0].  Reads are counted for all indexs.
Writes are counted only for the `INDEX_NAME` with a value of `NULL`.

Bug fix: cleanup output column vars in perfSchemaTableWaitsRows.

[0]: http://dev.mysql.com/doc/refman/5.6/en/table-waits-summary-tables.html#table-io-waits-summary-by-index-usage-table

### DIFF
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -29,6 +29,18 @@ var (
 		"collect.perf_schema.tableiowaits", true,
 		"Collect metrics from performance_schema.table_io_waits_summary_by_table",
 	)
+	perfTableIOWaitsTime = flag.Bool(
+		"collect.perf_schema.tableiowaitstime", true,
+		"Collect time metrics from performance_schema.table_io_waits_summary_by_table",
+	)
+	perfIndexIOWaits = flag.Bool(
+		"collect.perf_schema.indexiowaits", true,
+		"Collect metrics from performance_schema.table_io_waits_summary_by_index_usage",
+	)
+	perfIndexIOWaitsTime = flag.Bool(
+		"collect.perf_schema.indexiowaitstime", true,
+		"Collect time metrics from performance_schema.table_io_waits_summary_by_index_usage",
+	)
 )
 
 // Metric name parts.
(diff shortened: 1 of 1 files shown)


mysqld_exporter.go
```diff
@@ -80,6 +92,21 @@ var (
 		"The total number of table I/O wait events for each table and operation.",
 		[]string{"schema", "name", "operation"}, nil,
 	)
+	performanceSchemaTableWaitsTimeDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, performanceSchema, "table_io_waits_picoseconds_total"),
+		"The total time of table I/O wait events for each table and operation.",
+		[]string{"schema", "name", "operation"}, nil,
+	)
+	performanceSchemaIndexWaitsDesc = prometheus.NewDesc(
```
brian-brazil: This and below form a summary. Are you sure you want to add a 3rd time unit to this exporter?

SuperQ: Yea, mysql is inconsistent in things.  Also chatting with @fabxc and @juliusv, I think I will convert all the time data to seconds if possible

SuperQ:
Ok, I did a bunch of cleanup and bug fixes.  This is ready for review. @fabxc :) :)

mysqld_exporter.go
```diff
@@ -340,6 +357,148 @@ func (e *Exporter) scrape(ch chan<- prometheus.Metric) {
 			)
 		}
 	}
+
+	if *perfTableIOWaitsTime {
+		// Timers here are returned in picoseconds
```
fabxc: Terminate comments with `.`

mysqld_exporter.go
```diff
@@ -340,6 +357,148 @@ func (e *Exporter) scrape(ch chan<- prometheus.Metric) {
...
+			countInsert  int64
+			countUpdate  int64
+			countDelete  int64
+		)
+
+		for perfSchemaIndexWaitsRows.Next() {
+			if err := perfSchemaIndexWaitsRows.Scan(
+				&objectSchema, &objectName, &indexName, &countFetch, &countInsert, &countUpdate, &countDelete,
+			); err != nil {
+				log.Println("error getting result set:", err)
```
fabxc: I assume returning here is fine instead of trying the next row, which will most likely fail in the same way.
However, we have to `e.error.Set(1)` here, right?

mysqld_exporter.go
```diff
@@ -340,6 +357,148 @@ func (e *Exporter) scrape(ch chan<- prometheus.Metric) {
...
+			timeInsert   int64
+			timeUpdate   int64
+			timeDelete   int64
+		)
+
+		for perfSchemaIndexWaitsTimeRows.Next() {
+			if err := perfSchemaIndexWaitsTimeRows.Scan(
+				&objectSchema, &objectName, &indexName, &timeFetch, &timeInsert, &timeUpdate, &timeDelete,
+			); err != nil {
+				log.Println("error getting result set:", err)
```
fabxc: Same as above.

mysqld_exporter.go
SuperQ: This is copy-n-paste of @beorn7's method above.

fabxc: We are returning and logging an error before – so this is pretty clear a failed scrape to me.
Chances are it was just forgotten.

SuperQ: Ok, added error logging here.

fabxc:
:+1:

PR #20, opened 2015-08-25
Add auto_increment column metrics

SuperQ:
Add metrics for reporting the current value of auto_increment columns

Based on the reporting query:
https://github.com/SuperQ/bk-tools/blob/master/pk-full-ratio.sql

### DIFF
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -25,6 +25,10 @@ var (
 		"web.telemetry-path", "/metrics",
 		"Path under which to expose metrics.",
 	)
+	autoIncrementColumns = flag.Bool(
+		"collect.auto_increment.columns", true,
+		"Collect auto_increment columns and max values from information_schema",
+	)
 	perfTableIOWaits = flag.Bool(
 		"collect.perf_schema.tableiowaits", true,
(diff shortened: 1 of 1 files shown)


SuperQ:
@matthiasr This adds the current value and max metrics for `auto_increment` columns.

SuperQ:
@ebartels This will let us use native prometheus reporting, and the new magic "predict" functionality to just give us an estimated date in the future when things will fail.

mysqld_exporter.go
```diff
@@ -49,9 +53,27 @@ const (
...
 	globalStatus      = "global_status"
+	informationSchema = "info_schema"
 	performanceSchema = "perf_schema"
+	slaveStatus       = "slave_status"
+)
+
+// Metric SQL Queries.
+const (
+	infoSchemaAutoIncrementQuery = `
+SELECT table_schema, table_name, column_name, auto_increment,
```
juliusv: Why not indent one tab more than the `infoSchemaAutoIncrementQuery`? Makes indentation levels visually clearer.

juliusv:
:+1: otherwise on the Go side, can't judge the MySQLy bits too much.

ebartels:
:+1:

PR #21, opened 2015-08-25
Add SHOW GLOBAL VARIABLES metris

SuperQ:
SHOW GLOBAL VARIABLES provides gauge-like views of various MySQL
internal tunable settings.  These can be used for expressions that
compare current status (mysql_global_status_connections /
mysql_global_variables_max_connections)

### DIFF
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -54,6 +54,7 @@ const (
 	// Subsystems.
 	exporter          = "exporter"
 	globalStatus      = "global_status"
+	globalVariables   = "global_variables"
 	informationSchema = "info_schema"
 	performanceSchema = "perf_schema"
 	slaveStatus       = "slave_status"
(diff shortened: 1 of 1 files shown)


SuperQ:
@roman-vynar This will give us all the metrics for `SHOW GLOBAL VARIABLES`.

roman-vynar:
That's great :+1:

mysqld_exporter.go
```diff
@@ -302,6 +303,34 @@ func (e *Exporter) scrape(ch chan<- prometheus.Metric) {
...
+		e.error.Set(1)
+		return
+	}
+	defer globalVariablesRows.Close()
+
+	var globalVarKey string
+	var globalVarValue sql.RawBytes
+
+	for globalVariablesRows.Next() {
+		globalVarKey = strings.ToLower(globalVarKey)
```
matthiasr: Why is this before the `Scan`? Seems like a NOOP here.

matthiasr:
:+1: when the toLower is in the right spot

SuperQ:
@matthiasr Ahh, right, had it in the wrong place.

PR #22, opened 2015-08-25
Make global status key always lower case

SuperQ:
Avoid miss-matched string case.


SuperQ:
@matthiasr Another ToLower fix

matthiasr:
:+1:

PR #23, opened 2015-08-26
Support for monitoring crashed tables

wrouesnel:
Adds -collect.database.table_status which allows specifying a list of databases for which table_status should be scraped on every collection run. 

This option is far too expensive to use across even moderately sized database servers, but vital for monitoring "mission critical" tables.

Currently it just exports whether or not the table is crashed by inferring the NULL values in the status output.

Intended to be added atop the userstats pull.

### DIFF
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -56,6 +56,8 @@ var (
 	)
 	userStat = flag.Bool("collect.info_schema.userstats", false,
 		"If running with userstat=1, set to true to collect user statistics")
+	tableStatus = flag.String("collect.database.table_status", "",
+		"Comma (,) separated list of database names to collect table status for.")
 )
 
 // Metric name parts.
@@ -69,6 +71,7 @@ const (
 	informationSchema = "info_schema"
 	performanceSchema = "perf_schema"
 	slaveStatus       = "slave_status"
+	database		  = "database"
 )
 
 // Metric SQL Queries.
@@ -304,6 +307,11 @@ var (
 				"The number of times this user’s connections connected using SSL to the server.",
 				[]string{"user"}, nil)},
 	}
+	databaseTableStatusCrashedDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, database, "table_crashed_bool"),
+		"Indicates if the table is marked as crashed by MySQL",
+		[]string{"database", "table"}, nil,
+	)
 )
 
 // Various regexps.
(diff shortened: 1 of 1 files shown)


SuperQ:
I see you turned off a bunch of the new collectors.  What version are you having troubles with?  I don't have any older than MySQL 5.6 servers to test on anymore.  `collect.auto_increment.columns` should be safe enough to work all the way back to MySQL 5.1.  The others using performance_schema should probably be off.

I agree with the least surprise, but I want to provide the most functionality as possible while we're still very early in the development.

wrouesnel:
We're using 5.5 at the moment.

As I note in the commit, possibly the different collector blocks should be wrapped in func() closures so they can abort without blocking collectors below them?

SuperQ:
Yea, I think we could do the func() thing, or something similar.  I will discuss the options today.

SuperQ:
Ok, other changes have been merged, please rebase.

wrouesnel:
There's a second bug-fix commit in here. I can break it out into another module, but it looks like I must've been running an old version when testing some clean ups to userstats since it was pre-broken while I was rebasing. #24 0.9.0 will be broken till it's merged

mysqld_exporter.go
```diff
@@ -69,6 +71,7 @@ const (
 	informationSchema = "info_schema"
 	performanceSchema = "perf_schema"
 	slaveStatus       = "slave_status"
+	database		  = "database"
```
SuperQ: Please add this in alphabetical order.

SuperQ: Also, I think this should be `tableStatus = "table_status"` since the data comes from `SHOW TABLE STATUS`

SuperQ:
@wrouesnel Apparently the cost of running `SHOW TABLE STATUS` is greatly reduced with MySQL 5.6[0].

One of my example databases:
```
> SHOW TABLE STATUS FROM xxxxxxxxxx;
...
109 rows in set (0.00 sec)
```

[0]: https://dev.mysql.com/doc/refman/5.6/en/show-table-status.html

SuperQ:
@wrouesnel I think I would like to have this change use `information_schema.tables`[0] instead of `SHOW TABLE STATUS`.  This way you don't need to specify the list of databases if you don't want to.

I would do this similarly to the other implementations.

`SELECT * FROM information_schema.tables WHERE TABLE_SCHEMA NOT IN ('mysql', 'performance_schema', 'information_schema')`

While you're at it, would you mind adding the columns `TABLE_ROWS`, `DATA_LENGTH`, `INDEX_LENGTH`, and `DATA_FREE`.  This should give us some interesting approximate size information.

[0]: https://dev.mysql.com/doc/refman/5.1/en/tables-table.html

wrouesnel:
109 rows is pretty low. I deal with servers with closer to 1000+ where table status takes an age to complete. Hence the "database" notation since it's information about specific databases and not the server as a whole.

wrouesnel:
I had been entertaining the idea of having a fast-check and a slow-check parameter. So a go-routine in the background cycles through all the databases and tables lazily, and 'fast check' databases get examined every scrape.

wrouesnel:
For example running a full check on a server with 398 databases:
```bash
# time ( mysql -B -e 'show databases;' | while read line; do mysql -e "show table status from $line"; done )

real	5m38.335s
user	0m0.810s
sys	0m0.807s
```

SuperQ:
Yea, that seems to be a problem with MySQL <= 5.6.

SuperQ:
Ping @wrouesnel This needs another rebase, we we did another round of code cleanup. :)

wrouesnel:
Indeed. I've refactored this anyway, need the clean it up and rebase.

On 8 September 2015 10:03:34 pm AEST, Ben Kochie <notifications@github.com> wrote:
>Ping @wrouesnel This needs another rebase, we we did another round of
>code cleanup. :)
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/prometheus/mysqld_exporter/pull/23#issuecomment-138532848

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

SuperQ:
@wrouesnel Ping, I have some additional metrics I would like to scrape from `information_schema.tables`.  If you want, I could work on this and make it more generic.  For example, I would like to have metrics for `TABLE_ROWS`, `DATA_LENGTH`, `INDEX_LENGTH`, etc.

wrouesnel:
I actually added those already to my working branch, I just got frustrated by the rather fluffy way the mysqld plugin types its outputs compared to postgres and then got super busy. I'll see if I can't get it finished up this week.

SuperQ:
Ping. :smile:

SuperQ:
Ping, it would be cool to get the refactored version of this in, I would like to be able to scrape info_schema for table size data.

wrouesnel:
I am actually seeing these! I've had a busy week or so so I didn't get round to looking at this yet.

SuperQ:
More ping!  Just let me know if you want me to work on this feature.

SuperQ:
Replaced by https://github.com/prometheus/mysqld_exporter/pull/51

PR #25, opened 2015-08-31
Fix regression that broke userstats

wrouesnel:
Refactoring the pointer dancing done in the userstats module led to a regression which broke collecting these statistics at all. This commit has been tested and fixes the issue.


bjk-soundcloud:
:+1: Thanks for the fix.

PR #26, opened 2015-08-31
General refactoring.

beorn7:
Break up the giant scrape method.
Do the error handling a bit less repetitive.

I also had some initial attempts on unit testing in my branch, but that didn't go far yet.

@SuperQ

### DIFF
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -73,6 +73,9 @@ const (
 
 // Metric SQL Queries.
 const (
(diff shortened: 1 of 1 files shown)


beorn7:
@SuperQ This should be relatively trivial to review and test. But it's not trivial to keep this up to date if other commits go in. So could you have a look soonish?

SuperQ:
Yup, tested, looks good. :+1:

PR #28, opened 2015-09-07
Limit events_statements to recently used

SuperQ:
Don't collecting metrics that have not been used recently.
* Default to 1 day.
* Fix query string formatting.

### DIFF
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -54,6 +54,10 @@ var (
 		"collect.perf_schema.eventsstatements.limit", 250,
 		"Limit the number of events statements digests by response time",
 	)
+	perfEventsStatementsTimeLimit = flag.Int(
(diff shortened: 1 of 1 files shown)


mysqld_exporter.go
```diff
@@ -124,6 +124,7 @@ const (
 		    SUM_CREATED_TMP_TABLES
 		  FROM performance_schema.events_statements_summary_by_digest
 		  WHERE SCHEMA_NAME NOT IN ('mysql', 'performance_schema', 'information_schema')
+		    AND last_seen > DATE_SUB(NOW(), INTERVAL 1 DAY)
```
beorn7: How about making this configurable via flag?

SuperQ:
@beorn7 Ok, I made it a param, I also realized I was being silly with the string formatting for the query.  Fixed it for both params.

beorn7:
:+1:

PR #35, opened 2015-09-21
Add a collect.binlog_size flag that enables an aggregated binlog size gauge

rtreffer:
SHOW MASTER LOGS or SHOW BINARY LOGS can be used to retrieve a
list of all binary logs (including size). Summing up the sizes
gives a rough counter for the current binlog size.

See https://dev.mysql.com/doc/refman/5.7/en/show-binary-logs.html
for details.

This gauge simply sums the File_size column. (Note that the total size will grow/shrink as new binlogs are written and old ones are purged)

Required mysql permission: SUPER or REPLICATION CLIENT

### DIFF
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -30,6 +30,10 @@ var (
 		"collect.auto_increment.columns", false,
 		"Collect auto_increment columns and max values from information_schema",
 	)
+	binlogSize = flag.Bool(
+		"collect.binlog_size", false,
+		"Collect the current size of all registered binlog files",
+	)
 	perfTableIOWaits = flag.Bool(
 		"collect.perf_schema.tableiowaits", false,
 		"Collect metrics from performance_schema.table_io_waits_summary_by_table",
@@ -78,6 +82,7 @@ const (
 	informationSchema = "info_schema"
 	performanceSchema = "perf_schema"
 	slaveStatus       = "slave_status"
+	binlog            = "binlog"
 )
 
 // Metric SQL Queries.
@@ -85,6 +90,7 @@ const (
 	globalStatusQuery            = `SHOW GLOBAL STATUS`
 	globalVariablesQuery         = `SHOW GLOBAL VARIABLES`
 	slaveStatusQuery             = `SHOW SLAVE STATUS`
+	binlogQuery                  = `SHOW BINARY LOGS`
 	infoSchemaAutoIncrementQuery = `
 		SELECT table_schema, table_name, column_name, auto_increment,
 		  pow(2, case data_type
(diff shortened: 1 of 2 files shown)


SuperQ:
Nice.  Maybe also count the number of binlog files?

rtreffer:
@SuperQ good idea, added.

rtreffer:
Is there anything left that I should do for this pr?

mysqld_exporter.go
```diff
@@ -154,6 +160,16 @@ var landingPage = []byte(`<html>
 
 // Metric descriptors for dynamically created metrics.
 var (
+	binlogSizeDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, binlog, "size"),
```
SuperQ: this should be `size_total`

mysqld_exporter.go
```diff
@@ -154,6 +160,16 @@ var landingPage = []byte(`<html>
...
+		"Combined size of all registered binlog files.",
+		[]string{}, nil,
+	)
+	binlogFilesDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, binlog, "files"),
```
SuperQ: Same here, `files_total`

mysqld_exporter.go
```diff
@@ -78,6 +82,7 @@ const (
 	informationSchema = "info_schema"
 	performanceSchema = "perf_schema"
 	slaveStatus       = "slave_status"
+	binlog            = "binlog"
```
SuperQ: I'm trying to decide if I like this name.  What do you think of calling this `master_logs` to match the source of the data.  This way the final metric name would be `mysql_master_logs_size_total`.

SuperQ:
Everything is good code wise. I had a think about the naming, some suggestions added.  Let me know what you think.

mysqld_exporter.go
SuperQ: @juliusv Oh wait, is this the correct way to name gauge metrics?

rtreffer:
I have no preference on the names.

The files are usually called "mysql-bin" on disk, the options are called binlog (e.g. max_binlog_size) and the command is either "show master logs" or "show binary logs".

How about changing the query to SHOW BINARY LOGS ?

mysqld_exporter.go
matthiasr: No, `_total` is for counters ("total over all time ")

mysqld_exporter.go
```diff
@@ -154,6 +160,16 @@ var landingPage = []byte(`<html>
 
 // Metric descriptors for dynamically created metrics.
 var (
+	binlogSizeDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, binlog, "size_total"),
```
SuperQ: Oops, I guess I was wrong about `_total` for gauges.  Sorry.

SuperQ:
Ahh, I see, yes, it makes sense to use `binlog` if many of the control flags are named this way.  `SHOW BINARY LOGS` would be fine, is more consistent with the internal naming.

SuperQ:
Please squash the commits.

rtreffer:
@SuperQ ok, will squash

What's the correct name for the gauge values? Was size and files correct? Or size + files_total? Or any other mix?

SuperQ:
The previous names of `size` and `files` were correct.

rtreffer:
Squashed.

SuperQ:
:+1: :shipit:

mysqld_exporter.go
juliusv: But if this is a size gauge in bytes, add a unit suffix like `..._binlog_size_bytes`. If it's just a current count of something (like the `..._binlog_files` metric below, no suffix is needed).

PR #38, opened 2015-09-21
Performance Schema values are uint

SuperQ:
Update performance_schema table output values to be unsinged int.


SuperQ:
@grobie Yes, this will cause loss of precision for the counters, but it should be fine for the time counters.

grobie:
:+1: while the lack of tests should be addressed of course :P

SuperQ:
Yes, tests would be shiny.

PR #40, opened 2015-09-22
Add metrics for table locks

SuperQ:
Gather counter and time metrics from
`performance_schema.table_lock_waits_summary_by_table`[0]

Separate out SQL and External locks

[0]: https://dev.mysql.com/doc/refman/5.6/en/table-waits-summary-tables.html#table-lock-waits-summary-by-table-table

### DIFF
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -46,6 +46,14 @@ var (
 		"collect.perf_schema.indexiowaitstime", false,
 		"Collect time metrics from performance_schema.table_io_waits_summary_by_index_usage",
 	)
+	perfTableLockWaits = flag.Bool(
+		"collect.perf_schema.tablelocks", false,
+		"Collect metrics from performance_schema.table_lock_waits_summary_by_table",
+	)
+	perfTableLockWaitsTime = flag.Bool(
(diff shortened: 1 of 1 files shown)


mysqld_exporter.go
```diff
@@ -204,6 +248,26 @@ var (
 		"The total time of index I/O wait events for each index and operation.",
 		[]string{"schema", "name", "index", "operation"}, nil,
 	)
+	performanceSchemaSQLTableLockWaitsDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, performanceSchema, "sql_lock_waits_total"),
+		"The total number of sql lock wait events for each table and operation.",
```
juliusv: "sql" -> "SQL" here and in the help string below.

juliusv:
:+1: otherwise, but I don't know anything about MySQL internals.

SuperQ:
@juliusv In theory, this should help to identify tables that are blocking on internal thread locking.  However the upstream MySQL documentation is very lacking on what each of the actual lock methods mean. :hankey:

fabxc:
:+1:

PR #43, opened 2015-09-28
Fix some metrics suffixes.

juliusv:
Gauges should end with a unit suffix (unless they are counts of
some object that is already mentioned in the metric name).

Counters should end with a _total suffix.


juliusv:
@SuperQ @brian-brazil

mysqld_exporter.go
```diff
@@ -205,7 +205,7 @@ var landingPage = []byte(`<html>
 // Metric descriptors for dynamically created metrics.
 var (
 	binlogSizeDesc = prometheus.NewDesc(
-		prometheus.BuildFQName(namespace, binlog, "size"),
+		prometheus.BuildFQName(namespace, binlog, "size_bytes"),
```
juliusv: Actually, I'm not even sure if this is in bytes or not, which is exactly the problem - please verify.

mysqld_exporter.go
```diff
@@ -349,79 +349,79 @@ var (
 				"The number of concurrent connections for this user.",
 				[]string{"user"}, nil)},
 		"CONNECTED_TIME": {prometheus.CounterValue,
-			prometheus.NewDesc(prometheus.BuildFQName(namespace, informationSchema, "user_statistics_connected_time"),
+			prometheus.NewDesc(prometheus.BuildFQName(namespace, informationSchema, "user_statistics_connected_time_total"),
```
brian-brazil: Do we know the unit here?

mysqld_exporter.go
brian-brazil: It's bytes.

mysqld_exporter.go
juliusv: Oh right, same for the metrics below, they also need a time unit. I don't know it, hopefully @SuperQ does.

juliusv: Ah, help string says seconds already, will update the metric name.

juliusv: Updated to include `_seconds`.

brian-brazil:
:+1:

SuperQ:
:+1: Thanks for fixing.

PR #50, opened 2015-10-18
Simplify `digest_text`

SuperQ:
Simplify `events_statements_summary_by_digest` `digest_text` by
including the text label in all metrics.  Since the digest and
digest_text are 1:1 there is no need to separate the text out.


SuperQ:
@fabxc As we discussed, this will simplify things for mysqld_exporter users.

fabxc:
:+1:

PR #52, opened 2015-10-27
Remove example query section

fabxc:
With the simplified digest_text labels this section no longer applies.
Also nothing else worth replacing this with comes to mind.

@SuperQ


SuperQ:
:+1:

PR #54, opened 2015-11-11
Add option to automatic ANALYZE TABLE

SuperQ:
MySQL >= 5.6 does not automatically update table statistics due to
performance impact.

This allows the mysqld_exporter scrapes to trigger stats updates on
demand.  This could have a load impact on MySQL so best to use in an
out-of-band MySQL server.  For example a hot-backup slave.

### DIFF
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"flag"
 	"fmt"
+	"math/rand"
 	"net/http"
 	"os"
 	"regexp"
@@ -41,6 +42,10 @@ var (
 	tableSchemaAutoAnalyze = flag.Bool(
 		"collect.info_schema.tables.auto_analyze", false,
 		"Automatically run ANALYZE TABLE to update stats",
+	)
+	tableSchemaAutoAnalyzeDuration = flag.Duration(
+		"collect.info_schema.tables.auto_analyze.min_duration", 5*time.Minute,
+		"Minimum seconds between automatic ANALYZE TABLE calls",
 	)
 	collectGlobalStatus = flag.Bool(
 		"collect.global_status", true,
@@ -668,6 +673,9 @@ type Exporter struct {
 	duration, error prometheus.Gauge
 	totalScrapes    prometheus.Counter
 }
+
+// Last run tracking variable for scrapeTableSchema().
+var tableSchemaLastRun = map[string]time.Time{}
 
 // NewExporter returns a new MySQL exporter for the provided DSN.
 func NewExporter(dsn string) *Exporter {
@@ -1656,8 +1664,12 @@ func scrapeTableSchema(db *sql.DB, ch chan<- prometheus.Metric) error {
 	}
 
 	for _, database := range dbList {
-		if *tableSchemaAutoAnalyze {
+		random := 1 + rand.Float64() * 0.25
+		nextRun := time.Duration(float64(*tableSchemaAutoAnalyzeDuration)*random)
+		if *tableSchemaAutoAnalyze && time.Since(tableSchemaLastRun[database]) > nextRun {
 			tablesRows, err := db.Query(fmt.Sprintf(tableListQuery, database))
+			tableSchemaLastRun[database] = time.Now()
+			log.Debugf("Ran AutoAnalyze on %s", database)
 			if err != nil {
 				return err
 			}
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -38,6 +38,10 @@ var (
 		"collect.info_schema.tables.databases", "*",
 		"The list of databases to collect table stats for, or '*' for all",
 	)
+	tableSchemaAutoAnalyze = flag.Bool(
+		"collect.info_schema.tables.auto_analyze", false,
+		"Automatically run ANALYZE TABLE to update stats",
+	)
 	collectGlobalStatus = flag.Bool(
 		"collect.global_status", true,
 		"Collect from SHOW GLOBAL STATUS",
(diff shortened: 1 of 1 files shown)


SuperQ:
@wrouesnel I don't know if you've looked at the output of information_schema.tables, but it seems like the stats don't update unless you manually `ANALYZE TABLE`.

wrouesnel:
I did not know this. Although I suspect in my environment mysql goes up and down so frequently I don't notice (cpanel hosting) and we mostly use it for crashed table detection.

wrouesnel:
It might be an idea to have that option accept a Duration so the maximum frequency of stats updates can be set.

SuperQ:
Ahh, that's an interesting idea, I'll see about adding that.

mysqld_exporter.go
```diff
@@ -38,6 +38,10 @@ var (
 		"collect.info_schema.tables.databases", "*",
 		"The list of databases to collect table stats for, or '*' for all",
 	)
+	tableSchemaAutoAnalyze = flag.Bool(
+		"collect.info_schema.tables.auto_analyze", false,
+		"Automatically run ANALIZE TABLE to update stats",
```
beorn7: ANALIZE → ANALYZE

beorn7:
WRT the "max frequency of updates" option: That seems simple enough. Just save the results in a variable and record `lastRun := time.Now()`. Whenever a scrape happens again, return the values from the variable if `time.Since(lastRun) < flag.MinUpdateDurationwhatever`. Modulo goroutine safety.

Happy to assist for details.

I'd probably only have one flag (the duration), and at its default value, it will do nothing.

SuperQ:
@beorn7 @fabxc I've added the Duration control, including randomness to avoid running `ANALYZE TABLE` in a single scrape.

mysqld_exporter.go
```diff
@@ -656,6 +674,11 @@ type Exporter struct {
 	totalScrapes    prometheus.Counter
 }
 
+// Last run tracking variable for scrapeTableSchema()
+var (
```
fabxc: Terminate comment with `.`
A single var shouldn't have a `()` block for consistency.

mysqld_exporter.go
```diff
@@ -1643,6 +1666,33 @@ func scrapeTableSchema(db *sql.DB, ch chan<- prometheus.Metric) error {
 	}
 
 	for _, database := range dbList {
+		if _, ok := tableSchemaLastRun[database]; !ok {
```
fabxc: You can actually skip the whole thing.
A map will return you the zero value if the element doesn't exist. Thus `tableSchemaLastRun["neverseenbefore"] == time.Time{}`.

beorn7:
Will have a look ASAP. (Still working through my backlog...)

mysqld_exporter.go
```diff
@@ -1643,6 +1666,33 @@ func scrapeTableSchema(db *sql.DB, ch chan<- prometheus.Metric) error {
...
+			tableSchemaLastRun[database] = time.Time{}
+		}
+		random := rand.Float64() * 0.25
+		nextRun := *tableSchemaAutoAnalyzeDuration + time.Duration(float64(*tableSchemaAutoAnalyzeDuration)*random)
```
beorn7: The following should have the same effect:

```
random := 1 + rand.Float64() * 0.25
nextRun := time.Duration(float64(*tableSchemaAutoAnalyzeDuration)*random)
```

mysqld_exporter.go
```diff
@@ -1643,6 +1666,33 @@ func scrapeTableSchema(db *sql.DB, ch chan<- prometheus.Metric) error {
...
+		if *tableSchemaAutoAnalyze && time.Since(tableSchemaLastRun[database]) > nextRun {
```
beorn7: What @fabxc meant above (in case that isn't clear anyway): This line will work just fine because for unknown databases, the zero value of time.Time is returned (which is very far in the past), so the analyze will be run on new databases (which, actually, with what you are doing above, it will only run 5min later...)

beorn7: In different news, since scrapes could run multiple times in parallel, you need to protect access to `tableSchemaLastRun` by a mutex. Probably best to protect this whole if block.

SuperQ:
@fabxc @beorn7 Updated, PTAL

beorn7:
My concurrency concerns might have been missed now.

mysqld_exporter.go
fabxc: No, he initializes with the zero time value as well. It's literally a no-op.

fabxc:
:+1:

beorn7:
Yeah, the mutex is important. So :-1:, i.e. please don't merge yet.

beorn7:
@SuperQ Are you still on this? Happy to help you with the code in person, once you are healthy again and back in the office.

SuperQ:
Yes, let's pair on this soon.

SuperQ:
I've thought about this some more and decided to close this feature.  While it would be cool if the exporter would trigger these automatically, I think it's much more appropriate to do this out-of-band from the exporter.

beorn7:
Wait, but why? ;)

roman-vynar:
`ANALYZE TABLE` is not a safe operation  as it acquires a read lock for InnoDB/MyISAM table being analyzed and having a monitoring tool doing this would be not great. 
Moreover, an exporter is supposed to perform read-only job to expose data and definitely not bother the system such as rebuilding stats, fixing replication etc. 🤓

SuperQ:
@roman-vynar Yup, that pretty much sums it up.

beorn7:
Thanks.

PR #60, opened 2016-01-05
Avoid slow query logging

SuperQ:
Add a session log_slow_filter[0] to avoid slow query logging of scrapes.
* Add a flag, default false, because this is not supported by Oracle MySQL.

[0]: https://www.percona.com/doc/percona-server/5.6/diagnostics/slow_extended.html

### DIFF
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -26,6 +26,10 @@ var (
 		"web.telemetry-path", "/metrics",
 		"Path under which to expose metrics.",
 	)
+	slowLogFilter = flag.Bool(
+		"log_slow_filter", false,
+		"Add a log_slow_filter to avoid exessive MySQL slow logging.  NOTE: Not supported by Oracle MySQL.",
+	)
 	collectProcesslist = flag.Bool(
 		"collect.info_schema.processlist", false,
 		"Collect current thread state counts from the information_schema.processlist",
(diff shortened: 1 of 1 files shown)


mysqld_exporter.go
```diff
@@ -26,6 +26,10 @@ var (
 		"web.telemetry-path", "/metrics",
 		"Path under which to expose metrics.",
 	)
+	slowLogFilter = flag.Bool(
+		"log_slow_filter", false,
+		"Add a log_slow_filter to avoid exessive MySQL slow logging",
```
beorn7: End with period for consistency (same below, actually…).

mysqld_exporter.go
```diff
@@ -737,6 +742,14 @@ func (e *Exporter) scrape(ch chan<- prometheus.Metric) {
 	}
 	defer db.Close()
 
+	if *slowLogFilter {
+		sessionSettingsRows, err := db.Query(sessionSettingsQuery)
+		if err != nil {
+			return err
+		}
+		defer sessionSettingsRows.Close()
```
beorn7: Is there a reason to defer the Close call to the end of the function? sessionSettingsRows goes out of scope here, so you could close it right ahead.

mysqld_exporter.go
beorn7: Perhaps add a comment that Oracly MySQL doesn't support it?

SuperQ:
@beorn7 Updated with your requested changes.

beorn7:
:+1:

PR #61, opened 2016-01-15
Leverage lock-free SHOW SLAVE STATUS

roman-vynar:
Leverage lock-free `SHOW SLAVE STATUS` by choosing the right suffix on the first run and than using it on the next queries.

[Percona Server 5.5] (https://www.percona.com/doc/percona-server/5.5/reliability/show_slave_status_nolock.html) supports non-blocking `SHOW SLAVE STATUS NOLOCK`.
[Percona Server 5.6] (https://www.percona.com/doc/percona-server/5.6/reliability/show_slave_status_nolock.html) supports non-blocking `SHOW SLAVE STATUS NOLOCK` as well as `SHOW SLAVE STATUS NONBLOCKING` to match initial MySQL 5.7.
[Oracle MySQL 5.7] (http://dev.mysql.com/doc/refman/5.7/en/show-slave-status.html) initially supported `SHOW SLAVE STATUS NONBLOCKING`. As of 5.7.6 they removed the suffix due to locking changes and it became unnecessary.

### DIFF
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -251,6 +251,7 @@ const (
 		  WHERE SCHEMA_NAME NOT IN ('mysql', 'performance_schema', 'information_schema')
 		`
 )
+var slaveStatusQuerySuffixes = [3]string{" NONBLOCKING", " NOLOCK", ""}
 
 // landingPage contains the HTML served at '/'.
 // TODO: Make this nicer and more informative.
@@ -918,7 +919,17 @@ func scrapeGlobalVariables(db *sql.DB, ch chan<- prometheus.Metric) error {
 }
 
 func scrapeSlaveStatus(db *sql.DB, ch chan<- prometheus.Metric) error {
-	slaveStatusRows, err := db.Query(slaveStatusQuery)
+	var (
+		slaveStatusRows *sql.Rows
+		err error
+	)
+	// Leverage lock-free SHOW SLAVE STATUS by guessing the right suffix
+	for _, suffix := range slaveStatusQuerySuffixes {
+		slaveStatusRows, err = db.Query(fmt.Sprint(slaveStatusQuery, suffix))
+		if err == nil {
+			break
+		}
+	}
 	if err != nil {
 		return err
 	}


SuperQ:
Hmm, this is tricky because every scrape is a new mysql connection.  This means if a mysqld changes out from under the exporter, it can remember the wrong setting.

SuperQ:
I think I would prefer this to be a simple flag that can be passed, rather than try and make it automatic.

roman-vynar:
I have added a reset in the last commit in case MySQL changes.
Does it make sense?

SuperQ:
Hmmm, I still think it would be better to make this a configuration option.  This way it's easy to deploy the change in sync with a configuration management system.

Making this exporter more "automatic" is on the TODO list, but I think probing this way is the wrong approach.

roman-vynar:
Actually, we have implemented similar thing in [Cacti/Nagios plugins] (https://github.com/percona/percona-monitoring-plugins) and private MHA wrappers and I can say it works well. Those suffixes which are invalid are only causing MySQL client-side error and not affecting a server at all. So guessing the right suffix is really not a problem.

From other point of view, take a look how many times it was changed from version to version 😀 no chance than guessing it 😎

Furthermore, `SHOW SLAVE STATUS` may cause a deadlock, that's why it would be great to avoid it if we can taking into account Prometheus triggers this on a second resolution, any FTWRL and MySQL stuck:
http://bugs.mysql.com/bug.php?id=67879
http://bugs.mysql.com/bug.php?id=56676
https://bugs.launchpad.net/percona-monitoring-plugins/+bug/1030490/comments/6

SuperQ:
I understand the importance of doing this, but I question the method.  I will think about this a bit and get back to you soon.

SuperQ:
@beorn7 Can you take a look at this?

beorn7:
On my review queue.

mysqld_exporter.go
```diff
@@ -656,6 +656,9 @@ type Exporter struct {
 	totalScrapes    prometheus.Counter
 }
 
+// Suffix to use with SHOW SLAVE STATUS
+var slaveStatusSuffix string
```
beorn7: I would put this higher up and not between the `Exporter` definition and its constructor.

mysqld_exporter.go
```diff
@@ -918,8 +921,26 @@ func scrapeGlobalVariables(db *sql.DB, ch chan<- prometheus.Metric) error {
 }
 
 func scrapeSlaveStatus(db *sql.DB, ch chan<- prometheus.Metric) error {
-	slaveStatusRows, err := db.Query(slaveStatusQuery)
+	var (
+		slaveStatusRows *sql.Rows
+		err error
+	)
+	// Leverage lock-free SHOW SLAVE STATUS
+	if slaveStatusSuffix == "" {
```
beorn7: Scrapes can happen concurrently, so this block will have a race condition.

Perhaps find out the suffix upon start and not as part of the first scrape?

beorn7:
OK, doing it once on startup would defeat the purpose of what you discussed already. :)

So you needed to do same mutex magic around this (which is probably OK), but makes it even more complex.

SuperQ:
Another option is to simply test and execute every time.  This should be inexpensive.

roman-vynar:
So let's do this every time?

SuperQ:
Yea, I think that's the easiest solution.

roman-vynar:
Done.

mysqld_exporter.go
```diff
@@ -918,7 +918,19 @@ func scrapeGlobalVariables(db *sql.DB, ch chan<- prometheus.Metric) error {
 }
 
 func scrapeSlaveStatus(db *sql.DB, ch chan<- prometheus.Metric) error {
-	slaveStatusRows, err := db.Query(slaveStatusQuery)
+	var (
+		slaveStatusRows *sql.Rows
+		err error
+	)
+	// Leverage lock-free SHOW SLAVE STATUS.
+	// Iterate over possible query suffixes to find a working one and remember it for the next usage.
```
beorn7: This comment is obsolete now.

roman-vynar:
Fixed.

beorn7:
:+1: Thanks. @SuperQ Feel free to merge if you are fine with the change.

mysqld_exporter.go
```diff
@@ -918,7 +918,18 @@ func scrapeGlobalVariables(db *sql.DB, ch chan<- prometheus.Metric) error {
 }
 
 func scrapeSlaveStatus(db *sql.DB, ch chan<- prometheus.Metric) error {
-	slaveStatusRows, err := db.Query(slaveStatusQuery)
+	var (
+		slaveStatusRows *sql.Rows
+		err error
+	)
+	// Leverage lock-free SHOW SLAVE STATUS by guessing the right suffix
+	querySuffixes := []string{" NONBLOCKING", " NOLOCK", " "}
```
SuperQ: The last item can be empty string (`""`) now that this doesn't need to remember.

SuperQ: Actually, this querySuffixes can be moved to a global const.  Maybe `slaveStatusQuerySuffixes`

SuperQ:
Yea, this is overall much simpler, thanks for being patient with the review process.

mysqld_exporter.go
```diff
@@ -918,7 +918,18 @@ func scrapeGlobalVariables(db *sql.DB, ch chan<- prometheus.Metric) error {
 }
 
 func scrapeSlaveStatus(db *sql.DB, ch chan<- prometheus.Metric) error {
-	slaveStatusRows, err := db.Query(slaveStatusQuery)
+	var (
+		slaveStatusRows *sql.Rows
+		err error
+	)
+	// Leverage lock-free SHOW SLAVE STATUS by guessing the right suffix
+	querySuffixes := []string{" NONBLOCKING", " NOLOCK", ""}
```
SuperQ: Oops, my second comment got lost.

This querySuffixes can be moved to a global const.  Maybe `slaveStatusQuerySuffixes`

roman-vynar:
Any better now? 😇

mysqld_exporter.go
```diff
@@ -251,6 +251,7 @@ const (
 		  WHERE SCHEMA_NAME NOT IN ('mysql', 'performance_schema', 'information_schema')
 		`
 )
+var slaveStatusQuerySuffixes = [3]string{" NONBLOCKING", " NOLOCK", ""}
```
SuperQ: s/var/const/

SuperQ:
Much better, almost done. :)

mysqld_exporter.go
roman-vynar:
```
./mysqld_exporter.go:254: const initializer [3]string literal is not a constant
```
Looks like there are no const arrays in golang as they are mutable in nature.

beorn7: True. But you can still have it as a top level var.

roman-vynar: It is a top-level var.

SuperQ: Ahh, my mistake on the const thing.. looks great.

PR #63, opened 2016-01-16
Early abort slave status

SuperQ:
Early abort scrapeSlaveStatus() when there is no output from
`SHOW SLAVE STATUS`

### DIFF
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -1074,35 +1074,37 @@ func scrapeSlaveStatus(db *sql.DB, ch chan<- prometheus.Metric) error {
 	}
 	defer slaveStatusRows.Close()
 
-	if slaveStatusRows.Next() {
-		// There is either no row in SHOW SLAVE STATUS (if this is not a
-		// slave server), or exactly one. In case of multi-source
-		// replication, things work very much differently. This code
-		// cannot deal with that case.
-		slaveCols, err := slaveStatusRows.Columns()
-		if err != nil {
-			return err
-		}
+	// There is either no row in SHOW SLAVE STATUS (if this is not a
+	// slave server), or exactly one. In case of multi-source
+	// replication, things work very much differently. This code
+	// cannot deal with that case.
+	if !slaveStatusRows.Next() {
+		return nil
+	}
 
-		// As the number of columns varies with mysqld versions,
-		// and sql.Scan requires []interface{}, we need to create a
-		// slice of pointers to the elements of slaveData.
-		scanArgs := make([]interface{}, len(slaveCols))
-		for i := range scanArgs {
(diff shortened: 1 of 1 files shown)


SuperQ:
@grobie Thanks for the idea :)

mysqld_exporter.go
```diff
@@ -937,35 +937,38 @@ func scrapeSlaveStatus(db *sql.DB, ch chan<- prometheus.Metric) error {
...
-		// slave server), or exactly one. In case of multi-source
-		// replication, things work very much differently. This code
-		// cannot deal with that case.
-		slaveCols, err := slaveStatusRows.Columns()
-		if err != nil {
-			return err
-		}
+	// Don't bother if there is no slave status row
+	if !slaveStatusRows.Next() {
+		return nil
```
brian-brazil: You still need to return the metrics in this case, otherwise you'll be working off stale data.

SuperQ: Sorry, that's not really possible.  This scrape function works off of `SHOW SLAVE STATUS` which returns an empty set result with no column headers so we have no idea what metrics are there or not.

If you look at the change more closely, there is no functional difference, this just allows us to remove one level of indent, and abort early.

brian-brazil: We know the main metrics that we're expecting, we should export them as 0 or NaN.

SuperQ: No, look more closely.  We don't know what the "main metrics" are at all.  This exporter feature is completely dynamic, the list of slave metrics are generated every scrape.

brian-brazil: I'm aware of that, I'm also aware that we know what metrics we can expect to be returned. Show slave status is pretty standard.

SuperQ: Are you sure about that?  I know for a fact that it changes from version to version, hence the dynamic nature of this exporter.

brian-brazil: It changes, but things like delay and log position have been the same for a long time.

mysqld_exporter.go
```diff
@@ -937,35 +937,38 @@ func scrapeSlaveStatus(db *sql.DB, ch chan<- prometheus.Metric) error {
...
 
-		// As the number of columns varies with mysqld versions,
-		// and sql.Scan requires []interface{}, we need to create a
-		// slice of pointers to the elements of slaveData.
-		scanArgs := make([]interface{}, len(slaveCols))
-		for i := range scanArgs {
-			scanArgs[i] = &sql.RawBytes{}
-		}
+	// There is either no row in SHOW SLAVE STATUS (if this is not a
+	// slave server), or exactly one. In case of multi-source
```
grobie: If there is no row, then `slaveStatusRows.Next` will already fail I assume, so this comment needs to be moved up.

PR #64, opened 2016-01-17
Fixing the return of the function and logging the error

knabben:
Compiling with default settings I got:

./mysqld_exporter.go:748: too many arguments to return


SuperQ:
Ahh, good catch, I thought I had tested that one better.

PR #68, opened 2016-01-19
Add labels to global status "handlers" counters

SuperQ:
SHOW GLOBAL STATUS Handler_.* is similar to Com_.*

Aggregate the handlers into a single metric with labels for the various
handler types.


roman-vynar:
:+1:

PR #74, opened 2016-01-27
Check log_bin before running SHOW BINARY LOGS

roman-vynar:
`SHOW BINARY LOGS` unlike other SQL queries produces the error when `log-bin` is OFF:
```
INFO[0000] Error scraping binlog size: Error 1381: You are not using binary logging  file=mysqld_exporter.go line=797
```
This causes all other scrapes coming after `scrapeBinlogSize()` to not execute, in particular all `scrapePerf*` ones. Added a check that prevents running this query when `log_bin` if OFF.

### DIFF
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -127,6 +127,7 @@ const (
 	globalVariablesQuery       = `SHOW GLOBAL VARIABLES`
 	slaveStatusQuery           = `SHOW SLAVE STATUS`
 	binlogQuery                = `SHOW BINARY LOGS`
+	logbinQuery                = `SELECT @@log_bin`
 	infoSchemaProcesslistQuery = `
 		SELECT COALESCE(command,''),COALESCE(state,''),count(*)
 		FROM information_schema.processlist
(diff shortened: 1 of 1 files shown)


mysqld_exporter.go
```diff
@@ -1012,6 +1012,16 @@ func scrapeInformationSchema(db *sql.DB, ch chan<- prometheus.Metric) error {
 }
 
 func scrapeBinlogSize(db *sql.DB, ch chan<- prometheus.Metric) error {
+	var logBin int8
+	err := db.QueryRow("SELECT @@log_bin").Scan(&logBin)
```
SuperQ: I would prefer that all queries are templated in the section up at line #124.

SuperQ:
Good catch, besides the query string, looks like a good idea.  I have plans to cleanup this exporter to not error when a single scrape collector fails.

roman-vynar:
Fixed.

And yes, it would be great especially if the exporter does not fail on any SELECT from `performance_schema` tables when they do not exist on a given MySQL version.

PR #77, opened 2016-02-02
Added .my.cnf support

roman-vynar:
Adding support for reading MySQL credentials from `.my.cnf`. Using environmental variable to pass the DSN is not really flexible from automation point of view.

### DIFF
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -1796,18 +1796,14 @@ func parseMycnf() (string) {
 		}
 		return in
 	})
-	port := cfg.Section("client").Key("port").Validate(func(in string) string {
-		if len(in) == 0 {
-			return "3306"
-		}
-		return in
-	})
+	host := cfg.Section("client").Key("host").MustString("localhost")
+	port := cfg.Section("client").Key("port").MustUint(3306)
 	socket := cfg.Section("client").Key("socket").String()
 	var dsn string
 	if socket != "" {
 		dsn = fmt.Sprintf("%s:%s@unix(%s)/", user, password, socket)
 	} else {
-		dsn = fmt.Sprintf("%s:%s@tcp(localhost:%s)/", user, password, port)
+		dsn = fmt.Sprintf("%s:%s@tcp(%s:%d)/", user, password, host, port)
 	}
 	log.Debugln(dsn)
 	return dsn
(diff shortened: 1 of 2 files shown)


roman-vynar:
Oh, looks like I messed the branch, solving the conflict.

SuperQ:
I don't understand why DSN is not flexible for automation.  We automatically deploy DSN with Chef and other methods.

Here's how it works:
We have our automation deploy the DSN to a file with chef:
```
file "/srv/mysqld_exporter/env/DATA_SOURCE_NAME" do
  owner   'root'
  group   'root'
  mode    '0600'
  content "#{mysql_user}:#{mysql_pass}@/"
  action  :create
  notifies :restart, 'service[mysqld_exporter]'
end
```

Then the service is setup to run mysqld exporter like this:

```
exec chpst -u nobody -e /srv/mysqld_exporter/env \
  /srv/mysqld_exporter/mysqld_exporter \
   -options...
```

SuperQ:
There are also easy ways to wrap the launch in a simple shell script and source the ENV file.

matthiasr:
I think this is a worthwhile addition anyway, `my.cnf` _is_ the default mechanism to set mysql credentials. It's not 12factory but that is not the point for a mysql exporter anyway.

roman-vynar:
That what we are currently doing, having own startup files with ENV var set.

However, in our work we find this unnecessary because we employ `.my.cnf` in user's home directory. Many tools support `.my.cnf`, Python's MySQLdb supports `read_default_file=`, golang doesn't and employs those DSNs.

Still this is optional and I don't try to impose the way how to use this exporter.
I am following the best practices in MySQL configuration.
Thanks.

SuperQ:
Yea, I guess, probably a good idea.

mysqld_exporter.go
```diff
@@ -1773,12 +1779,46 @@ func parseStatus(data sql.RawBytes) (float64, bool) {
...
+			return "3306"
+		}
+		return in
+	})
+	socket := cfg.Section("client").Key("socket").String()
+	var dsn string
+	if socket != "" {
+		dsn = fmt.Sprintf("%s:%s@unix(%s)/", user, password, socket)
+	} else {
+		dsn = fmt.Sprintf("%s:%s@tcp(localhost:%s)/", user, password, port)
```
SuperQ: Please add optional support for `cfg.Section("host")` with default `localhost`

SuperQ:
:+1:

roman-vynar:
Thanks 🍺

PR #81, opened 2016-02-12
Fixed uint for scrapeInnodbMetrics() and gofmt

roman-vynar:
Looks like we need the tests and CI 🙂


SuperQ:
Yea, I need to learn how to setup database mocking in go so we can have everything under unit tests.

PR #84, opened 2016-02-14
Soft error on collector failure

SuperQ:
Instead of aborting the collection when a single collector fails, record
an error and continue.

### DIFF
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -708,6 +708,7 @@ type Exporter struct {
 	dsn             string
 	duration, error prometheus.Gauge
 	totalScrapes    prometheus.Counter
+	scrapeErrors    *prometheus.CounterVec
 }
 
 // NewExporter returns a new MySQL exporter for the provided DSN.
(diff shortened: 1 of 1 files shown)


SuperQ:
@fabxc This should fix the issue where some collectors fail because of incompatible mysql versions or mysql bugs.

mysqld_exporter.go
```diff
@@ -726,6 +729,14 @@ func NewExporter(dsn string) *Exporter {
 			Name:      "scrapes_total",
 			Help:      "Total number of times MySQL was scraped for metrics.",
 		}),
+		scrapeErrors: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Namespace: namespace,
+			Subsystem: exporter,
+			Name:      "scrape_errors_total",
+			Help:      "Total number of times an error occured scraping a MySQL.",
+		},
```
fabxc: This is syntactically inconsistent with how we do it elsewhere.
Let's move the labels and the closing `)` directly into this line.

Also `scrapeErrorLabels` is not used elsewhere and pretty simple. We can just directly inline it here – makes it easier to read.

fabxc:
Aside from comment :+1:

mysqld_exporter.go
SuperQ: I did it this way because i was getting compiler errors because of other bugs. I though it was my labels literal that was causing problems.

PR #85, opened 2016-02-14
Fix innodb_metrics collector

SuperQ:
Use correct type exposed by the `information_schema.innodb_metrics` table.
* Work around issues where some counters return negative numbers.

It turns out that MySQL is inconsistent in which metrics are in `SHOW
GLOBAL STATUS` and which are in `information_schema.innodb_metrics`.
* Include all `status_counter` metrics as counters.


SuperQ:
@PeterZaitsev This should fix the collector from hitting negative counter values.

SuperQ:
See upstream bug: http://bugs.mysql.com/bug.php?id=75966

PR #87, opened 2016-02-16
Collect time metrics for processlist

SuperQ:
Collect the total time for each processlist state in addition to the
count of threads in each state.


SuperQ:
@matthiasr Trying to add a test for this, but having trouble.  Not sure the mock driver can handle the query needed for `scrapeProcesslist`

brian-brazil:
:+1:

PR #88, opened 2016-02-16
Parse auto increment values and maximum as float64.

matthiasr:
For large enough values, MySQL returns exponential notation from
`POW()`.  The metric value is a float64 anyway, so we can fix this
mismatch by parsing this as a float in the first place.

@SuperQ

### DIFF
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -1063,7 +1063,7 @@ func scrapeInformationSchema(db *sql.DB, ch chan<- prometheus.Metric) error {
 
 	var (
 		schema, table, column string
-		value, max            uint64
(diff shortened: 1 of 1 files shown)


matthiasr:
Note that we should probably fix all places where we parse output from MySQL as `uint64` and then convert it to a `float64` straight away, but this is a start and fixes the immediate issue we found. I'm not sure about the ramifications of doing it _everywhere_.

SuperQ:
In many places we use specific types that represent the mysql table definition.  In this case the output is coming from a mysql math function which can produce float64 output.

SuperQ:
:+1:

PR #97, opened 2016-02-28
Vendoring

grobie:
Use govendor to vendor all dependencies and shield against changing APIs
or repository unavailability.

PR #98, opened 2016-02-28
Common log

### LINKED ISSUE #95: Add -log.format parameter so we can use syslog
Other exporters have -log.format that allows you to send logs to syslog, this one doesn't have it.

grobie:
Depends on #97.
Fixes #95.

@SuperQ


SuperQ:
:+1: :shipit: :rocket:

Issue #96, opened 2016-02-28
Include individual index IO stats


SuperQ:
It turns out it's incorrect to ignore stats from indexes that are not the "NULL" stat in `collect.perf_schema.indexiowaits`.  There are per-index io stats that are probably useful to see where the database is spending it's time updating individual indexes.

We need to re-evaluate which index stats are included.

SuperQ:
@PeterZaitsev I did some checking, and I can find no cases where `performance_schema.table_io_waits_summary_by_index_usage.COUNT_INSERT` was non-zero outside of the NULL index name.  I found some use of `COUNT_UPDATE` and `COUNT_DELETE`.

Can you confirm this behavior?

SuperQ:
Fixed

PeterZaitsev:
Hi Ben,

Sorry I did not "confirm" it in time. Really I do not check.  In reality there is cost maintaining indexes for IO operations so it  might be bug or not implemented yet.  I know you do not like metrics which appear and dissapear but I think better approach in cases like this to capture stuff which has been non zero (since uptime)  as if the metric is implemented enabled and make sense for given workload it is likely not to have zero value.

SuperQ:
The new version captures everything but INSERT metrics, for now this should be sufficient.  We can always revisit this in the future.

Also, we're working on a new compression method that reduces the overhead for collecting zeros to sub 1 byte per sample. :-)

PR #100, opened 2016-03-13
Include index io stats for update and delete

SuperQ:
As reported by https://github.com/prometheus/mysqld_exporter/issues/96,
we do need to collect metrics for update and delete index IO waits.


roman-vynar:
The test panics because sqlmock expects a regex as an argument for ExpectQuery().

The fix is to escape brackets by rewritting sanitizeQuery():
```
func sanitizeQuery(q string) string {
	q = strings.Join(strings.Fields(q), " ")
	q = strings.Replace(q, "(", "\\(", -1)
	q = strings.Replace(q, ")", "\\)", -1)
	return q
}
```

SuperQ:
Ahh, I see, thanks!

SuperQ:
@roman-vynar Ok, tests pass now, yay!

PR #102, opened 2016-03-14
Fix tests

SuperQ:
Don't use unnamed imports. :crying_cat_face:

### DIFF
--- a/mysqld_exporter_test.go
+++ b/mysqld_exporter_test.go
@@ -6,7 +6,7 @@ import (
 
 	"github.com/prometheus/client_golang/prometheus"
 	dto "github.com/prometheus/client_model/go"
-	. "github.com/smartystreets/goconvey/convey"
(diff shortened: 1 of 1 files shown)


SuperQ:
@beorn7 :joy_cat:

brian-brazil:
:+1:

beorn7:
:+1: 

It would be fine to use a different name (like a one-letter abbreviation). But not just nothing.

SuperQ:
Yea, I think it's short enough as-is.

roman-vynar:
That was their example 😇 https://github.com/smartystreets/goconvey#quick-start

SuperQ:
Yea, apparently that's a bad recommendation according to Go standards.  I had no idea until @beorn7 pointed it out.

beorn7:
For reference: https://github.com/golang/go/wiki/CodeReviewComments#import-dot

PR #106, opened 2016-04-03
Fix scrape errors for MySQL 5.7

SuperQ:
Standardize collector error messages
* Always use the name of collector flag in the error message to avoid confusion when looking at logs.

"DELAYED" Inserts were deprecated in MySQL 5.6.6[0].  These metrics
cause MySQL 5.7 scrapes to fail.

[0]: https://dev.mysql.com/doc/refman/5.6/en/table-waits-summary-tables.html


juliusv:
:+1:

PR #110, opened 2016-04-10
Refactor collectors

SuperQ:
Move `GLOBAL STATUS` scraper to a separate file.

### DIFF
--- a/collector/collector_test.go
+++ b/collector/collector_test.go
@@ -1,0 +1,42 @@
+package collector
+
+import (
+	"strings"
+
+	"github.com/prometheus/client_golang/prometheus"
+	dto "github.com/prometheus/client_model/go"
+)
+
+type labelMap map[string]string
+
+type MetricResult struct {
+	labels     labelMap
+	value      float64
+	metricType dto.MetricType
+}
+
+func readMetric(m prometheus.Metric) MetricResult {
+	pb := &dto.Metric{}
+	m.Write(pb)
+	labels := make(labelMap, len(pb.Label))
+	for _, v := range pb.Label {
+		labels[v.GetName()] = v.GetValue()
+	}
+	if pb.Gauge != nil {
+		return MetricResult{labels: labels, value: pb.GetGauge().GetValue(), metricType: dto.MetricType_GAUGE}
+	}
+	if pb.Counter != nil {
+		return MetricResult{labels: labels, value: pb.GetCounter().GetValue(), metricType: dto.MetricType_COUNTER}
+	}
+	if pb.Untyped != nil {
+		return MetricResult{labels: labels, value: pb.GetUntyped().GetValue(), metricType: dto.MetricType_UNTYPED}
+	}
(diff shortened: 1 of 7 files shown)


SuperQ:
@roman-vynar This is the first step in refactoring all of the individual collectors out of the big main.go. :-)

roman-vynar:
Ben, awesome! 

And the same we can implement for tests. Too much scrolling :)

roman-vynar:
Ah, I see a test is also here, cool.

collector/collector_test.go
```diff
@@ -0,0 +1,42 @@
+package collector
+
+import (
+	"strings"
+
+	"github.com/prometheus/client_golang/prometheus"
+	dto "github.com/prometheus/client_model/go"
+)
+
+type LabelMap map[string]string
```
juliusv: Btw. (not really about this refactoring, so maybe do it later), there's https://github.com/prometheus/common/blob/master/model/labelset.go#L28 already. Even in case you don't want to import that package just for that type, maybe call it `LabelSet` instead of `LabelMap` to be consistent.

juliusv: Also, if you do your own type here, it shouldn't be exported, since it's only used in tests in the same package.

mysqld_exporter.go
```diff
@@ -56,10 +58,6 @@ var (
 		"collect.info_schema.innodb_metrics", false,
 		"Collect metrics from information_schema.innodb_metrics",
 	)
-	collectGlobalStatus = flag.Bool(
```
juliusv: Usually you'll want to confine any flag definitions to the binary package, and optimally to its main function. I'd keep those in this package.

SuperQ: This seems to be the way it's done in node_exporter. :shrug:

juliusv: There's different opinions on whether that should even be like that there, but I'd see these flags here as different. They are more about enabling/disabling a collector completely rather than doing some detailed internal config of it. That is also done from main in the node exporter.

juliusv:
:+1: in general otherwise. I didn't look in detail at the moved code, because I assume it's just being moved.

mysqld_exporter.go
brian-brazil: It really boils down to whether there's someone using this code as a library that may want a different operational mode.

If it's just a random tuning variable, inside the file itself is fine. The node exporter ones fall under this, as plumbing them all into main wouldn't make sense, and noone is using it as a library.

Prometheus has gone too far in the other direction and plumbed through everything to main, which just makes it harder to add flags. This has also lead to us having more hardcoded constants than we should have.

SuperQ:
Yup, moved code is moved.

mysqld_exporter.go
SuperQ: Yup, makes sense.

For the collector specific tuning flags, I will be moving those into the collector package.

PR #115, opened 2016-04-14
Add special handling of "buffer_page_io" subsystem

SuperQ:
For information_schema.innodb_metrics we can split out labels based on
the type of buffer page IO operation in the "buffer_page_io" subsystem.
* Simplify scrapeInnodbMetrics() by casting value to float64.

### DIFF
--- a/collector/info_schema_innodb_metrics.go
+++ b/collector/info_schema_innodb_metrics.go
@@ -4,8 +4,10 @@ package collector
 
 import (
 	"database/sql"
+	"regexp"
 
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/common/log"
 )
 
 const infoSchemaInnodbMetricsQuery = `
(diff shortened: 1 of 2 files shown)


collector/info_schema_innodb_metrics.go
```diff
@@ -35,6 +53,21 @@ func ScrapeInnodbMetrics(db *sql.DB, ch chan<- prometheus.Metric) error {
 		); err != nil {
 			return err
 		}
+		// Special handling of the "buffer_page_io" subsystem.
+		if subsystem == "buffer_page_io" {
+			match := bufferPageRE.FindStringSubmatch(name)
+			switch match[1] {
```
juliusv: Check before this that `len(match)` is really at least 3 (and maybe log an error if not?). Otherwise the whole exporter will just crash in case this is ever not true.

juliusv:
:+1: otherwise

collector/info_schema_innodb_metrics.go
SuperQ: Wouldn't you want `if len(match) != 2` to detect that the regexp failed?  It looks lke [FindStringSubmatch](https://golang.org/pkg/regexp/#Regexp.FindStringSubmatch) returns a nil if the regexp fails.  So `len(match) != 2) would be a good choice.

SuperQ:
@juliusv Ok, I fixed the issue with invalid data from MySQL, and added a test for it.

juliusv:
Cool!

PR #116, opened 2016-04-15
Add collector for innodb_sys_tablespaces

SuperQ:
Implment colection of InnoDB tablespace metrics from
`information_schema.innodb_sys_tablespaces`

### DIFF
--- a/collector/info_schema_innodb_sys_tablespaces.go
+++ b/collector/info_schema_innodb_sys_tablespaces.go
@@ -1,0 +1,87 @@
+// Scrape `information_schema.innodb_sys_tablespaces`.
+
+package collector
+
+import (
+	"database/sql"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const innodbTablespacesQuery = `
+	SELECT
+	    SPACE,
+	    NAME,
+	    ifnull(FILE_FORMAT, 'NONE') as FILE_FORMAT,
+	    ifnull(ROW_FORMAT, 'NONE') as ROW_FORMAT,
+	    ifnull(SPACE_TYPE, 'NONE') as SPACE_TYPE,
+	    FILE_SIZE,
+	    ALLOCATED_SIZE
+	  FROM information_schema.innodb_sys_tablespaces
+	`
+
+var (
+	infoSchemaInnodbTablesspaceInfoDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, informationSchema, "innodb_tablespace_space_info"),
+		"The Tablespace information and Space ID.",
+		[]string{"tablespace_name", "file_format", "row_format", "space_type"}, nil,
+	)
+	infoSchemaInnodbTablesspaceFileSizeDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, informationSchema, "innodb_tablespace_file_size_bytes"),
+		"The apparent size of the file, which represents the maximum size of the file, uncompressed.",
+		[]string{"tablespace_name"}, nil,
+	)
+	infoSchemaInnodbTablesspaceAllocatedSizeDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, informationSchema, "innodb_tablespace_allocated_size_bytes"),
+		"The actual size of the file, which is the amount of space allocated on disk.",
+		[]string{"tablespace_name"}, nil,
+	)
+)
+
+// ScrapeInfoSchemaInnodbTablespaces collects from `information_schema.innodb_sys_tablespaces`.
+func ScrapeInfoSchemaInnodbTablespaces(db *sql.DB, ch chan<- prometheus.Metric) error {
+	tablespacesRows, err := db.Query(innodbTablespacesQuery)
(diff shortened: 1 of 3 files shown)


collector/info_schema_innodb_sys_tablespaces.go
```diff
@@ -0,0 +1,89 @@
...
+
+package collector
+
+import (
+	"database/sql"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
```
juliusv: Just one const, avoid block?

collector/info_schema_innodb_sys_tablespaces.go
```diff
@@ -0,0 +1,89 @@
...
+var (
+	infoSchemaInnodbTablesspaceInfoDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, informationSchema, "innodb_tablespace_space_info"),
+		"The Tablespace information and Space ID.",
+		[]string{"name", "file_format", "row_format", "space_type"}, nil,
+	)
+	infoSchemaInnodbTablesspaceFileSizeDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, informationSchema, "innodb_tablespace_file_size_bytes"),
+		"The apparent size of the file, which represents the maximum size of the file, uncompressed.",
+		[]string{"name"}, nil,
```
juliusv: "name" is a bad label because it's so generic and could mean anything. Is this the name of the table or what?

juliusv: Ah yeah, it is. I'd name it "table".

juliusv: Ok, the var name below was misleading, it's not a table name, but the tablespace name. Maybe "tablespace" then?

SuperQ: name is what it's called in the metrics table.  But, I think "tablespace" is probably better.

SuperQ: Actually, I think I will call it "tablespace_name" since mysql assigns both a tablespace space ID number and a name.

SuperQ:
@juliusv Updated with suggestions, PTAL.

collector/info_schema_innodb_sys_tablespaces.go
```diff
@@ -0,0 +1,87 @@
...
+
+var (
+	infoSchemaInnodbTablesspaceInfoDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, informationSchema, "innodb_tablespace_space_info"),
+		"The Tablespace information and Space ID.",
+		[]string{"tablespace_name", "file_format", "row_format", "space_type"}, nil,
+	)
+	infoSchemaInnodbTablesspaceFileSizeDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, informationSchema, "innodb_tablespace_file_size_bytes"),
+		"The apparent size of the file, which represents the maximum size of the file, uncompressed.",
```
juliusv: Should have caught this earlier... This docstring doesn't tell me anything without context. "of the file"? Which file?

juliusv: Similar for the others...

SuperQ: These are copy-n-paste from the upstream docs[0].  It makes sense for people familiar with MySQL.

[0]: https://dev.mysql.com/doc/refman/5.7/en/innodb-sys-tablespaces-table.html

juliusv: Hmm, ok. But it only makes sense in combination with the metric name. The doc string doesn't stand on its own even for mysql-familiar people?

Maybe that's how the docstrings in this exporter work though...

SuperQ: Yes, I've tried to make all the docs as close to the upstream docs as
possible.  Even tho they're not always obvious to newcomers.

MySQL upstream docs are pretty horrible.  I could easily rant about this.
:-)

I have the idea to write some better metrics guides for this exporter, but
need to find the time.

juliusv: In this case, the upstream docs seem fine, it's just that context is missing for this docstring. One wouldn't know that it's about tablespaces without also reading the metric name... but it's a minor wart and shouldn't block you.

:+1: in any case...

brunomvsouza:
This PR broke compatibility with MySQL 5.6 as INNODB_SYS_TABLESPACES columns changed dramatically [from 5.6](https://dev.mysql.com/doc/refman/5.6/en/innodb-sys-tablespaces-table.html) -> [to 5.7](https://dev.mysql.com/doc/refman/5.7/en/innodb-sys-tablespaces-table.html).

> time="2016-04-21T14:45:19-03:00" level=error msg="Error scraping for collect.info_schema.innodb_sys_tablespaces: Error 1054: Unknown column 'SPACE_TYPE' in 'field list'" source="mysqld_exporter.go:288"

SuperQ:
@brunomvsouza Doh!  I'll file an issue to deal with this.

PR #119, opened 2016-04-22
Fix collect.info_schema.innodb_tablespaces

SuperQ:
* Disable by default to avoid breaking with old MySQL versions
* Document flag as only compatible with >= 5.7


roman-vynar:
👍

PR #128, opened 2016-05-20
Soft fail with -collect.info_schema.innodb_tablespaces for MySQL prior 5.7

roman-vynar:
All our collectors are user-friendly across various versions of MySQL. However, `-collect.info_schema.innodb_tablespaces` is erroring on MySQL <5.7 which is expected. I suggest to soft fail on this deliberately expected behaviour so we are nice in the logs.

### DIFF
--- a/collector/info_schema_innodb_sys_tablespaces.go
+++ b/collector/info_schema_innodb_sys_tablespaces.go
@@ -6,6 +6,7 @@ import (
 	"database/sql"
 
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/common/log"
 )
 
 const innodbTablespacesQuery = `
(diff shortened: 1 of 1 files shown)


SuperQ:
All collectors soft-fail now, we fixed it a while back so individual collectors do not break the scrape.  There is a top level error counter that deals with broken collectors.

matthiasr:
I think it's right not to spew error messages all the time if there is nothing wrong. I'm not sure about checking the error message though, wouldn't it be cleaner to really gate on the MySQL version, as the debug message suggests?

SuperQ:
Yea, I think we should do a real MySQL version check.  This could be tricky tho.

* Break up the main scrape function into supported minimum versions
* Pass a version struct to each scraper to do their own version testing.

roman-vynar:
I agree, better would be to do version checking. Closing this one.

PR #136, opened 2016-06-16
Support client statistics

mfouilleul:
Same work that you did for "Users"...


SuperQ:
Cool, please add a test for this.  There are several examples that you can use.

Issue #148, opened 2016-08-01
Broken user/client statistics for Percona Server


roman-vynar:
This commit broke user statistics for Percona Server https://github.com/prometheus/mysqld_exporter/pull/138/files#diff-1cfdb3ffdb1892d4e0c3e4ea1ee8af1bL13

MariaDB has one set of columns https://mariadb.com/kb/en/mariadb/information-schema-user_statistics-table/
Percona Server has different https://www.percona.com/doc/percona-server/5.5/diagnostics/user_stats.html#USER_STATISTICS

Now it only works with MariaDB.

The same thing with client statistics.

SuperQ:
Tricky, it looks like there are a bunch of differences.  I guess we need to move this back to a `SELECT *` and handle all the variations.

roman-vynar:
Heh, it neither works properly with MariaDB, e.g. `ROWS_SENT` is not the map to evaluate https://github.com/prometheus/mysqld_exporter/pull/138/files#diff-1cfdb3ffdb1892d4e0c3e4ea1ee8af1bL54

So yeah, `SELECT *` would be better to go back.

PR #149, opened 2016-08-01
Improved and fixed user/client statistics

roman-vynar:
* Doing the same check as for Query Response Time: first check whether detailed stats available at all (presence of `@@userstat` var) and then whether that var is `ON`.
* Fixes https://github.com/prometheus/mysqld_exporter/issues/148 where table columns for MariaDB and Percona Server are different.
* Fixed tests. Not sure how come it was passed here https://circleci.com/gh/prometheus/mysqld_exporter/49 and https://travis-ci.org/prometheus/mysqld_exporter/jobs/140888221 We had UNTYPED but should have been COUNTER. The tests were passed but shouldn't have. Now they run fine locally.


SuperQ:
`SELECT @@userstat` will result in this not working for MySQL 5.1.  We also need to check the old name `userstat_running`.

Maybe `show global variables like 'userstat%'`?

roman-vynar:
Fixed.

SuperQ:
Awesome. :+1:

PR #152, opened 2016-08-16
Added the last binlog file number metric.

roman-vynar:
We have binlog count and total size, however it's impossible to measure how frequently the binlogs are rotated. The count does not tell this as it can be constantly on the same number.

I have added a new metric to expose what binlog file number currently is:
```
# HELP mysql_binlog_file_number The last binlog file number.
# TYPE mysql_binlog_file_number gauge
mysql_binlog_file_number 16
```
This stays for say `mysql-bin.000016`. The part after `.` is exposed as a value.

Also added the test.

### DIFF
--- a/collector/binlog.go
+++ b/collector/binlog.go
@@ -4,6 +4,8 @@ package collector
 
 import (
 	"database/sql"
+	"strconv"
+	"strings"
 
 	"github.com/prometheus/client_golang/prometheus"
 )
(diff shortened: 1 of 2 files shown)


SuperQ:
Somehow I thought we already had a metric for this.  Possibly more interesting is `mysql_global_status_binlog_snapshot_position`.  But :+1: this is useful.

collector/binlog.go
```diff
@@ -71,6 +78,11 @@ func ScrapeBinlogSize(db *sql.DB, ch chan<- prometheus.Metric) error {
 	ch <- prometheus.MustNewConstMetric(
 		binlogFilesDesc, prometheus.GaugeValue, float64(count),
 	)
+	// The last row contains the last binlog file number.
+	value, _ := strconv.ParseFloat(strings.Split(filename, ".")[1], 64)
+	ch <- prometheus.MustNewConstMetric(
+		binlogFileNumberDesc, prometheus.GaugeValue, value,
```
SuperQ: I think it's safe enough to assume this is a Counter.  I think in all cases MySQL will only reset to 0.

brian-brazil: It's a Gauge, as you care about the absolute value, it doesn't reset to 0 on process start and it can go down by operator action

SuperQ: :+1:

PR #156, opened 2016-08-26
Add an example recording rules file

SuperQ:
Add some basic recording and alerting rules.


RichiH:
As metric names tend to be Perl underscore notation, not CamlCase, it probably makes sense to maintain the same scheme in alert names; else +1.

SuperQ:
Alert names are done as CamlCase[0], don't ask me why.

[0]: https://prometheus.io/docs/alerting/rules/#templating

PR #158, opened 2016-09-05
Updated MySQL version supported for -collect.perf_schema.file_events

roman-vynar:
MySQL 5.5 gives this error when `-collect.perf_schema.file_events` enabled:
```
collect.perf_schema.file_events: Error 1054: Unknown column 'SUM_TIMER_READ' in 'field list'"...
```

It is actually 5.6 focused, not sure there is a reason to support it in 5.5.
Let's fix this in README.


SuperQ:
Hmm, I thought for sure I tested this one on 5.5, or we had already marked this one as 5.6 only.

There are some metrics available in 5.5, but I agree, probably not worth supporting at this time.

PR #79, opened 2016-02-11
Added query response time distribution

roman-vynar:
https://www.percona.com/doc/percona-server/5.6/diagnostics/response_time_distribution.html
https://mariadb.com/kb/en/mariadb/query_response_time-plugin/

Sample:
```
# HELP mysql_info_schema_query_response_time_count The number of queries according to the length of time they took to execute.
# TYPE mysql_info_schema_query_response_time_count histogram
mysql_info_schema_query_response_time_count_bucket{le="1e-06"} 281
mysql_info_schema_query_response_time_count_bucket{le="1e-05"} 756
mysql_info_schema_query_response_time_count_bucket{le="0.0001"} 8410
mysql_info_schema_query_response_time_count_bucket{le="0.001"} 11247
mysql_info_schema_query_response_time_count_bucket{le="0.01"} 11904
mysql_info_schema_query_response_time_count_bucket{le="0.1"} 11916
mysql_info_schema_query_response_time_count_bucket{le="1"} 11917
mysql_info_schema_query_response_time_count_bucket{le="10"} 11917
mysql_info_schema_query_response_time_count_bucket{le="100"} 11917
mysql_info_schema_query_response_time_count_bucket{le="1000"} 11917
mysql_info_schema_query_response_time_count_bucket{le="10000"} 11917
mysql_info_schema_query_response_time_count_bucket{le="100000"} 11917
mysql_info_schema_query_response_time_count_bucket{le="1e+06"} 11917
mysql_info_schema_query_response_time_count_bucket{le="+Inf"} 11917
mysql_info_schema_query_response_time_count_sum 3.002952
mysql_info_schema_query_response_time_count_count 11917
# HELP mysql_info_schema_query_response_time_total Total time of queries according to the length of time they took to execute separately.
# TYPE mysql_info_schema_query_response_time_total counter
mysql_info_schema_query_response_time_total{le="+Inf"} 3.002952
mysql_info_schema_query_response_time_total{le="0.0001"} 0.288892
mysql_info_schema_query_response_time_total{le="0.001"} 1.119809
mysql_info_schema_query_response_time_total{le="0.01"} 2.380771
mysql_info_schema_query_response_time_total{le="0.1"} 2.735583
mysql_info_schema_query_response_time_total{le="1"} 3.002952
mysql_info_schema_query_response_time_total{le="10"} 3.002952
mysql_info_schema_query_response_time_total{le="100"} 3.002952
mysql_info_schema_query_response_time_total{le="1000"} 3.002952
[...]

PR #166, opened 2016-10-11
Added read/write split to query response time distribution.

roman-vynar:
Adding read/write split to query response time distribution. Percona Server 5.6/5.7 supports that, PS 5.5 and MariaDB - only totals.

```
# HELP mysql_info_schema_query_response_time_seconds The number of all queries by duration they took to execute.
# TYPE mysql_info_schema_query_response_time_seconds histogram
mysql_info_schema_query_response_time_seconds_bucket{le="1e-06"} 33758
mysql_info_schema_query_response_time_seconds_bucket{le="1e-05"} 97717
mysql_info_schema_query_response_time_seconds_bucket{le="0.0001"} 520449
mysql_info_schema_query_response_time_seconds_bucket{le="0.001"} 701573
mysql_info_schema_query_response_time_seconds_bucket{le="0.01"} 795705
mysql_info_schema_query_response_time_seconds_bucket{le="0.1"} 805960
mysql_info_schema_query_response_time_seconds_bucket{le="1"} 806175
mysql_info_schema_query_response_time_seconds_bucket{le="10"} 806175
mysql_info_schema_query_response_time_seconds_bucket{le="100"} 806175
mysql_info_schema_query_response_time_seconds_bucket{le="1000"} 806175
mysql_info_schema_query_response_time_seconds_bucket{le="10000"} 806175
mysql_info_schema_query_response_time_seconds_bucket{le="100000"} 806175
mysql_info_schema_query_response_time_seconds_bucket{le="1e+06"} 806175
mysql_info_schema_query_response_time_seconds_bucket{le="+Inf"} 806175
mysql_info_schema_query_response_time_seconds_sum 628.913333
mysql_info_schema_query_response_time_seconds_count 806175
# HELP mysql_info_schema_read_query_response_time_seconds The number of read queries by duration they took to execute.
# TYPE mysql_info_schema_read_query_response_time_seconds histogram
mysql_info_schema_read_query_response_time_seconds_bucket{le="1e-06"} 33758
mysql_info_schema_read_query_response_time_seconds_bucket{le="1e-05"} 97717
mysql_info_schema_read_query_response_time_seconds_bucket{le="0.0001"} 520447
mysql_info_schema_read_query_response_time_seconds_bucket{le="0.001"} 701567
mysql_info_schema_read_query_response_time_seconds_bucket{le="0.01"} 795685
mysql_info_schema_read_query_response_time_seconds_bucket{le="0.1"} 805937
mysql_info_schema_read_query_response_time_seconds_bucket{le="1"} 806151
mysql_info_schema_read_query_response_time_seconds_bucket{le="10"} 806151
mysql_info_schema_read_query_response_time_seconds_bucket{le="100"} 806151
mysql_info_schema_read_query_response_time_seconds_bucket{le="1000"} 806151
mysql_info_schema_read_query_response_time_seconds_bucket{le="10000"} 806151
mysql_info_schema_read_query_response_time_seconds_bucket{le="100000"} 806151
mysql_info_schema_read_query_response_time_seconds_bucket{le="1e+06"} 806151
mysql_info_schema_read_query_response_time_seconds_bucket{le="+Inf"} 806151
mysql_info_schema_read_query_response_time_seconds_sum 628.600932
mysql_info_schema_read_query_response_time_seconds_count 806151
# HELP mysql_info_schema_write_query_response_time_seconds The number of write queries by duration they took to execute.
# TYPE mysql_info_schema_write_query_response_time_seconds histogram
mysql_info_schema_write_query_response_time_seconds_bucket{le="1e-06"} 0
mysql_info_schema_write_query_response_time_seconds_bucket{le="1e-05"} 0
mysql_info_schema_write_query_response_time_seconds_bucket{le="0.0001"} 3
mysql_info_schema_write_query_response_time_seconds_bucket{le="0.001"} 7
mysql_info_schema_write_query_response_time_seconds_bucket{le="0.01"} 21
mysql_info_schema_write_query_response_time_seconds_bucket{le="0.1"} 24
mysql_info_schema_write_query_response_time_seconds_bucket{le="1"} 25
mysql_info_schema_write_query_response_time_seconds_bucket{le="10"} 25
mysql_info_schema_write_query_response_time_seconds_bucket{le="100"} 25
mysql_info_schema_write_query_response_time_seconds_bucket{le="1000"} 25
mysql_info_schema_write_query_response_time_seconds_bucket{le="10000"} 25
mysql_info_schema_write_query_response_time_seconds_bucket{le="100000"} 25
mysql_info_schema_write_query_response_time_seconds_bucket{le="1e+06"} 25
mysql_info_schema_write_query_response_time_seconds_bucket{le="+Inf"} 25
mysql_info_schema_write_query_response_time_seconds_sum 0.312432
mysql_info_schema_write_query_response_time_seconds_count 25
```

### DIFF
--- a/collector/info_schema_query_response_time.go
+++ b/collector/info_schema_query_response_time.go
@@ -24,17 +24,17 @@ var (
 
 	infoSchemaQueryResponseTimeCountDescs = [3]*prometheus.Desc{
 		prometheus.NewDesc(
-			prometheus.BuildFQName(namespace, informationSchema, "query_response_time"),
+			prometheus.BuildFQName(namespace, informationSchema, "query_response_time_seconds"),
 			"The number of all queries by duration they took to execute.",
 			[]string{}, nil,
 		),
 		prometheus.NewDesc(
-			prometheus.BuildFQName(namespace, informationSchema, "read_query_response_time"),
+			prometheus.BuildFQName(namespace, informationSchema, "read_query_response_time_seconds"),
 			"The number of read queries by duration they took to execute.",
 			[]string{}, nil,
 		),
 		prometheus.NewDesc(
-			prometheus.BuildFQName(namespace, informationSchema, "write_query_response_time"),
+			prometheus.BuildFQName(namespace, informationSchema, "write_query_response_time_seconds"),
 			"The number of write queries by duration they took to execute.",
 			[]string{}, nil,
 		),
--- a/collector/info_schema_query_response_time.go
+++ b/collector/info_schema_query_response_time.go
@@ -4,7 +4,6 @@ package collector
 
 import (
 	"database/sql"
-	"fmt"
 	"strconv"
 	"strings"
 
@@ -25,36 +24,19 @@ var (
 
 	infoSchemaQueryResponseTimeCountDescs = [3]*prometheus.Desc{
 		prometheus.NewDesc(
-			prometheus.BuildFQName(namespace, informationSchema, "query_response_time_count"),
+			prometheus.BuildFQName(namespace, informationSchema, "query_response_time"),
 			"The number of all queries by duration they took to execute.",
 			[]string{}, nil,
 		),
 		prometheus.NewDesc(
-			prometheus.BuildFQName(namespace, informationSchema, "query_response_time_read_count"),
+			prometheus.BuildFQName(namespace, informationSchema, "read_query_response_time"),
 			"The number of read queries by duration they took to execute.",
 			[]string{}, nil,
 		),
 		prometheus.NewDesc(
-			prometheus.BuildFQName(namespace, informationSchema, "query_response_time_write_count"),
+			prometheus.BuildFQName(namespace, informationSchema, "write_query_response_time"),
 			"The number of write queries by duration they took to execute.",
 			[]string{}, nil,
-		),
-	}
-	infoSchemaQueryResponseTimeTotalDescs = [3]*prometheus.Desc{
-		prometheus.NewDesc(
-			prometheus.BuildFQName(namespace, informationSchema, "query_response_time_total"),
-			"Total time of all queries by duration they took to execute.",
-			[]string{"le"}, nil,
-		),
-		prometheus.NewDesc(
-			prometheus.BuildFQName(namespace, informationSchema, "query_response_time_read_total"),
-			"Total time of read queries by duration they took to execute.",
-			[]string{"le"}, nil,
-		),
-		prometheus.NewDesc(
-			prometheus.BuildFQName(namespace, informationSchema, "query_response_time_write_total"),
-			"Total time of write queries by duration they took to execute.",
-			[]string{"le"}, nil,
 		),
 	}
 )
@@ -95,16 +77,7 @@ func processQueryResponseTimeTable(db *sql.DB, ch chan<- prometheus.Metric, quer
 			continue
 		}
 		countBuckets[length] = histogramCnt
-		// No histogram with query total times because they are float
-		ch <- prometheus.MustNewConstMetric(
-			infoSchemaQueryResponseTimeTotalDescs[i], prometheus.CounterValue, histogramSum,
-			fmt.Sprintf("%v", length),
-		)
 	}
-	ch <- prometheus.MustNewConstMetric(
-		infoSchemaQueryResponseTimeTotalDescs[i], prometheus.CounterValue, histogramSum,
-		"+Inf",
-	)
 	// Create histogram with query counts
 	ch <- prometheus.MustNewConstHistogram(
 		infoSchemaQueryResponseTimeCountDescs[i], histogramCnt, histogramSum, countBuckets,
@@ -127,7 +100,7 @@ func ScrapeQueryResponseTime(db *sql.DB, ch chan<- prometheus.Metric) error {
 	for i, query := range queryResponseTimeQueries {
 		err := processQueryResponseTimeTable(db, ch, query, i)
 		// The first query should not fail if query_response_time_stats is ON,
-		// unlike the other two when the tables exist only with Percona Server.
+		// unlike the other two when the read/write tables exist only with Percona Server 5.6/5.7.
 		if i == 0 && err != nil {
 			return err
 		}
--- a/collector/info_schema_query_response_time.go
+++ b/collector/info_schema_query_response_time.go
@@ -1,4 +1,4 @@
-// Scrape `information_schema.query_response_time`.
+// Scrape `information_schema.query_response_time*` tables.
 
 package collector
 
@@ -12,42 +12,55 @@ import (
 	"github.com/prometheus/common/log"
 )
 
-const (
-	queryResponseCheckQuery = `SELECT @@query_response_time_stats`
-	queryResponseTimeQuery  = `
-		SELECT
-		    TIME, COUNT, TOTAL
-		  FROM information_schema.query_response_time
-		`
-)
+const queryResponseCheckQuery = `SELECT @@query_response_time_stats`
 
 var (
-	infoSchemaQueryResponseTimeCountDesc = prometheus.NewDesc(
-		prometheus.BuildFQName(namespace, informationSchema, "query_response_time_count"),
-		"The number of queries according to the length of time they took to execute.",
-		[]string{}, nil,
-	)
-	infoSchemaQueryResponseTimeTotalDesc = prometheus.NewDesc(
-		prometheus.BuildFQName(namespace, informationSchema, "query_response_time_total"),
-		"Total time of queries according to the length of time they took to execute separately.",
-		[]string{"le"}, nil,
-	)
+	// Use uppercase for table names, otherwise read/write split will return the same results as total
+	// due to the bug.
+	queryResponseTimeQueries = [3]string{
+		"SELECT TIME, COUNT, TOTAL FROM INFORMATION_SCHEMA.QUERY_RESPONSE_TIME",
+		"SELECT TIME, COUNT, TOTAL FROM INFORMATION_SCHEMA.QUERY_RESPONSE_TIME_READ",
+		"SELECT TIME, COUNT, TOTAL FROM INFORMATION_SCHEMA.QUERY_RESPONSE_TIME_WRITE",
+	}
+
+	infoSchemaQueryResponseTimeCountDescs = [3]*prometheus.Desc{
+		prometheus.NewDesc(
+			prometheus.BuildFQName(namespace, informationSchema, "query_response_time_count"),
+			"The number of all queries by duration they took to execute.",
+			[]string{}, nil,
+		),
+		prometheus.NewDesc(
+			prometheus.BuildFQName(namespace, informationSchema, "query_response_time_read_count"),
+			"The number of read queries by duration they took to execute.",
+			[]string{}, nil,
+		),
+		prometheus.NewDesc(
+			prometheus.BuildFQName(namespace, informationSchema, "query_response_time_write_count"),
+			"The number of write queries by duration they took to execute.",
+			[]string{}, nil,
+		),
+	}
+	infoSchemaQueryResponseTimeTotalDescs = [3]*prometheus.Desc{
+		prometheus.NewDesc(
+			prometheus.BuildFQName(namespace, informationSchema, "query_response_time_total"),
+			"Total time of all queries by duration they took to execute.",
+			[]string{"le"}, nil,
+		),
+		prometheus.NewDesc(
+			prometheus.BuildFQName(namespace, informationSchema, "query_response_time_read_total"),
+			"Total time of read queries by duration they took to execute.",
+			[]string{"le"}, nil,
+		),
+		prometheus.NewDesc(
+			prometheus.BuildFQName(namespace, informationSchema, "query_response_time_write_total"),
+			"Total time of write queries by duration they took to execute.",
+			[]string{"le"}, nil,
+		),
+	}
 )
 
-// ScrapeQueryResponseTime collects from `information_schema.query_response_time`.
-func ScrapeQueryResponseTime(db *sql.DB, ch chan<- prometheus.Metric) error {
-	var queryStats uint8
-	err := db.QueryRow(queryResponseCheckQuery).Scan(&queryStats)
-	if err != nil {
-		log.Debugln("Query response time distribution is not present.")
-		return nil
-	}
-	if queryStats == 0 {
-		log.Debugln("MySQL @@query_response_time_stats is OFF.")
-		return nil
-	}
-
-	queryDistributionRows, err := db.Query(queryResponseTimeQuery)
+func processQueryResponseTimeTable(db *sql.DB, ch chan<- prometheus.Metric, query string, i int) error {
+	queryDistributionRows, err := db.Query(query)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


SuperQ:
Hmm, would it make sense to not include the response time totals when we have read+write data?

roman-vynar:
Only totals are available for PS 5.5 and MariaDB. Totals+read/write split - for PS 5.6, 5.7. 
That's why we still have to include them.

SuperQ:
I'm saying, we could detect if there is a split or not, and change the output based on that.

roman-vynar:
In this case, we won't be consistent and the promql queries will turn into `(read + write) or total` which could be way too long with those histograms and `le` label.

SuperQ:
Hmm, yea, messy.

SuperQ:
@beorn7, master of histograms, mind taking a quick look at this? :-)

beorn7:
On my queue…

collector/info_schema_query_response_time.go
```diff
@@ -12,42 +12,55 @@ import (
...
-// ScrapeQueryResponseTime collects from `information_schema.query_response_time`.
-func ScrapeQueryResponseTime(db *sql.DB, ch chan<- prometheus.Metric) error {
-	var queryStats uint8
-	err := db.QueryRow(queryResponseCheckQuery).Scan(&queryStats)
-	if err != nil {
-		log.Debugln("Query response time distribution is not present.")
-		return nil
+	infoSchemaQueryResponseTimeCountDescs = [3]*prometheus.Desc{
+		prometheus.NewDesc(
+			prometheus.BuildFQName(namespace, informationSchema, "query_response_time_count"),
```
beorn7: Why the `_count` suffix? It doesn't really add any information (does it?), and it collides with the automatically added `_count` suffix in the histogram (resulting in `_count_count`).

beorn7: Also, the unit should be part of the metric name.

So how about `query_response_time_seconds`?

collector/info_schema_query_response_time.go
```diff
@@ -12,42 +12,55 @@ import (
...
+			"The number of all queries by duration they took to execute.",
+			[]string{}, nil,
+		),
+		prometheus.NewDesc(
+			prometheus.BuildFQName(namespace, informationSchema, "query_response_time_read_count"),
```
beorn7: Same as above, I wouldn't use `_count`. And how about `read_query_response_time_seconds`?

collector/info_schema_query_response_time.go
```diff
@@ -12,42 +12,55 @@ import (
...
+			"The number of read queries by duration they took to execute.",
+			[]string{}, nil,
+		),
+		prometheus.NewDesc(
+			prometheus.BuildFQName(namespace, informationSchema, "query_response_time_write_count"),
```
beorn7: Suggestion: `write_query_response_time_seconds`

collector/info_schema_query_response_time.go
```diff
@@ -12,42 +12,55 @@ import (
...
+			"The number of write queries by duration they took to execute.",
+			[]string{}, nil,
+		),
 	}
-	if queryStats == 0 {
-		log.Debugln("MySQL @@query_response_time_stats is OFF.")
-		return nil
+	infoSchemaQueryResponseTimeTotalDescs = [3]*prometheus.Desc{
+		prometheus.NewDesc(
+			prometheus.BuildFQName(namespace, informationSchema, "query_response_time_total"),
```
beorn7: A `_total` suffix strongly suggests this is a counter, but it is an (emulated) histogram.

How about `total_query_response_time_seconds`?

beorn7: OK, sorry, you seem to mean this as a counter, indeed.

I have to look more closely to even understand what the intention is here.

collector/info_schema_query_response_time.go
roman-vynar: We have 2 things here: sums (query durations by length) and counts (number of queries by individual query duration).

Counts is a histogram, sums is a pseudo-histogram because we can't put float into bucket. More background on this is here https://github.com/prometheus/mysqld_exporter/pull/79

It colides around `_count_count` because we have count of all counters which represent the actual histogram :D lol

collector/info_schema_query_response_time.go
beorn7: OK, got it.

Then, the suggested name would be, in agreement with above (add the unit):
`query_response_time_seconds_total`

collector/info_schema_query_response_time.go
```diff
@@ -12,42 +12,55 @@ import (
...
+			"Total time of all queries by duration they took to execute.",
+			[]string{"le"}, nil,
+		),
+		prometheus.NewDesc(
+			prometheus.BuildFQName(namespace, informationSchema, "query_response_time_read_total"),
```
beorn7: Suggestion: `read_query_response_time_seconds_total`

collector/info_schema_query_response_time.go
roman-vynar: About `query_response_time_seconds` - it's not seconds, it's `The number of all queries by duration they took to execute`.

beorn7:
Apparently, the naming has already been passed in #79 . I just find it weird, but I guess this PR is not the right place to change it.

roman-vynar:
@beorn7 looking again into this, may be it makes sense to do those renaming:

For histogram:
`query_response_time_count` > `query_response_time_count`
`query_response_time_read_count` > `read_query_response_time_count`
`query_response_time_write_count` > `write_query_response_time_count`
Not sure there could be better word for `_count` to avoid `_count_count`.

For counters:
`query_response_time_total` > `query_response_time_seconds`
`query_response_time_read_total` > `read_query_response_time_seconds`
`query_response_time_write_total` > `write_query_response_time_seconds`
Despite `_total` is still good it will be less confusing with the above histogram.

beorn7:
Counters should end on `_total`. That's a very strong convention.

But I don't really get the need for `_count` for the histogram. A histogram`s buckets are implicitly counting the things in the bucket. So the fact that they end on `_bucket` already implies that.

The distinction between the "real" histogram and the bucketed counters of response time is actually quite aptly done with the `_total` suffix vs `_bucket`.

roman-vynar:
`_count` tells that we have number (count) of queries in the bucket. W/o suffix it's not clear whether we have number of queries or something else. In our case something else is a pseudo histogram - a set of counters representing sum of query duration by their individual length.

In other words, it won't be clear what is `query_response_time_(bucket|count|total)` and what is `query_response_time_total`.

How about this renaming?

For histogram:
`query_response_time_count` > `query_response_time` (_bucket, _total, _count imply implicitly).

For counters:
`query_response_time_total` > `query_response_time_secs_total`

And prepend read_, write_ accordingly to query_ for read/write split.

beorn7:
Disclaimer: I didn't really want to start a fundamental discussion about the naming. If you want to keep it as it is to not break users, and come back to a thorough revisit of metric names at some later point, that's totally fine with me.

Having said that:

Your latest suggestion is quite close to the conventions, with the following deviations:

Convention is to always have the unit in the metric name, and to spell it out, i.e. `_seconds` > `_sec` > nothing. And for a histogram, the unit applies to the bucket boundaries and to the value of the implicit `…_sum` time series (but not to the values of the bucket time series and not to the `…._count` time series, those are just (integer) counters).

So call the "real" histogram `query_response_time_seconds` (so that the time series visible in the text format and on the Prometheus server will be called `query_response_time_seconds_(bucket|count|sum)`).

Call the "bucketed counters" `query_response_time_seconds_total`.

And everything with `(read|write)_` prefix for the split as you said.

The "bucketed counters" are a bit weird[1], to be honest, but as I understand, that's the way those metrics are organized internally in MySQL, so let's go with it. The weirdness of the approach just explains why there is no clear naming strategy in Prometheus for it. I think, with the suggestion above we are sticking to all the conventions, while still keeping the necessary expressiveness to mark what's going on.

[1] Why should I be interested in the average response time of all responses that took between 1s and 10s? Obviously, it will be somewhere between 1s and 10s… If I'm interested in the distribution within that bucket, I should use a finer bucketing scheme in the histogram and not come up with this arcane meta-histogram… whatever, just my opinion.

brian-brazil:
Considering the counters are emulating a histogram, I think they should be named accordingly with the `_bucket` suffix. Having them otherwise hints that a sum/average of them would make sense - which it doesn't.

beorn7:
The counters are not emulating a histogram. It's a weird contraption. Letting it appear as a histogram would be confusing.
Obviously, it's also confusing that you cannot sum or average the counters.
Perhaps they should not be cumulative. In that case, they could be summed.

brian-brazil:
In what way is it not the buckets of a histogram?

beorn7:
A histogram bucket counts how often an observation falls into that bucket. The pseudo histograms in MySQL "weigh" observations by their value, so to speak. I don't know if there is a name for that in the literature. I also don't really know what it is good for. It is definitely not a histogram as we know it.

brian-brazil:
Ah right, that'd be a counter then as you can sum across the labels.

beorn7:
But only if the counters are not cumulative. As far as I can see, the current implementation creates cumulative counts, with the likely intention to mimic the Prometheus histogram.

However, I think, the only reason for these "bucketed response time counters" is to plot average latency of "fast" or "slow" queries separately. For that purpose, cumulative counts are not helpful. Thus, I guess, they should not be cumulative, and they should have a label for the lower bound, too. Perhaps it would be good to not use the notorious `le` label to avoid confusion with a "real" histogram. More like `lower_bound` and `upper_bound` or something. (The reason why we picked a difficult-to-read label name like `le` was that it occurs so copiously in histograms that it should be really short.)

roman-vynar:
Hmm, we can get rid of those counters ending with `_total`. I have created them initially thinking it will be useful, but looks like it's not. We don't really need `Total time of all queries by duration they took to execute.` - noone probably cares how many time queries with say duration 1-10s took overall (obviously (1-10s or ~5s avg) * count of them). We have their counts in the real histogram.

Our use cases
https://pmmdemo.percona.com/graph/dashboard/db/mysql-query-response-time

Average Query Response Time
```
rate(mysql_info_schema_query_response_time_count_sum{instance="$host"}[$interval]) / 
rate(mysql_info_schema_query_response_time_count_count{instance="$host"}[$interval]) * 1000 or 
irate(mysql_info_schema_query_response_time_count_sum{instance="$host"}[5m]) / 
irate(mysql_info_schema_query_response_time_count_count{instance="$host"}[5m]) * 1000
```

Query Response Time Distribution
Queries 100ms - 1s
```
(rate(mysql_info_schema_query_response_time_count_bucket{instance="$host",le='1'}[$interval]) 
- on (instance) 
rate(mysql_info_schema_query_response_time_count_bucket{instance="$host",le='0.1'}[$interval])) 
or (irate(mysql_info_schema_query_response_time_count_bucket{instance="$host",le='1'}[5m]) 
- on (instance) 
irate(mysql_info_schema_query_response_time_count_bucket{instance="$host",le='0.1'}[5m]))
```

Another suggestions:
* remove counters `_total` (fake histogram)
* rename `query_response_time_count` > `query_response_time` for histogram
* put read/write as a prefix

How about this?

roman-vynar:
Is it better now?

```
# HELP mysql_info_schema_query_response_time The number of all queries by duration they took to execute.
# TYPE mysql_info_schema_query_response_time histogram
mysql_info_schema_query_response_time_bucket{le="1e-06"} 32631
mysql_info_schema_query_response_time_bucket{le="1e-05"} 94578
mysql_info_schema_query_response_time_bucket{le="0.0001"} 503169
mysql_info_schema_query_response_time_bucket{le="0.001"} 679067
mysql_info_schema_query_response_time_bucket{le="0.01"} 770551
mysql_info_schema_query_response_time_bucket{le="0.1"} 780566
mysql_info_schema_query_response_time_bucket{le="1"} 780774
mysql_info_schema_query_response_time_bucket{le="10"} 780774
mysql_info_schema_query_response_time_bucket{le="100"} 780774
mysql_info_schema_query_response_time_bucket{le="1000"} 780774
mysql_info_schema_query_response_time_bucket{le="10000"} 780774
mysql_info_schema_query_response_time_bucket{le="100000"} 780774
mysql_info_schema_query_response_time_bucket{le="1e+06"} 780774
mysql_info_schema_query_response_time_bucket{le="+Inf"} 780774
mysql_info_schema_query_response_time_sum 611.6681330000001
mysql_info_schema_query_response_time_count 780774
# HELP mysql_info_schema_read_query_response_time The number of read queries by duration they took to execute.
# TYPE mysql_info_schema_read_query_response_time histogram
mysql_info_schema_read_query_response_time_bucket{le="1e-06"} 32631
mysql_info_schema_read_query_response_time_bucket{le="1e-05"} 94578
mysql_info_schema_read_query_response_time_bucket{le="0.0001"} 503167
mysql_info_schema_read_query_response_time_bucket{le="0.001"} 679061
mysql_info_schema_read_query_response_time_bucket{le="0.01"} 770531
mysql_info_schema_read_query_response_time_bucket{le="0.1"} 780543
mysql_info_schema_read_query_response_time_bucket{le="1"} 780750
mysql_info_schema_read_query_response_time_bucket{le="10"} 780750
mysql_info_schema_read_query_response_time_bucket{le="100"} 780750
mysql_info_schema_read_query_response_time_bucket{le="1000"} 780750
mysql_info_schema_read_query_response_time_bucket{le="10000"} 780750
mysql_info_schema_read_query_response_time_bucket{le="100000"} 780750
mysql_info_schema_read_query_response_time_bucket{le="1e+06"} 780750
mysql_info_schema_read_query_response_time_bucket{le="+Inf"} 780750
mysql_info_schema_read_query_response_time_sum 611.355733
mysql_info_schema_read_query_response_time_count 780750
# HELP mysql_info_schema_write_query_response_time The number of write queries by duration they took to execute.
# TYPE mysql_info_schema_write_query_response_time histogram
mysql_info_schema_write_query_response_time_bucket{le="1e-06"} 0
mysql_info_schema_write_query_response_time_bucket{le="1e-05"} 0
mysql_info_schema_write_query_response_time_bucket{le="0.0001"} 3
mysql_info_schema_write_query_response_time_bucket{le="0.001"} 7
mysql_info_schema_write_query_response_time_bucket{le="0.01"} 21
mysql_info_schema_write_query_response_time_bucket{le="0.1"} 24
mysql_info_schema_write_query_response_time_bucket{le="1"} 25
mysql_info_schema_write_query_response_time_bucket{le="10"} 25
mysql_info_schema_write_query_response_time_bucket{le="100"} 25
mysql_info_schema_write_query_response_time_bucket{le="1000"} 25
mysql_info_schema_write_query_response_time_bucket{le="10000"} 25
mysql_info_schema_write_query_response_time_bucket{le="100000"} 25
mysql_info_schema_write_query_response_time_bucket{le="1e+06"} 25
mysql_info_schema_write_query_response_time_bucket{le="+Inf"} 25
mysql_info_schema_write_query_response_time_sum 0.312432
mysql_info_schema_write_query_response_time_count 25
```

beorn7:
Much better. Just the unit is missing now, i.e.
`…_response_time` → `…_response_time_seconds`

roman-vynar:
Done. The first message output is updated.

collector/info_schema_query_response_time.go
```diff
@@ -1,53 +1,48 @@
...
-// ScrapeQueryResponseTime collects from `information_schema.query_response_time`.
-func ScrapeQueryResponseTime(db *sql.DB, ch chan<- prometheus.Metric) error {
-	var queryStats uint8
-	err := db.QueryRow(queryResponseCheckQuery).Scan(&queryStats)
-	if err != nil {
-		log.Debugln("Query response time distribution is not present.")
-		return nil
+	// Use uppercase for table names, otherwise read/write split will return the same results as total
+	// due to the bug.
+	queryResponseTimeQueries = [3]string{
```
beorn7: Just a note: Using arrays is somewhat weird in Go. You would only do that where it is required for performance reasons. The performance overhead for using a slice here is most likely not noticeable.

beorn7:
Cool. :+1: 
@SuperQ for final review.

collector/info_schema_query_response_time.go
roman-vynar: 👍

SuperQ:
@beorn7 Thanks, :+1:

PR #169, opened 2016-11-10
Add mysql_galera_gcache_size_bytes metric.

roman-vynar:
Added `gcache.size` as a metric from `wsrep_provider_options` global variable for PXC/Galera.

This is useful to compare with the rate of `sum(wsrep%bytes)` status variables to figure out how much time a node can be out of the cluster before it will require SST vs IST. [Details](http://severalnines.com/blog/how-avoid-sst-when-adding-new-node-galera-cluster)

### DIFF
--- a/collector/global_variables.go
+++ b/collector/global_variables.go
@@ -4,6 +4,8 @@ package collector
 
 import (
 	"database/sql"
+	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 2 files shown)


roman-vynar:
@SuperQ Thanks.

Then you can graph or calculate available downtime before SST is required:
```mysql_galera_gcache_size_bytes{instance="$host"}/(rate(mysql_global_status_wsrep_replicated_bytes{instance="$host"}[5m])+rate(mysql_global_status_wsrep_received_bytes{instance="$host"}[5m])) * 300```

SuperQ:
What is the `* 300` for?  The calculation looks like it would return the number of seconds in the cache.  I would think you would want to divide the calculation by `60` or `3600` to return the number of minutes or hours available in the cache.

roman-vynar:
This is for [the graph](https://pmmdemo.percona.com/graph/dashboard/db/pxc-galera-graphs?panelId=47&fullscreen): I was using hardcoded 5m in the Interval field on Grafana, so it's data point duration for which available downtime is calculated, thus x 300s.

PR #172, opened 2016-11-21
collector: fix 32bit int overflow in tests

filippog:
See also the build failures for i386 at https://buildd.debian.org/status/package.php?p=prometheus-mysqld-exporter#problem-3

The fix is trivial but I was surprised it happened at all since mysqld_exporter is already built/released for i386 by promu. Any idea why this was the case?

### DIFF
--- a/collector/info_schema_clientstats_test.go
+++ b/collector/info_schema_clientstats_test.go
@@ -21,7 +21,7 @@ func TestScrapeClientStat(t *testing.T) {
 
 	columns := []string{"CLIENT", "TOTAL_CONNECTIONS", "CONCURRENT_CONNECTIONS", "CONNECTED_TIME", "BUSY_TIME", "CPU_TIME", "BYTES_RECEIVED", "BYTES_SENT", "BINLOG_BYTES_WRITTEN", "ROWS_READ", "ROWS_SENT", "ROWS_DELETED", "ROWS_INSERTED", "ROWS_UPDATED", "SELECT_COMMANDS", "UPDATE_COMMANDS", "OTHER_COMMANDS", "COMMIT_TRANSACTIONS", "ROLLBACK_TRANSACTIONS", "DENIED_CONNECTIONS", "LOST_CONNECTIONS", "ACCESS_DENIED", "EMPTY_QUERIES"}
 	rows := sqlmock.NewRows(columns).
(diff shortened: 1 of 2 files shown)


SuperQ:
Hrm, I don't know why this would happen.  The columns in the information schema table are specified as `bigint(21) unsigned` so 64-bit values must be supported.

I'm not sure this is the correct fix.

filippog:
I'm reporting the error here too for clarity:

```
# github.com/prometheus/mysqld_exporter/collector
src/github.com/prometheus/mysqld_exporter/collector/info_schema_clientstats_test.go:24: constant 2565104853 overflows int
src/github.com/prometheus/mysqld_exporter/collector/info_schema_clientstats_test.go:24: constant 2380108042 overflows int
src/github.com/prometheus/mysqld_exporter/collector/info_schema_userstats_test.go:24: constant 2565104853 overflows int
src/github.com/prometheus/mysqld_exporter/collector/info_schema_userstats_test.go:24: constant 2380108042 overflows int
FAIL	github.com/prometheus/mysqld_exporter/collector [build failed]
```

I think the problem here is with go numeric constant for integers, defaulting to `int` and thus overflowing on 32bit archs. I agree my proposed fix might not be the right fix, I can open a related issue for it too if that's better.

SuperQ:
What if you wrap the constant in `unit64()`?

filippog:
@SuperQ `uint64()` barfs when adding the value to `expected`, though `float64()` works for me on i386

```
# github.com/prometheus/mysqld_exporter/collector
src/github.com/prometheus/mysqld_exporter/collector/info_schema_clientstats_test.go:41: cannot use uint64(2565104853) (type uint64) as type float64 in field value
src/github.com/prometheus/mysqld_exporter/collector/info_schema_clientstats_test.go:43: cannot use uint64(2380108042) (type uint64) as type float64 in field value
src/github.com/prometheus/mysqld_exporter/collector/info_schema_userstats_test.go:41: cannot use uint64(2565104853) (type uint64) as type float64 in field value
src/github.com/prometheus/mysqld_exporter/collector/info_schema_userstats_test.go:43: cannot use uint64(2380108042) (type uint64) as type float64 in field value
FAIL    github.com/prometheus/mysqld_exporter/collector [build failed]
```

SuperQ:
`float64()` is actually fine, as that's the expected output format anyway.

filippog:
@SuperQ should be fixed with `float64()`, PTAL!

PR #176, opened 2016-12-13
Update example.rules

lswith:


SuperQ:
This does not take into account `mysql_slave_status_sql_delay`.  See the recording rule at the top of the file.

PR #183, opened 2017-02-20
Implement heartbeat metrics

roidelapluie:
pt-heartbeat measures actual replication lab on a mysql server. It does
not rely on data stored in SHOW SLAVE STATUS, as it can be unreliable.

pt-heartbeat website:
https://www.percona.com/doc/percona-toolkit/2.2/pt-heartbeat.html

Signed-off-by: Julien Pivotto <roidelapluie@inuits.eu>

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,9 @@ collect.perf_schema.indexiowaits                       | 5.6           | Collect
 collect.perf_schema.tableiowaits                       | 5.6           | Collect metrics from performance_schema.table_io_waits_summary_by_table.
 collect.perf_schema.tablelocks                         | 5.6           | Collect metrics from performance_schema.table_lock_waits_summary_by_table.
 collect.slave_status                                   | 5.1           | Collect from SHOW SLAVE STATUS (Enabled by default)
+collect.heartbeat                                      | 5.1           | Collect from [heartbeat](#heartbeat).
+collect.heartbeat.database                             | 5.1           | Database from where to collect heartbeat data. (default: heartbeat)
+collect.heartbeat.table                                | 5.1           | Table from where to collect heartbeat data. (default: heartbeat)
 
 
 ### General Flags
@@ -96,6 +99,14 @@ docker pull prom/mysqld-exporter
 docker run -d -p 9104:9104 --link=my_mysql_container:bdd  \
         -e DATA_SOURCE_NAME="user:password@(bdd:3306)/database" prom/mysqld-exporter
 ```
+
+## heartbeat
+
+With `collect.heartbeat` enabled, mysqld_exporter will scrape replication delay
+measured by heartbeat mechanisms. [Pt-heartbeat][pth] is the
+reference heartbeat implementation supported.
+
+[pth]:https://www.percona.com/doc/percona-toolkit/2.2/pt-heartbeat.html
 
 ## Example Rules
 
--- a/collector/heartbeat.go
+++ b/collector/heartbeat.go
@@ -1,0 +1,89 @@
+// Scrape heartbeat data.
+
+package collector
+
+import (
+	"database/sql"
+	"fmt"
+	"strconv"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	// heartbeat is the Metric subsystem we use.
+	heartbeat = "heartbeat"
+	// heartbeatQuery is the query used to fetch the stored and current
+	// timestamps. %s will be replaced by the database and table name.
+	// The second column allows gets the server timestamp at the exact same
+	// time the query is run.
+	heartbeatQuery = "SELECT UNIX_TIMESTAMP(ts), UNIX_TIMESTAMP(NOW(6)), server_id from `%s`.`%s`"
+)
+
+// Metric descriptors.
+var (
+	HeartbeatStoredDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, heartbeat, "stored_timestamp_seconds"),
+		"Timestamp stored in the heartbeat table.",
+		[]string{"server_id"}, nil,
+	)
+	HeartbeatNowDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, heartbeat, "now_timestamp_seconds"),
+		"Timestamp of the current server.",
+		[]string{"server_id"}, nil,
+	)
+)
+
+// ScrapeHeartbeat scrapes from the heartbeat table.
+// This is mainly targeting pt-heartbeat, but will work with any heartbeat
+// implementation that writes to a table with two columns:
+// CREATE TABLE heartbeat (
+//  ts                    varchar(26) NOT NULL,
+//  server_id             int unsigned NOT NULL PRIMARY KEY,
+// );
+func ScrapeHeartbeat(db *sql.DB, ch chan<- prometheus.Metric, collectDatabase, collectTable *string) error {
+	query := fmt.Sprintf(heartbeatQuery, *collectDatabase, *collectTable)
+	heartbeatRows, err := db.Query(query)
+	if err != nil {
+		return err
+	}
+	defer heartbeatRows.Close()
+
+	var (
+		now, ts  sql.RawBytes
+		serverId int
+	)
+
+	for heartbeatRows.Next() {
+		if err := heartbeatRows.Scan(&ts, &now, &serverId); err != nil {
+			return err
+		}
+
+		tsFloatVal, err := strconv.ParseFloat(string(ts), 64)
+		if err != nil {
+			return err
+		}
+
+		nowFloatVal, err := strconv.ParseFloat(string(now), 64)
+		if err != nil {
+			return err
+		}
+
+		serverId := strconv.Itoa(serverId)
+
+		ch <- prometheus.MustNewConstMetric(
+			HeartbeatNowDesc,
+			prometheus.GaugeValue,
+			nowFloatVal,
+			serverId,
+		)
+		ch <- prometheus.MustNewConstMetric(
+			HeartbeatStoredDesc,
+			prometheus.GaugeValue,
+			tsFloatVal,
+			serverId,
+		)
+	}
+
+	return nil
+}
--- a/collector/heartbeat_test.go
+++ b/collector/heartbeat_test.go
@@ -1,0 +1,49 @@
+package collector
+
+import (
+	"testing"
+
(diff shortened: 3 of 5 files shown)


brian-brazil:
How many different heartbeat implementations are there out there?

roidelapluie:
pt-heartbeat has been there for a very long time; it is widely used.

grypyrg:
@brian-brazil : `pt-heartbeat` is the most widely used (if not only) alternative tool to measure replication lag properly.
(Of course, there are many companies that do heartbeats using a custom built solution)

Given these 2 new metrics, can we easily get the time difference with them?
Maybe we should also calculate in `mysqld_exporter` itself already how much the difference is and write it to a different metric to make it easier to report the replication lag.

SuperQ:
Could be interesting.

In what way is `SHOW SLAVE STATUS` unreliable?

Monitoring for slave lag with Prometheus

SuperQ:
Argh, wrong button. And no way to edit on mobile.

grypyrg:
`SHOW SLAVE STATUS` is unreliable when:
- parallel replication is used (`slave_parallel_workers>0`). several metrics are only updated in regular intervals (https://dev.mysql.com/doc/refman/5.7/en/replication-options-slave.html#option_mysqld_slave-checkpoint-group & https://dev.mysql.com/doc/refman/5.7/en/replication-options-slave.html#option_mysqld_slave-checkpoint-period
- when IO thread is not catching up with the master (due to limited bandwith), and the sql thread has caught up with the IO thread, `seconds_behind_master` will reported as `0`
- in multi-tiered architectures, replication lag is always the timestamp from where the query was originally executed. with `pt-heartbeat` you can run the heartbeat on the intermediate server and get replication lag for that tier only
- when timezones are different between servers you can have for example `+3600` or `-3600` report seconds replication lag.
- https://bugs.mysql.com/bug.php?id=80084
- ...

roidelapluie:
Fixed:
- added _seconds suffix as recommended in Prometheus documentation.
- added the actual lag time
- rounded to integers. Makes no sense to go below the second as pt-heartbeat only updates the table once per second.

SuperQ:
@grypyrg Awesome, thanks for documenting those issues here.

SuperQ:
Even with second-only precision hearbeats, it's still worth keeping the float data as is valid.  Especially if `pt-heartbeat` attempts to write on-the-second and not just `sleep(1)`

Maybe for compatibility, we should implement a `pt-hearbeat` compatible writer and include it in the `mysqld_exporter` repo.  We could also implement it to write out millisecond precision timestamps.

It's really too bad that `pt-heartbeat` uses a varchar for the timestamp instead of a `datetime(6)`.

collector/pt_heartbeat.go
```diff
@@ -0,0 +1,95 @@
...
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	// Metric subsystem
+	ptHeartbeat = "pt_heartbeat"
+	// pt-heartbeat query
+	// the second column allows gets the server timestamp at the exact same time the query is run
+	ptHeartbeatQuery = "SELECT UNIX_TIMESTAMP(ts), UNIX_TIMESTAMP(CONCAT(DATE(NOW()), ' ', CURTIME(6))), server_id from %s"
```
SuperQ: This can be simplified with `UNIX_TIMESTAMP(NOW(6))`.  I tested on 5.5 and 5.6.

roidelapluie:
@SuperQ pt-heartbeat'i very initial version dates from https://sourceforge.net/p/maatkit/code/857/ 2007-09-05. The latest version is the result of almost 10 years of maturing. Older releases were using timestamps, and that got changed to VARCHAR later. I could not track back the reason.

I do agree that a Go version of that tool might be handy, but is out of scope of this pull request.

collector/pt_heartbeat.go
```diff
@@ -0,0 +1,95 @@
...
+		prometheus.BuildFQName(namespace, ptHeartbeat, "stored_timestamp_seconds"),
+		"Timestamp stored in the pt-heartbeat table.",
+		[]string{"server_id"}, nil,
+	)
+	PtHeartbeatNowDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, ptHeartbeat, "now_timestamp_seconds"),
+		"Timestamp of the current server.",
+		[]string{"server_id"}, nil,
+	)
+	PtHeartbeatDelayDesc = prometheus.NewDesc(
```
SuperQ: This metric isn't necessary (and against Prometheus design) since you can do this in PromQL.

README.md
```diff
@@ -66,6 +66,8 @@ collect.perf_schema.indexiowaits                       | 5.6           | Collect
 collect.perf_schema.tableiowaits                       | 5.6           | Collect metrics from performance_schema.table_io_waits_summary_by_table.
 collect.perf_schema.tablelocks                         | 5.6           | Collect metrics from performance_schema.table_lock_waits_summary_by_table.
 collect.slave_status                                   | 5.1           | Collect from SHOW SLAVE STATUS (Enabled by default)
+collect.pt-heartbeat                                   | 5.1           | Collect from pt-heartbeat
```
SuperQ: I would prefer to call this `heartbeat`, as there could be other implementations.

SuperQ:
Overall, I like this idea, some comments in-line.

roidelapluie:
Should I rename ptHeartbeat to heartbeat everywhere or just in the CLI flags?

collector/pt_heartbeat.go
roidelapluie: @grypyrg That delay can be made at the prometheus level using https://prometheus.io/docs/querying/rules/ . Once this patch gets in, maybe we can document an example recording rile for this on a blogpost.

roidelapluie:
Updated after latest comments.

@SuperQ do you want more documintatior for this in README?

SuperQ:
@roidelapluie Totally agree, re-implementing is out of scope for this feature.  I was mostly just thinking out loud.

Yes, please remove `pt` from everything.  But add a note and a link to `pt-heartbeat` in the README that "this is a supported implementation of heartbeat checking".

collector/heartbeat.go
```diff
@@ -0,0 +1,88 @@
...
+)
+
+const (
+	// heartbeat is the Metric subsystem we use.
+	heartbeat = "heartbeat"
+	// heartbeatQuery is the query used to fetch the stored and current
+	// timestamps. %s will be replaced by the table name.
+	// The second column allows gets the server timestamp at the exact same
+	// time the query is run.
+	heartbeatQuery = "SELECT UNIX_TIMESTAMP(ts), UNIX_TIMESTAMP(NOW(6)), server_id from `%s`"
```
roidelapluie: @SuperQ how can we prevent this to turn into sql injection?

grypyrg:
@roidelapluie : `pt-heartbeat` does support subsecond checking. See https://www.percona.com/doc/percona-toolkit/2.1/pt-heartbeat.html#cmdoption-pt-heartbeat--interval 
```
The minimum (fastest) interval is 0.01, and the maximum precision is two decimal places, 
so 0.015 will be rounded to 0.02.
```

collector/heartbeat.go
roidelapluie: Apparently not much https://github.com/golang/go/issues/18478

grypyrg:
And maybe fetch another column to precalculate the `seconds_lag`

```
SELECT UNIX_TIMESTAMP(ts) AS executed_time,
       UNIX_TIMESTAMP(NOW(6)) AS server_time,
       UNIX_TIMESTAMP(NOW(6)) - UNIX_TIMESTAMP(ts) AS seconds_lag,
       server_id
FROM `%s`
```

Example output (to also show it's subsecond):
```
+-------------------+-------------------+-------------+-----------+
| executed_time     | server_time       | seconds_lag | server_id |
+-------------------+-------------------+-------------+-----------+
| 1487668183.001080 | 1487668278.442121 |   95.441041 |         1 |
+-------------------+-------------------+-------------+-----------+
```

roidelapluie:
@grypyrg According to @grypyrg the lag should be computed at the promsql level.

roidelapluie:
Updated with all the commands taken in consideration; also split database and table parameters so we call better escape them.

collector/pt_heartbeat.go
roidelapluie: fixed

collector/pt_heartbeat.go
roidelapluie: removed

README.md
roidelapluie: renamed everywhere

collector/heartbeat.go
roidelapluie: split in two: database and table, for better escaping

SuperQ:
Maybe make a parameter for the column?

roidelapluie:
Two parameters then. Is that complexity needed for an initial release?

SuperQ:
No, not necssary, it was just an idea.  I think it's fine to keep the columns hard-coded.

collector/heartbeat.go
```diff
@@ -0,0 +1,88 @@
...
+// );
+func ScrapeHeartbeat(db *sql.DB, ch chan<- prometheus.Metric, collectDatabase, collectTable *string) error {
+	query := fmt.Sprintf(heartbeatQuery, *collectDatabase, *collectTable)
+	heartbeatRows, err := db.Query(query)
+	if err != nil {
+		return err
+	}
+	defer heartbeatRows.Close()
+
+	var now sql.RawBytes
```
SuperQ: This should be:
```
var (
  now, ts  sql.RawBytes
  serverId int
)
```

roidelapluie:
Go nit fixed.

roidelapluie:
-2 now until I check my english wording

roidelapluie:
Wording fixed in README.md and mysqld_exporter.go, so for me it is good to merge.

SuperQ:
Yup, looks fine, thanks.

grypyrg:
Thanks @roidelapluie !

roidelapluie:
Fer reference, promsql query:

mysql_heartbeat_now_timestamp_seconds - on (server_id) mysql_heartbeat_stored_timestamp_seconds

SuperQ:
Normally Prometheus will be able to match up the labels and you don't need to specify the `on (server_id)`.

roidelapluie:
Oh thanks. Nice trick.

Issue #126, opened 2016-05-08
Implement Detailed file statistics capture


PeterZaitsev:
Hi,

Currently there is "event name" based file statistic implemented which allows to see the kind of IO but not towards what file it goes.  For example innodb_data_file will be used both for main tablespace and individual tables which have IO. 

file_summary_by_event_name

provides more detailed information showing actual file name.

**************************\* 4. row ***************************
                FILE_NAME: /var/lib/mysql/ibdata1
               EVENT_NAME: wait/io/file/innodb/innodb_data_file
    OBJECT_INSTANCE_BEGIN: 140501553632064
               COUNT_STAR: 2522
           SUM_TIMER_WAIT: 59152857149693
           MIN_TIMER_WAIT: 434922
           AVG_TIMER_WAIT: 23454740938
           MAX_TIMER_WAIT: 4821765466863
               COUNT_READ: 246
           SUM_TIMER_READ: 1295329342336
           MIN_TIMER_READ: 26980248
           AVG_TIMER_READ: 5265566296
           MAX_TIMER_READ: 397393276998
 SUM_NUMBER_OF_BYTES_READ: 6144000
              COUNT_WRITE: 1718
          SUM_TIMER_WRITE: 267658895693
          MIN_TIMER_WRITE: 2211063
          AVG_TIMER_WRITE: 155796770
          MAX_TIMER_WRITE: 115813693743
SUM_NUMBER_OF_BYTES_WRITE: 28147712
               COUNT_MISC: 558
           SUM_TIMER_MISC: 57589868911664
           MIN_TIMER_MISC: 434922
           AVG_TIMER_MISC: 103207650106
           MAX_TIMER_MISC: 4821765466863
**************************\* 5. row ***************************

SuperQ:
Oh, so you just want to have the `FILE_NAME` column as a label for this collector

https://github.com/prometheus/mysqld_exporter/blob/master/collector/perf_schema_file_events.go#L11-L17

PeterZaitsev:
Ben,

Somewhat.  I do not think that particular table has information about file
name:

mysql> select * from performance_schema.file_summary_by_event_name limit 1
\G
*************************** 1. row ***************************
               EVENT_NAME: wait/io/file/sql/map
               COUNT_STAR: 0
           SUM_TIMER_WAIT: 0
           MIN_TIMER_WAIT: 0
           AVG_TIMER_WAIT: 0
           MAX_TIMER_WAIT: 0
[...]

PR #189, opened 2017-03-20
Add basic file_summary_by_instance metrics

pskiwi:
This adds MySQL 5.5 compatible file access metrics. It might be related to #126.

I needed this to monitor table access in MySQL 5.5 where newer performance_schema tables are not present.

It could be enabled with:

    -collect.perf_schema.file_instances

The emitted metrics per file are (example):

    mysql_perf_schema_file_instances_bytes{file_name="table.MYD",mode="read"} 244
    mysql_perf_schema_file_instances_bytes{file_name="table.MYD",mode="write"} 0
    mysql_perf_schema_file_instances_total{file_name="table.MYD",mode="read"} 4
    mysql_perf_schema_file_instances_total{file_name="table.MYD",mode="write"} 0

Filtering 'file_name' is possible via a regexp to restrict the number of emitted metrics:

     -collect.perf_schema.file_instances.filter=basedir/.*(MYD|MYI)

The path-prefix of the 'file_name' label is automatically stripped, but could be re-enabled with

    -collect.perf_schema.file_instances.remove_prefix=false

It is mostly adapted from _perf_schema_file_events.go_.

### DIFF
--- a/collector/perf_schema_file_instances.go
+++ b/collector/perf_schema_file_instances.go
@@ -8,12 +8,12 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 
 	"flag"
-	"regexp"
+	"strings"
 )
 
 const perfFileInstancesQuery = `
 	SELECT
-	    FILE_NAME
+	    FILE_NAME,
 	    COUNT_READ, COUNT_WRITE,
 	    SUM_NUMBER_OF_BYTES_READ, SUM_NUMBER_OF_BYTES_WRITE
 	  FROM performance_schema.file_summary_by_instance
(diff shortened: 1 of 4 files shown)


collector/perf_schema_file_instances.go
```diff
@@ -0,0 +1,96 @@
...
+		if err := perfSchemaFileInstancesRows.Scan(
+			&fileName,
+			&countRead, &countWrite,
+			&sumBytesRead, &sumBytesWritten,
+		); err != nil {
+			return err
+		}
+
+		if *performanceSchemaFileInstancesRemovePrefix && *performanceSchemaFileInstancesFilter != ".*" {
+			pos := strings.LastIndexAny(fileName, "/\\")
```
SuperQ: Maybe use `path/filepath`[0] functions instead of custom string handling here.

[0]: https://golang.org/pkg/path/filepath/#Base

pskiwi: Yes, this would be more elegant. But it doesn't work if the exporter runs on Linux and MySQL runs on Windows. Then you would have to replace all separator chars of the MySQL Host OS with the exporter Host OS ones first. I don't know a more elegant solution than this:

    p := "c:\\windows\\myfile"
    var other string
	this := strconv.QuoteRune(filepath.Separator)
	if filepath.Separator == '/' {
		other = "\\"
	} else {
		other = "/"
	}
	p = strings.Replace(p, other, this, -1)
	fmt.Printf("Base[%s]\n", filepath.Base(p))

What do you think?

SuperQ: It is always recommended to run the mysqld_exporter on localhost, so it should be the same platform as the server.

pskiwi: (It's sometimes handy to work cross platform and to monitor instances you don't have full access.)
I changed it to

    fileName = path.Base(filepath.ToSlash(fileName))

because _path.Base_ only works with '/'. But _ToSlash_ doesn't do much on Linux.

Issue #198, opened 2017-05-19
How to count TPS?


reymont:
Base MySQL document:

TPS = (Com_commit + Com_rollback) / Seconds

I found the prometheus metric is 
TPS = (mysql_global_status_commands_total{command="commit"}+mysql_global_status_commands_total{command="rollback"})

but whats wrong with 

`(mysql_global_status_commands_total{command="commit"}+mysql_global_status_commands_total{command="rollback"})`

result "No datapoints found."? The two variable cannot add directly?

`http://XXX:9090/graph?g0.range_input=1h&g0.expr=mysql_global_status_commands_total%7Bcommand%3D%22commit%22%7D%2Bmysql_global_status_commands_total%7Bcommand%3D%22rollback%22%7D&g0.tab=0`

prometheus and mysqld_exporter version is:

mysqld_exporter_build_info{branch="master",goversion="go1.8.1",revision="80680068f15474f87847c8ee8f18a2939a26196a",version="0.10.0"}
prometheus_build_info{branch="master",goversion="go1.8.1",revision="4666df502c0e239ed4aa1d80abbbfb54f61b23c3",version="1.6.1"}

SuperQ:
This is done slightly differently with Prometheus, as you are trying to deal with two labels on the same metric.  The `+` operator is designed for handling addition of different metric names.  What you're hitting is a labeling miss-match.

Thankfully, this is a pretty easy operation in PromQL.

The query you're looking for is this:

```
sum(rate(mysql_global_status_commands_total{command=~"(commit|rollback)"}[5m])) without (command)
```

With PromQL, we can use label filtering to select the metrics we're interested in.  Then we apply the rate function to give us the "per second" of those metrics.   Then we sum up all of the metrics, but only without the command label. This will keep all of the other labels in the metric.

This might be a good one for our recording rules example.  Do you have a link to the MySQL documentation that talks about this?

reymont:
Thank you, the answer fixs my problems.

MySQL documentation is here:

> Transactions are atomic units of work that can be committed or rolled back

[tps](https://dev.mysql.com/doc/refman/5.7/en/glossary.html#glos_tps)
[transaction](https://dev.mysql.com/doc/refman/5.7/en/glossary.html#glos_transaction)

zhougui150:
thanks

zhougui150:
Also fix my problems

hmahdiany:
> ```
[...]

PR #199, opened 2017-05-19
Add a recording rule example for TPS

SuperQ:
Add a recording rule example for "Transactions Per Second".

[0]: https://dev.mysql.com/doc/refman/5.7/en/glossary.html#glos_transaction

Closes: https://github.com/prometheus/mysqld_exporter/issues/198


SuperQ:
CC: @reymont

Issue #203, opened 2017-06-01
metrics collect problem when use  mysql_exporter to collect performance schema metrics


justlooks:
the state of mysql endpoint is down and it says "server returned HTTP status 500 Internal Server Error" and i check the metrics, find the following error, anyone can tell me why?

> # curl http://172.16.20.7:9104/metrics
> 
> collected metric mysql_perf_schema_file_instances_total label:<name:"file_name" value:"db.opt" > label:<name:"mode" value:"read" > counter:<value:0 >  was collected before with the same name and label values

this is my start command

root       116     1  0 08:28 ?        00:00:01 /opt/mysqld_exporter-0.10.0.linux-amd64/mysqld_exporter -collect.binlog_size=true -collect.info_schema.processlist=true -collect.perf_schema.eventsstatements=true -collect.perf_schema.tableiowaits=true -collect.perf_schema.indexiowaits=true -collect.perf_schema.tablelocks=true -collect.perf_schema.eventswaits=true -collect.perf_schema.file_events=true -collect.perf_schema.file_instances=true

SuperQ:
I suspect there is a problem with that collector.  For now I suggest you disable the file_instances collector.

If you could attach a copy of the query results for the file instances [query](https://github.com/prometheus/mysqld_exporter/blob/master/collector/perf_schema_file_instances.go#L15-L22) that would help.

SuperQ:
Ok, I think I may see the issue with that collector.  It seems like the query/metric needs to include both the `FILE_NAME` and `EVENT_NAME` columns as labels in order to avoid duplication problems.

SuperQ:
I found two issues with collector, PR to fix them out for review.

justlooks:
that's means i need set both collect.perf_schema.file_instances.filter and collect.perf_schema.file_instances.remove_prefix ? but they both have default value

SuperQ:
The file_instances filter default value is fine.  The main problem is the default for `remove_prefix` isn't safe for all use cases.  Setting `-collect.perf_schema.file_instances.remove_prefix=false` will help.

But there is still the issue that we are missing a label from the `EVENT_NAME` column.

PR #205, opened 2017-06-01
Fix file_instances metric collector

SuperQ:
Include the event name column as a label for metrics gathered from `performance_schema.file_summary_by_instance`.

Default to not cleaning the path on `performance_schema.file_summary_by_instance`
names to avoid name colisions.

Closes: https://github.com/prometheus/mysqld_exporter/issues/203

### DIFF
--- a/collector/perf_schema_file_instances.go
+++ b/collector/perf_schema_file_instances.go
@@ -14,7 +14,7 @@ import (
 
 const perfFileInstancesQuery = `
 	SELECT
-	    FILE_NAME,
+	    FILE_NAME, EVENT_NAME,
 	    COUNT_READ, COUNT_WRITE,
 	    SUM_NUMBER_OF_BYTES_READ, SUM_NUMBER_OF_BYTES_WRITE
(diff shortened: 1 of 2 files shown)


SuperQ:
CC @roman-vynar

roman-vynar:
I guess it came from @pskiwi originally.

SuperQ:
Yup, I was hoping you could validate the sanity of the change.

roman-vynar:
@SuperQ looks good but something tells me that FILE_NAME+EVENT_NAME may not be enough or when removing the path prefix there could be still duplicates. Actually, it is hard to predict w/o running on real world systems :)

SuperQ:
Yup, that's what I figured.  This is why I'm disabling the path prefix feature by default now.

I'm thinking about removing the path prefix automatic removal and replacing this with a new flag.  It would be something like `remove_prefix="/var/lib/mysql/"` and default to empty string.  This way we can chop off redundancy in the path, while still making sure we're duplicate label safe.

PR #209, opened 2017-06-15
Move Exporter to /collector so it can be used as a library

arvenil:
This decouples collector implementation from http server and cli client.
This way I can create my own binary with several collectors by using `collector.New`.

All scrappers were already in `/collector` subpackage, so I just moved `Exporter` struct which is implementation of prometheus `Collector`, also to `/collector` subpackage.

### DIFF
--- a/collector/exporter.go
+++ b/collector/exporter.go
@@ -1,0 +1,364 @@
+package collector
+
+import (
+	"database/sql"
+	"time"
+
+	_ "github.com/go-sql-driver/mysql"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/common/log"
+)
+
+// Metric name parts.
+const (
+	// Subsystem(s).
+	exporter = "exporter"
+)
+
+// SQL Queries.
+const (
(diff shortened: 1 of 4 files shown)


SuperQ:
Great idea!

arvenil:
Builds now, I can rebase if you wish.

brian-brazil:
> This way I can create my own binary with several collectors by using collector.New.

This is not a recommended way to deploy things, see https://prometheus.io/docs/instrumenting/writing_exporters/#deployment

arvenil:
@brian-brazil I know that. It doesn't change anything for prometheus exporters.

It lets me to implement my own http server, with my own modifications... which I will likely try to push upstream if they make sense for you guys, but there might be some that doesn't.

This keeps cli and http interface in one package, and collector in another one - making easier to develop things on top of the collector.

SuperQ:
Looks like this needs a rebase.

arvenil:
Done.

SuperQ:
@grobie PTAL

frimik:
@arvenil I found an issue caused by this one, see #245

PR #210, opened 2017-06-16
Use native DB.Ping() instead of self-written implementation

arvenil:
I know it's trivial :P But I'm curious why not use native `DB.Ping()` instead of in-house implementation? It should also avoid generating unnecessary `SELECT 1` statements.

### DIFF
--- a/collector/exporter.go
+++ b/collector/exporter.go
@@ -25,8 +25,6 @@ const (
 	// See: https://github.com/go-sql-driver/mysql#system-variables
 	sessionSettingsParam = `log_slow_filter=%27tmp_table_on_disk,filesort_on_disk%27`
 	timeoutParam         = `lock_wait_timeout=%d`
-
-	upQuery = `SELECT 1`
 )
 
 // Tunable flags.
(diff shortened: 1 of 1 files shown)


SuperQ:
Totally reasonable to me, the `SELECT 1` pattern came from some app implementations with bad SQL drivers.  This is better.

SuperQ:
One note, we need to update the go mysql driver vendoring, as the `Ping()` function is not implemented in the current version.  I will send out a general vendoring update PR.

SuperQ:
It looks like this was only implemented recently[0] and is not in an official release of the driver just yet.  I think I want to hold off on this and wait till the upstream driver declares it in a stable release.

[0]: https://github.com/go-sql-driver/mysql/pull/608

arvenil:
Just to clarify

https://golang.org/pkg/database/sql/#DB.Ping is there since like forever™️
I just read `DB.Ping` is making a new connection to DB (which often is enough) but with implementation of `Pinger` interface it  will now also send ping packet to db (definitely better and more close to `SELECT 1` query test).

👍

SuperQ:
Looks like this needs a rebase.

SuperQ:
FYI, I'm not ignoring this, but there is still no upstream [release](https://github.com/go-sql-driver/mysql/releases) of the driver that contains the new `db.Ping()` version.

SuperQ:
We've merged 1.4.0, please rebase. :smile:

arvenil:
Done.

PR #217, opened 2017-07-04
Update Travis CI configuration

AlekSi:
1. Test with tip Go version. Go 1.9 will add a lot of stuff to `database/sql` so it makes sense to test it now.
2. Set Go import path on Travis CI. That's changes nothing for this repository but allows forks to use Travis CI without hacking configuration and makes merging with upstream easier.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,9 @@ sudo: false
 language: go
 go:
 - 1.8.x
+- tip
+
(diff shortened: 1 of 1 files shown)


SuperQ:
Hrm, I think we need to add 1.9 support to [promu](https://github.com/prometheus/promu) first.

AlekSi:
I don't think that so different from https://github.com/prometheus/node_exporter/pull/611.

PR #222, opened 2017-08-14
Switch to kingpin

carlpett:
Switch flag handling to kingpin as per [Prometheus #2455](https://github.com/prometheus/prometheus/issues/2455).


SuperQ:
FYI, this needs to be rebased.

carlpett:
Done

PR #228, opened 2017-08-28
Add Docker Compose for integration tests

AlekSi:
/cc @SuperQ

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ dist: trusty
 sudo: required
 language: go
 
-language: go
 go:
 - 1.8.x
 - tip
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,29 @@
-sudo: false
+dist: trusty
+sudo: required
+language: go
 
 language: go
 go:
 - 1.8.x
 - tip
 
+env:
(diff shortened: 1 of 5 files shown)


SuperQ:
Very cool!  Thanks!

.travis.yml
```diff
@@ -1,11 +1,29 @@
-sudo: false
+dist: trusty
```
SuperQ: Any reason we shouldn't use `xenial` or `stretch` here?

.travis.yml
```diff
@@ -1,11 +1,29 @@
...
+sudo: required
+language: go
```
SuperQ: This seems like a dupe of Line 5.

.travis.yml
AlekSi: Travis CI doesn't have them: https://docs.travis-ci.com/user/reference/overview/

.travis.yml
AlekSi: Right, will fix it.

AlekSi: Done.

.travis.yml
SuperQ: Wow, ok, more reason to start moving Prometheus over to our new buildkite account. :-)

PR #233, opened 2017-09-14
Don't require a username or password when using a socket

joshuaspence:
A username and password aren't required when connecting to MySQL via a Unix socket.

### DIFF
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bytes"
 	"flag"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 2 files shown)


roman-vynar:
It's not about using a socket or not, it's all about security.
Username and password may not be required either for socket or TCP connection.
When not using any credentials, anonymous login (empty username/password on MySQL) is assumed which is a bad idea from security point of view.

joshuaspence:
Right, that is true that it technically applies to username/password login as well. But when you are using the `auth_socket` plugin there is no need for a password because MySQL uses the username of the user connecting to the socket for authentication.

drzraf:
This still matters. `auth_socket` / `unix_socket` is now the default for Debian+MariaDB.
With this auth plugin the way to setup connection is:
1. `CREATE USER IF NOT EXISTS 'prometheus'@'localhost' IDENTIFIED VIA unix_socket;`
then the command to run exporter could looks like:
2. `sudo -u prometheus -E DATA_SOURCE_NAME="unix(/run/mysqld/mysqld.sock)/" /usr/bin/prometheus-mysqld-exporter`

SuperQ:
@drzraf Thanks for the hint. I think we should add that to the README.

I've been thinking about changing the way this exporter handles connection/auth config. Because we're moving towards a "ENV is unsafe" line of thinking, I am considering dropping support for the ENV completely and going to using the `my.cnf` style only.

Issue #234, opened 2017-09-28
FR: support different intervals for different metrics collectors


santec:
It would be very nice to be able to configure different intervals for different collectors. For example, running queries on information_schema once a minute an show global status every 2 seconds.

I hope this is possible.

SuperQ:
The exporter has no knowledge of collection intervals, this is driven by the Prometheus server's scrape interval.

Selecting collectors via params is a feature we would like to have, but this is not currently implemented.  Previously it was not possible with the Prometheus client library, but we can now implement it with a custom handler and registry.

But I personally don't have time to implement this at the moment.

siavashs:
I will look into this 🙂

delgod:
JFI: implemented in [percona/mysqld_exporter/mysqld_exporter.go#L899-L925](https://github.com/percona/mysqld_exporter/blob/master/mysqld_exporter.go#L899-L925)

SuperQ:
Ahh, the entertainment of implementing basic auth, without ssl.  It makes me cringe that people think that's ok.

SuperQ:
@delgod There are a whole bunch of problems with that implementation.
* It creates separate subsystems for each metric.  This is generally a useless/bad idea as it will create duplicate metrics of the same data.
* It creates a limited set of 3 different static exporter modes rather than a more dynamic configuration with a URL parameter like `/metrics?collectors=global_status,global_values`.

PR #235, opened 2017-10-07
Add 'collect' URL parameter to filter enabled collectors

siavashs:
As discussed on #234 and IRC, this PR adds a `collect` URL parameter to filter currently enabled collectors.

Closes: https://github.com/prometheus/mysqld_exporter/issues/234

### DIFF
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -5,7 +5,6 @@ import (
 	"net/http"
 	"os"
 	"path"
-	"strings"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
@@ -191,12 +190,11 @@ func filter(filters map[string]bool, name string) bool {
 
 func handler(w http.ResponseWriter, r *http.Request) {
 	var filters map[string]bool
-	query := r.URL.Query().Get("collect")
-	log.Debugln("collect query:", query)
+	params := r.URL.Query()["collect[]"]
+	log.Debugln("collect query:", params)
 
-	if len(query) > 0 {
+	if len(params) > 0 {
 		filters = make(map[string]bool)
-		params := strings.Split(query, ",")
 		for _, param := range params {
 			f := flag.Lookup("collect." + param)
 			if f == nil {
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -5,8 +5,10 @@ import (
 	"net/http"
 	"os"
 	"path"
+	"strings"
 
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/prometheus/common/log"
 	"github.com/prometheus/common/version"
 	"gopkg.in/alecthomas/kingpin.v2"
@@ -132,6 +134,7 @@ var (
 		"collect.heartbeat.table",
 		"Table from where to collect heartbeat data",
 	).Default("heartbeat").String()
+  dsn string
 )
 
 // landingPage contains the HTML served at '/'.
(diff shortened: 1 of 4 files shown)


mysqld_exporter.go
```diff
@@ -170,6 +173,76 @@ func init() {
 	prometheus.MustRegister(version.NewCollector("mysqld_exporter"))
 }
 
+func filter(filters map[string]bool, name string) bool {
+	flg := flag.Lookup("collect." + name)
```
SuperQ: Minor nit, I think the golang style for this would just be `f`.

mysqld_exporter.go
```diff
@@ -170,6 +173,76 @@ func init() {
...
+		TableStat:            filter(filters, "info_schema.tablestats"),
+		QueryResponseTime:    filter(filters, "info_schema.query_response_time"),
+		EngineTokudbStatus:   filter(filters, "engine_tokudb_status"),
+		EngineInnodbStatus:   filter(filters, "engine_innodb_status"),
+		Heartbeat:            filter(filters, "heartbeat"),
+		HeartbeatDatabase:    *collectHeartbeatDatabase,
+		HeartbeatTable:       *collectHeartbeatTable,
+	}
+
+	//start := time.Now()
```
SuperQ: Please remove commented out extras.

mysqld_exporter.go
siavashs: I had used `f` in the original implementation, but changed it to `flg` later 😄 
I'll change it back.

SuperQ:
Very nice!  A few minor nits.

mysqld_exporter.go
siavashs: Ah, some leftovers from my lazy copy/pasting 😅

SuperQ:
Please update the README.md with usage info.

mysqld_exporter.go
```diff
@@ -170,6 +173,76 @@ func init() {
...
+}
+
+func handler(w http.ResponseWriter, r *http.Request) {
+	var filters map[string]bool
+	query := r.URL.Query().Get("collect")
+	log.Debugln("collect query:", query)
+
+	if len(query) > 0 {
+		filters = make(map[string]bool)
+		params := strings.Split(query, ",")
```
brian-brazil: The usage should be `?collect[]=foo&collect[]=bar` to keep in line with other places in Prometheus

siavashs: Good point 👌🏻

mysqld_exporter.go
```diff
@@ -170,6 +172,73 @@ func init() {
...
+		TableStat:            filter(filters, "info_schema.tablestats"),
+		QueryResponseTime:    filter(filters, "info_schema.query_response_time"),
+		EngineTokudbStatus:   filter(filters, "engine_tokudb_status"),
+		EngineInnodbStatus:   filter(filters, "engine_innodb_status"),
+		Heartbeat:            filter(filters, "heartbeat"),
+		HeartbeatDatabase:    *collectHeartbeatDatabase,
+		HeartbeatTable:       *collectHeartbeatTable,
+	}
+
+	registry := prometheus.NewRegistry()
```
brian-brazil: This won't include the default process stuff, so you should add it back in

roman-vynar:
@SuperQ awesome 😎 Much better than having 3 exporters.
/cc @rnovikovP

mysqld_exporter.go
```diff
@@ -170,6 +172,74 @@ func init() {
 	prometheus.MustRegister(version.NewCollector("mysqld_exporter"))
 }
 
+func filter(filters map[string]bool, name string) bool {
+	f := flag.Lookup("collect." + name)
```
SuperQ: The only down-side to this approach is that we want to switch to [kingpin](https://github.com/alecthomas/kingpin) for flag parsing.  I haven't been able to find an equivalent feature in kingpin. 😦

siavashs: Yes, it would have be nice if kingpin had a `Lookup()` method 😕
I can change the `filter()` implementation or submit a PR for kingpin 😅

SuperQ: Oh, I totally forgot to merge https://github.com/prometheus/mysqld_exporter/pull/222, which does the kingpin change.  I would like to merge that, and have the filter changed to match.  I think it will take too much time to get it implemented in kingpin.

Sorry, I should have taken care of the flag issue first.

siavashs: No problem, I'll update my branch.

siavashs:
As discussed on IRC, currently the `http` metrics are missing.
I could not find an example of how to instrument a handler function which initialises a custom registry, `blackbox` and `snmp` exporters only calculate the scrape duration and expose it.
[I also noticed that some instrumentation functions are deprecated now and we should use `promehttp` instead.](https://github.com/prometheus/client_golang/blob/535f493f5a875ebde111871209251257ef834ad1/prometheus/http.go#L187-L188)
So I need some help, maybe @beorn7 can give me a hint 😄

beorn7:
> So I need some help, maybe @beorn7 can give me a hint 😄 

The promhttp documentation is here: https://godoc.org/github.com/prometheus/client_golang/prometheus/promhttp

Prior to removing the deprecated functions, there will be more documentation, examples, and tooling. https://godoc.org/github.com/prometheus/client_golang/prometheus#InstrumentHandler is handy but doing so much wrong that we cannot just not declare it deprecated.

siavashs:
143b2d0 exposes the `http` metrics but the values are 0/Nan.

mysqld_exporter.go
```diff
@@ -172,6 +174,65 @@ func init() {
...
+		HeartbeatTable:       *collectHeartbeatTable,
+	}
+
+	registry := prometheus.NewRegistry()
+	registry.MustRegister(prometheus.NewGoCollector())
+	registry.MustRegister(prometheus.NewProcessCollector(os.Getpid(), ""))
+	registry.MustRegister(collector.New(dsn, collect))
+	prometheus.DefaultRegisterer = registry
+
+	// Delegate http serving to Promethues client library, which will call collector.Collect.
```
beorn7: Promethues -> Prometheus

mysqld_exporter.go
```diff
@@ -172,6 +174,65 @@ func init() {
...
+		Heartbeat:            filter(filters, "heartbeat", *collectHeartbeat),
+		HeartbeatDatabase:    *collectHeartbeatDatabase,
+		HeartbeatTable:       *collectHeartbeatTable,
+	}
+
+	registry := prometheus.NewRegistry()
+	registry.MustRegister(prometheus.NewGoCollector())
+	registry.MustRegister(prometheus.NewProcessCollector(os.Getpid(), ""))
+	registry.MustRegister(collector.New(dsn, collect))
+	prometheus.DefaultRegisterer = registry
```
beorn7: This will create a data race if multiple requests hit /metrics concurrently.

I assume you are only doing this here to make `prometheus.InstrumentHandler` work.

If you use the individual handler instrumentation helpers in promhttp, you can avoid that. Also, you can instrument the handler appropriately. (I'm pretty sure that some of the things that get auto-instrumented now are not needed, like request size. And response latency might be better done in a histogram so that you can aggregate over many exporters.)

mysqld_exporter.go
siavashs: I copied this typo from another exporter 😄

mysqld_exporter.go
siavashs: Yes, I noticed the issue while trying to make the instrumentation metrics work.
The next commit fixes this issue by creating a [`Gatherers`](https://godoc.org/github.com/prometheus/client_golang/prometheus#Gatherers) instance from default gatherer and newly created registry per request.

mysqld_exporter.go
```diff
@@ -172,6 +174,67 @@ func init() {
...
+		EngineTokudbStatus:   filter(filters, "engine_tokudb_status", *collectEngineTokudbStatus),
+		EngineInnodbStatus:   filter(filters, "engine_innodb_status", *collectEngineInnodbStatus),
+		Heartbeat:            filter(filters, "heartbeat", *collectHeartbeat),
+		HeartbeatDatabase:    *collectHeartbeatDatabase,
+		HeartbeatTable:       *collectHeartbeatTable,
+	}
+
+	registry := prometheus.NewRegistry()
+	registry.MustRegister(collector.New(dsn, collect))
+	prometheus.DefaultRegisterer = registry
```
beorn7: I think you can now delete this line. (As long as this line is there, you have a race.)

siavashs: Ah, of course.

beorn7:
:+1: from my side, leaving final say to @SuperQ .

mysqld_exporter.go
```diff
@@ -172,6 +174,66 @@ func init() {
...
+		QueryResponseTime:    filter(filters, "info_schema.query_response_time", *collectQueryResponseTime),
+		EngineTokudbStatus:   filter(filters, "engine_tokudb_status", *collectEngineTokudbStatus),
+		EngineInnodbStatus:   filter(filters, "engine_innodb_status", *collectEngineInnodbStatus),
+		Heartbeat:            filter(filters, "heartbeat", *collectHeartbeat),
+		HeartbeatDatabase:    *collectHeartbeatDatabase,
+		HeartbeatTable:       *collectHeartbeatTable,
+	}
+
+	registry := prometheus.NewRegistry()
+	registry.MustRegister(collector.New(dsn, collect))
```
arvenil: @SuperQ @siavashs I think this broke `scrapes_total` and `scrape_errors_total` metrics. Previously `collector.New()` was created once - in `main` func. Now it's created on every scrape and so is `scrapes_total` and `scrape_errors_total` leading to constant values as every time `scrapes_total` is reset to 0, so we get only 1 scrape. It's not `total` anymore.
https://github.com/prometheus/mysqld_exporter/blob/master/collector/exporter.go#L107-L118

siavashs: I can confirm this bug, I think we can fix it by having a separate collector for exporter metrics.

SuperQ: Sounds correct to me.  Thanks for the report.  I'll make a separate issue for this.

AlekSi: For the reference: https://github.com/prometheus/mysqld_exporter/issues/269

arvenil: @SuperQ @siavashs Actually, there is one more confusing thing. Looks like now this metric is constant `2`, and even after fixing it, this will be increased every http call by `2` because each call ask for metrics and their description, and `Describe` again increases `scrapes_total` metric. Should `scrapes_total` be increased also when `Describe` is called? This way each http call would increase this metric only by `1`. I'm working right now on fix for this so let me please know ;)

mysqld_exporter.go
```diff
@@ -181,47 +243,15 @@ func main() {
...
-		QueryResponseTime:    *collectQueryResponseTime,
-		EngineTokudbStatus:   *collectEngineTokudbStatus,
-		EngineInnodbStatus:   *collectEngineInnodbStatus,
-		Heartbeat:            *collectHeartbeat,
-		HeartbeatDatabase:    *collectHeartbeatDatabase,
-		HeartbeatTable:       *collectHeartbeatTable,
-	}
-
-	c := collector.New(dsn, collect)
-	prometheus.MustRegister(c)
```
arvenil: After this change, registering collector every http request, we now call `Describe()` every http request, and as in result we call Collect() twice every http request.

PR #237, opened 2017-10-16
Set last_scrape_error when an error occurs

### LINKED ISSUE #216: mysql_exporter_last_scrape_error 0 when mysql_up is 0
`mysql_exporter_last_scrape_error` is reporting 0 (success) also when `mysql_up` is 0 (down). 

This is a bit confusing if we rely on `mysql_exporter_last_scrape_error` to check for failures. 

Steps to reproduce:
- start `mysqld_exporter`
- saturate the connection pool of the MySQL user used
- `mysqld_exporter` will report: 
```
Jul 03 06:21:41 bash[14316]: time="2017-07-03T06:21:41-07:00" level=error msg="Error pinging mysqld: Error 1226: User 'monitoring_a' has exceeded the 'max_user_connections' resource (current value: 3)" source="mysqld_exporter.go:269"
```
while returning
```
mysql_exporter_last_scrape_error 0
mysql_up 0
mysql_exporter_scrapes_total 6716
```

siavashs:
Closes #216

### DIFF
--- a/collector/exporter.go
+++ b/collector/exporter.go
@@ -147,6 +147,7 @@ func (e *Exporter) scrape(ch chan<- prometheus.Metric) {
 	db, err := sql.Open("mysql", e.dsn)
 	if err != nil {
 		log.Errorln("Error opening connection to database:", err)
+		e.error.Set(1)
 		return
 	}
 	defer db.Close()
@@ -161,6 +162,7 @@ func (e *Exporter) scrape(ch chan<- prometheus.Metric) {
 	if err != nil {
 		log.Errorln("Error pinging mysqld:", err)
 		e.mysqldUp.Set(0)
+		e.error.Set(1)
 		return
 	}
 	isUpRows.Close()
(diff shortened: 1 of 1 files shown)


arvenil:
Just my curiosity, but isn't this and similar metrics prone to race conditions? Exporter can receive in theory 2 requests at the same time, as I understand it will process both requests in parallel, as so two `Describe` methods will update this metric in `*Exporter` at the same time?

arvenil:
@siavashs @SuperQ Just my curiosity, but isn't this and similar metrics prone to race conditions? Exporter can receive in theory 2 requests at the same time, as I understand it will process both requests in parallel, as so two `Describe` methods will update this metric in `*Exporter` at the same time? Or maybe default prometheus handler somehow queues requests?

SuperQ:
Yes, I was just thinking about this yesterday and realized this isn't going to work.  We should be doing this via a Const metric.

siavashs:
I think this should work fine since we create a new collector per request.

PR #238, opened 2017-10-19
drop support go1.5

s7v7nislands:
github.com/prometheus/client_golang says:
This library requires Go1.7 or later.


SuperQ:
We want this Makefile to match the [upstream](https://github.com/prometheus/prometheus/blob/master/Makefile#L14), so it would need to be changed there first.

Issue #153, opened 2016-08-18
Can't read from information_schema.table_statistics


aazon:
Hello,

I created a dedicated user `prom` with all [required grants](https://github.com/prometheus/mysqld_exporter#required-grants).
However there is no data related to `collect.info_schema.tables` in Prometheus. I tried to go inside the code in this repo and found out that the following query is used:

```sql
SELECT
  TABLE_SCHEMA,
  TABLE_NAME,
  ROWS_READ,
  ROWS_CHANGED,
  ROWS_CHANGED_X_INDEXES
  FROM information_schema.table_statistics
```

When I run this query on behalf of root I do receive results

```sql
mysql> select user();
+----------------+
| user()         |
+----------------+
| root@localhost |
+----------------+
1 row in set (0.00 sec)

mysql> SELECT count(*) FROM information_schema.table_statistics;
+----------+
| count(*) |
+----------+
|       30 |
+----------+
1 row in set (0.00 sec)
```
However when I run this query on behalf of prom` I receive an empty set.

```sql
mysql> select user();
+----------------+
| user()         |
+----------------+
| prom@localhost |
+----------------+
1 row in set (0.00 sec)

mysql> SELECT count(*) FROM information_schema.table_statistics;
+----------+
| count(*) |
+----------+
|        0 |
+----------+
1 row in set (0.00 sec)
```

I think this is the problem why `prom` can not get desired information and push it further to Prometheus. Could you please help to figure out what additional grants should user `prom` get in order to be able to fetch results from `information_schema.table_statistics`?

Current privileges are the following:

```sql
mysql> show grants for 'prom'@'localhost';
+----------------------------------------------------------------------------------------------------------------------------+
| Grants for prom@localhost                                                                                                  |
+----------------------------------------------------------------------------------------------------------------------------+
| GRANT PROCESS, REPLICATION CLIENT ON *.* TO 'prom'@'localhost' IDENTIFIED BY PASSWORD <secret> WITH MAX_USER_CONNECTIONS 3 |
| GRANT SELECT ON `performance_schema`.* TO 'prom'@'%'                                                                       |
+----------------------------------------------------------------------------------------------------------------------------+
2 rows in set (0.00 sec)
```

```sql
mysql> select version();
+-----------------+
| version()       |
+-----------------+
| 5.6.29-76.2-log |
+-----------------+
1 row in set (0.00 sec)
```

roman-vynar:
[...]

PR #242, opened 2017-10-31
Modified required grants in README.md

guidoiaquinti:
After reviewing the collectors code I think we can reduce the scope of the MySQL grants to strictly mentions what it needs to run.

I'm not sure we really need 

```GRANT SELECT ON `information_schema`.* TO 'exporter'@'localhost';```

as for InnoDB tables it should be indirectly granted by the `PROCESS` privilege (at least on MySQL 5.7+).

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ reference heartbeat implementation supported.
 
 [pth]:https://www.percona.com/doc/percona-toolkit/2.2/pt-heartbeat.html
 
-Please remind to add the following grant:
+You will need to grant the exporter access to the heartbeat table:
 
 ```sql
 GRANT SELECT ON `<db_name>.<table_name>` TO 'exporter'@'localhost';
--- a/README.md
+++ b/README.md
@@ -15,7 +15,9 @@ NOTE: Not all collection methods are supported on MySQL < 5.6
 
 ```sql
 CREATE USER 'exporter'@'localhost' IDENTIFIED BY 'XXXXXXXX' WITH MAX_USER_CONNECTIONS 3;
-GRANT PROCESS, REPLICATION CLIENT, SELECT ON *.* TO 'exporter'@'localhost';
+GRANT PROCESS, REPLICATION CLIENT ON *.* TO 'exporter'@'localhost';
+GRANT SELECT ON `information_schema`.* TO 'exporter'@'localhost';
+GRANT SELECT ON `performance_schema`.* TO 'exporter'@'localhost';
 ```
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: README.md [hunk 1]
```diff
@@ -108,6 +110,11 @@ reference heartbeat implementation supported.
 
 [pth]:https://www.percona.com/doc/percona-toolkit/2.2/pt-heartbeat.html
 
+Please remind to add the following grant:
```
SuperQ: How about: "You will need to grant the exporter access to the heartbeat table:"

brian-brazil:
I have a vague recollection that SELECT was required on *.* for some of the performance schema to work. Is that no longer the case?

guidoiaquinti:
I might be missing something but I don't think that's the case:
- [perf_schema_events_statements.go](https://github.com/prometheus/mysqld_exporter/blob/master/collector/perf_schema_events_statements.go#L14-L54)
- [perf_schema_events_waits.go](https://github.com/prometheus/mysqld_exporter/blob/master/collector/perf_schema_events_waits.go#L12-L13)
- [perf_schema_file_events.go](https://github.com/prometheus/mysqld_exporter/blob/master/collector/perf_schema_file_events.go#L12-L17)
- [perf_schema_file_instances.go](https://github.com/prometheus/mysqld_exporter/blob/master/collector/perf_schema_file_instances.go#L16-L21)
- [perf_schema_index_io_waits.go](https://github.com/prometheus/mysqld_exporter/blob/master/collector/perf_schema_index_io_waits.go#L12-L16)
- [perf_schema_table_io_waits.go](https://github.com/prometheus/mysqld_exporter/blob/master/collector/perf_schema_table_io_waits.go#L12-L17)
- [perf_schema_table_lock_waits.go](https://github.com/prometheus/mysqld_exporter/blob/master/collector/perf_schema_table_lock_waits.go#L12-L36)

brian-brazil:
I don't mean that the query reference it directly, I meant that there was additional access control under the covers in MySQLd that required a broad SELECT.

guidoiaquinti:
👍 got it. Unfortunately without integration tests running for all the supported MySQL versions I'm not sure how we can verify it.

brian-brazil:
Ah, #153

SuperQ:
@brian-brazil Yea, that's the one.  We still support MySQL 5.1, and `*.*` is required for those versions.

SuperQ:
Since this won't work on older versions, I think we have to close this for now.

PR #244, opened 2017-11-09
Add the ability to specify a file to read the DSN from.

bweston92:
This helps when using Docker secrets.

```
docker secret create mysql_exporter_dsn "dsn goes here"
docker service  create --name mysql_exporter -e DATA_SOURCE_NAME_FILE=/run/secrets/mysql_exporter_dsn --secret mysql_exporter_dsn prom/mysql_exporter
```

### DIFF
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"os"
 	"path"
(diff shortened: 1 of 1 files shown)


brian-brazil:
If we're adding a config file, then environment variable support should really be removed as it's not considered secure by Prometheus: https://prometheus.io/docs/operating/security/#secrets

SuperQ:
There is already DSN from file via the supported via the `my.cnf` style config file.

bweston92:
@brian-brazil I agree but for intention of this PR is to have the ability to use Docker secrets.
@SuperQ that requires more work then using Docker secrets.

bweston92:
@SuperQ I think you've closed that PR a bit premature to discussion. Docker secrets are also more secure then my.cnf file.

SuperQ:
Using the my.cnf style is pretty easy.

```
docker secret create my.cnf my.cnf
```

Then you specify the config with `--config.my-cnf=/run/secrets/my.cnf`.

PR #250, opened 2017-12-13
Add example for flags

dometec:


README.md
```diff
@@ -35,6 +35,11 @@ Running using ~/.my.cnf:
 
     ./mysqld_exporter <flags>
 
+Format for flags (example):
+  
+    -collect.auto_increment.columns
```
SuperQ: We have switched to kingpin for flags, so these have a different format now.

Boolean flags now take the form `--foo` and `--no-foo`.

dometec:
Hi SuperQ, thanks, I've update the example

Issue #167, opened 2016-11-07
Metadata lock interrupts metrics collection


matthiasr:
While a different connection holds a metadata lock, the exporter hangs on this query:

```
SELECT
                    TABLE_SCHEMA,
                    TABLE_NAME,
                    TABLE_TYPE,
                    ifnull(ENGINE, 'NONE') as ENGINE,
                    ifnull(VERSION, '0') as VERSION,
                    ifnull(ROW_FORMAT, 'NONE') as ROW_FORMAT,
                    ifnull(TABLE_ROWS, '0') as TABLE_ROWS,
                    ifnull(DATA_LENGTH, '0') as DATA_LENGTH,
                    ifnull(INDEX_LENGTH, '0') as INDEX_LENGTH,
                    ifnull(DATA_FREE, '0') as DATA_FREE,
                    ifnull(CREATE_OPTIONS, 'NONE') as CREATE_OPTIONS
                  FROM information_schema.tables
                  WHERE TABLE_SCHEMA = 'notifications' |
```

waiting for same metadata lock. Since no metrics are scrapeable, there is no more insight to be gained from Prometheus – specifically, the `mysql_up 1` doesn't make it to the server.

One way to avoid this would be to time out at some point and return what we have (or reconnect and carry on).

SuperQ:
It's possible to send the go-sql-driver timeouts via the DSN string.

https://github.com/go-sql-driver/mysql#readtimeout
https://github.com/go-sql-driver/mysql#timeout

I don't know if there's a way to do per-query timeouts with the normal Golang driver.

SuperQ:
@matthiasr Has this been tested to be "fixed" by setting the timeout?  I don't have a good reproduction case.

grypyrg:
For metadata locks you have to set `lock_wait_timeout` : http://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_lock_wait_timeout

EdSchouten:
I also ran into this issue earlier today. Turns out we have one... interesting customer who has tens of thousands of tables in one database. Solved this by setting `-collect.info_schema.tables=false`.

SuperQ:
@EdSchouten Can you give #252 a try?

EdSchouten:
In our case it's not really the locking that's problematic; it's just that our MySQL server contains too many tables, more than we're willing to ingest in the first place. Setting a locking timeout won't really make a difference in our case. :-)

PR #252, opened 2017-12-19
Set a lock_wait_timeout on the MySQL connection

SuperQ:
Set a session level `lock_wait_timeout` to avoid stuck metadata locks.

Closes: https://github.com/prometheus/mysqld_exporter/issues/167

### DIFF
--- a/collector/exporter.go
+++ b/collector/exporter.go
@@ -2,11 +2,13 @@ package collector
 
 import (
 	"database/sql"
+	"fmt"
 	"time"
 
 	_ "github.com/go-sql-driver/mysql"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/log"
+	"gopkg.in/alecthomas/kingpin.v2"
 )
 
 // Metric name parts.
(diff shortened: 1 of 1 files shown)


arvenil:
I'm surprised if this really works because you can't guarantee that db.Query() will be executed on the same session. Considering that this exporter limits number of open sessions, there is a chance it actually works... but this is still not guaranteed.

db is a session pool, and even with `db.SetMaxOpenConns(1)` you don't have guarantee that next `db.Query()` will use the same session as previous `db.Query()` call. It could always got close in background and new session could get open.

The only way to ensure you are using the same session is to use transactions.

https://github.com/go-sql-driver/mysql/issues/208

SuperQ:
`db` objects are created per scrape, it's not global for the exporter.  But, as was pointed out in another PR, there are better ways to deal with this via the DSN string.  I will probably refactor this later for better control.

arvenil:
@SuperQ just to clarify my position. Yes, you are right `db` is per scrape, but during scrape we are running multiple queries, and each query is picking a connection from connection pool. Yes, the connection pool is limited to just one connection, but this only means it's one connection at a time, not the same connection always.
For example here
https://golang.org/src/database/sql/sql.go?s=34389:34668#L1267
you can see we try to get connection few times, and if you dig deeper you will see it tries to grab cached one or if it doesn't work then it will create new one.

I just noticed new func, looks like recently added, and doc for this func explicitly says "Queries run on the same Conn will be run in the same database session. "
https://golang.org/pkg/database/sql/#DB.Conn
https://github.com/golang/go/commit/d234f9a75413fdae7643e4be9471b4aeccf02478
Never used that, so can't say how well it works, but could be an option for you.
(well, I don't agree with this 1-connection pool so maybe I shouldn't suggest it ;))

PR #254, opened 2017-12-25
Cleanup and simplify scrape logic

nussjustin:
The existing logic consists mostly of code tracking errors and timings, which is the same for each scrape "target".

SImplify the code by tracking all targets in a slice and ranging over it and moving the error and timing logic out into its own method.

Also add a comment to "log_slow_filter" that explains that it must be first. This was not obvious before and thus changes (e.g. #253) could accidentally change the behaviour. This can be avoided completely be settings the log_slow_filter variable via the DSN (see https://github.com/go-sql-driver/mysql/#system-variables), but this can be done in a later PR.

### DIFF
--- a/collector/exporter.go
+++ b/collector/exporter.go
@@ -177,6 +177,8 @@ func (e *Exporter) scrape(ch chan<- prometheus.Metric) {
 		label   string
 		fn      scrapeFunc
 	}{
+		// "log_slow_filter" needs to be the first, since it changes the session.
+		// TODO: Set via DSN parameter instead? See https://github.com/go-sql-driver/mysql/#system-variables
 		{e.collect.SlowLogFilter, "log_slow_filter", func(db *sql.DB, ch chan<- prometheus.Metric) error {
 			sessionSettingsRows, err := db.Query(sessionSettingsQuery)
 			if err != nil {
--- a/collector/exporter.go
+++ b/collector/exporter.go
@@ -139,9 +139,10 @@ func (e *Exporter) Collect(ch chan<- prometheus.Metric) {
 	ch <- e.mysqldUp
 }
 
+type scrapeFunc func(*sql.DB, chan<- prometheus.Metric) error
+
 func (e *Exporter) scrape(ch chan<- prometheus.Metric) {
 	e.totalScrapes.Inc()
-	var err error
 
 	scrapeTime := time.Now()
 	db, err := sql.Open("mysql", e.dsn)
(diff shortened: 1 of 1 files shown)


collector/exporter.go
```diff
@@ -169,222 +170,72 @@ func (e *Exporter) scrape(ch chan<- prometheus.Metric) {
...
-	}
-	if e.collect.TableSchema {
-		scrapeTime = time.Now()
-		if err = ScrapeTableSchema(db, ch); err != nil {
-			log.Errorln("Error scraping for collect.info_schema.tables:", err)
-			e.scrapeErrors.WithLabelValues("collect.info_schema.tables").Inc()
-			e.error.Set(1)
-		}
-		ch <- prometheus.MustNewConstMetric(scrapeDurationDesc, prometheus.GaugeValue, time.Since(scrapeTime).Seconds(), "collect.info_schema.tables")
+	scrapeTargets := []struct {
```
SuperQ: Maybe declare this instead of making an unnamed struct?

SuperQ:
Nice, thanks for the cleanup.  It's going to be a week before I can really review this.  Technically on holiday this week, and heading to 34c3. :smile:

collector/exporter.go
nussjustin: Done

I declared the struct inside the method, because it is currently not really useful outside of scrape since the enabled field depends on an *Exporter and to keep this change "small".

nussjustin:
No problem. Still, PTAL ;-)

collector/exporter.go
```diff
@@ -169,222 +170,74 @@ func (e *Exporter) scrape(ch chan<- prometheus.Metric) {
...
-	}
-	if e.collect.TableSchema {
-		scrapeTime = time.Now()
-		if err = ScrapeTableSchema(db, ch); err != nil {
-			log.Errorln("Error scraping for collect.info_schema.tables:", err)
-			e.scrapeErrors.WithLabelValues("collect.info_schema.tables").Inc()
-			e.error.Set(1)
-		}
-		ch <- prometheus.MustNewConstMetric(scrapeDurationDesc, prometheus.GaugeValue, time.Since(scrapeTime).Seconds(), "collect.info_schema.tables")
+	type scrapeTarget struct {
```
SuperQ: I think this should be in the top level `collector` package scope.  @juliusv what do you think?

juliusv: Since this helper struct is only used here, defining it within the method is fine, but I would even go one step further and keep the type anonymous:

```go
scrapeTargets := []struct{
  enabled bool
  label string
  fn scrapeFunc
}{
  {...},
  {...},
  ...
}
```

SuperQ: Ok, I guess I was wrong to ask for it to be split out in the first place. :smiley:

nussjustin: Reverted / removed the change that named the struct :-)

arvenil:
I've actually started doing similar thing for the fork (before reading this issue), work still in progress. My first idea was really similar to this one, but right now I'm thinking about creating an interface e.g.
```
type Scraper interface {
	Scrape(db *sql.DB, ch chan<- prometheus.Metric) error
	Name() string
	Version() float64
}
```
and then instead of func
```
func ScrapeEngineInnodbStatus(db *sql.DB, ch chan<- prometheus.Metric) error
```
create struct that fullifies this interface e.g.:
```
type ScrapeEngineInnodbStatus struct {
...
}

func (s *ScrapeEngineInnodbStatus) Scrape(db *sql.DB, ch chan<- prometheus.Metric) error {
...
}
```

And then pass all the enabled scrapers to *Exporter (so instead of Collect struct that has entry for each scraper, we could have a slice). This would also allow extending in forks with new Scrapes.

What do you think guys?

SuperQ:
@juliusv What do you think of this approach vs @arvenil's interface idea?

Issue #226, opened 2017-08-22
Error 9002: SSL connection is required.


rnkhouse:
### Host operating system: output of `uname -a`
Linux vagrant-ubuntu-trusty-64 3.13.0-121-generic #170-Ubuntu SMP Wed Jun 14 09:04:33 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux

### MySQL server version
5.6
### What did you do that produced an error?
.my.cnf file:
[client]
user=dsad@dsa
password=dasd@dsaW#
host=eastus1-a.control.database.windows.net
[mysqld]
ssl-cert=/etc/mysql/BaltimoreCyberTrustRoot.crt.pem
ssl-ca=/etc/mysql/BaltimoreCyberTrustRoot.crt.pem

Command to run mysqld_exporter: `sudo ./mysqld_exporter -config.my-cnf=".my.cnf"`

### What did you expect to see?
Successfully connected to Azure database

### What did you see instead?
INFO[0325] Error scraping slave state: Error 9002: SSL connection is required. Please specify SSL options and retry.   file=mysqld_exporter.go line=824
INFO[0325] Error scraping table schema: Error 9002: SSL connection is required. Please specify SSL options and retry.   file=mysqld_exporter.go line=836
INFO[0327] Error scraping global state: Error 9002: SSL connection is required. Please specify SSL options and retry.   file=mysqld_exporter.go line=812
INFO[0328] Error scraping global variables: Error 9002: SSL connection is required. Please specify SSL options and retry.   file=mysqld_exporter.go line=818

SuperQ:
The exporter currently doesn't support SSL connection parameters via the`.my.cnf` compatibility parser.  In order to use SSL, you need to use the [DATA_SOURCE_NAME](https://github.com/go-sql-driver/mysql#dsn-data-source-name) ENV variable to enable the [TLS parameter](https://github.com/go-sql-driver/mysql#tls).

For more advanced SSL use, like custom certificate verification, some additional configuration support is needed.  Happy to review pull requests to add this.

rnkhouse:
Instead of .my.cnf I exported the `DATA_SOURCE_NAME` variable with this value:
`user:password@tcp(xxx..rds.amazonaws.com:3306)?tls=skip-verify`. But, still its looking for config in .my.cnf file. 

FATA[0000] no user specified under [client] in /home/vagrant/.my.cnf  file=mysqld_exporter.go line=1958

[...]

PR #255, opened 2018-01-07
Support custom CA truststore and client SSL keypair.

jaloren:
Per discussion in #226, there is no way to specify a custom CA that the
mysqld exporter will trust when establishing a SSL connection to a mysql
server. So if a mysql server has a custom truststore, an operator would
need to set DSN, where tls=skip-verify.

With this change, a user can define the ssl options in the mysql cnf and
then the mysqld exporter will construct a DSN and a custom TLS config
based on those options.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,24 @@ The MySQL server's [data source name](http://en.wikipedia.org/wiki/Data_source_n
 must be set via the `DATA_SOURCE_NAME` environment variable.
 The format of this variable is described at https://github.com/go-sql-driver/mysql#dsn-data-source-name.
 
+
+## Customizing Configuration for a SSL Connection
+if The MySQL server supports SSL, you may need to specify a CA truststore to verify the server's chain-of-trust. You may also need to specify a SSL keypair for the client side of the SSL connection. To configure the mysqld exporter to use a custom CA certificate, add the following to the mysql cnf file:
+
+```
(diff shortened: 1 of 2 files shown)


SuperQ:
I would prefer these to be flags, rather than in the config file.  By utilizing flags, it's still possible to control normal connection settings via the `DATA_SOURCE_NAME` env var.

brian-brazil:
I believe these are standard my.cnf settings.

SuperQ:
I would like to make it possible to keep the ENV + flags option, rather than force users to use a config file.  (It makes it easier for docker users)

jaloren:
@SuperQ I debated whether to add that in but was on the fence. The thing that makes this a little tricky is that DSN will need to have the name of the custom TLS config struct and it was not entirely clear to me how that should be specified. For example, we could hardcode it to "custom" and the document that the DSN needs to set that accordingly.

Anyway, I'll update the PR to support both use cases in the next day or two.

SuperQ:
@jaloren Any chance you still want to work on the ENV option for TLS configs?

I guess we could merge this and add the ENV stuff later.

jaloren:
@SuperQ sorry I just saw this email pop up. I have been swamped at work and I am not sure when I'll have a chance to look at this again. Thanks for the consideration and merge!

Issue #256, opened 2018-01-10
mysqld exporter generates duplicate mysql_perf_schema_file_instance* metrics when there are multiple databases


jaloren:
Here is the sql that the collector code is running:

```
SELECT  FILE_NAME, EVENT_NAME,    COUNT_READ, COUNT_WRITE,     SUM_NUMBER_OF_BYTES_READ, SUM_NUMBER_OF_BYTES_WRITE   FROM performance_schema.file_summary_by_instance      where FILE_NAME REGEXP '.*' AND FILE_NAME ?;
```

If you multiple databases, each row will have a different absolute path but the same name basename: db.opt.

This code is removing everything but the basename, which is resulting in the same metric existing more than once in a single export.

```
github.com/prometheus/mysqld_exporter/collector/perf_schema_file_instances.go:72
		if *performanceSchemaFileInstancesRemovePrefix {
			fileName = path.Base(filepath.ToSlash(fileName))
		}
```

### Host operating system: output of `uname -a`
3.10.0-693.5.2.el7.x86_64 #1 SMP Fri Oct 20 20:32:50 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux

### mysqld_exporter version: output of `mysqld_exporter --version`
mysqld_exporter, version 0.10.0 (branch: master, revision: 80680068f15474f87847c8ee8f18a2939a26196a)
  build user:       root@3b0154cd9e8e
  build date:       20170425-11:24:12
  go version:       go1.8.1

### MySQL server version
5.6

### What did you do that produced an error?
I had multiple databases defined.

### What did you expect to see?
mysqld exporter actually exports mysqld metrics.

### What did you see instead?
exporter return http response code 500 and the following body

```
An error has occurred during metrics collection:
8 error(s) occurred:
* collected metric mysql_perf_schema_file_instances_total label:<name:"file_name" value:"db.opt" > label:<name:"mode" value:"read" > counter:<value:0 >  was collected before with the same name and label values
* collected metric mysql_perf_schema_file_instances_total label:<name:"file_name" value:"db.opt" > label:<name:"mode" value:"write" > counter:<value:0 >  was collected before with the same name and label values
* collected metric mysql_perf_schema_file_instances_bytes label:<name:"file_name" value:"db.opt" > label:<name:"mode" value:"read" > counter:<value:0 >  was collected before with the same name and label values
[...]

PR #257, opened 2018-01-10
Fix prefix removal in perf_schema_file_instances

SuperQ:
Splitting on path is unsafe if two tables have the same name, but are in
different databases.  Use a single string prefix to remove the noisy
path data from the labels.

Closes: https://github.com/prometheus/mysqld_exporter/issues/256


brian-brazil:
:+1:

PR #259, opened 2018-01-10
Move session params to DSN

SuperQ:
In order to avoid lost session params if the `db` object reconnects in
the background, move session params to the DSN string configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ Name                                       | Description
 config.my-cnf                              | Path to .my.cnf file to read MySQL credentials from. (default: `~/.my.cnf`)
 log.level                                  | Logging verbosity (default: info)
 exporter.lock_wait_timeout                 | Set a lock_wait_timeout on the connection to avoid long metadata locking. (default: 2 seconds)
-exporter.log_slow_filter                   | Add a log_slow_filter to avoid exessive MySQL slow logging.  NOTE: Not supported by Oracle MySQL.
+exporter.log_slow_filter                   | Add a log_slow_filter to avoid slow query logging of scrapes.  NOTE: Not supported by Oracle MySQL.
 web.listen-address                         | Address to listen on for web interface and telemetry.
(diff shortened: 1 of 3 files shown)


SuperQ:
/CC @arvenil I think this will fix your concerns in https://github.com/prometheus/mysqld_exporter/pull/252

arvenil:
@SuperQ You are the best, yes, thank you makes more sense and looks IMO more clean.

SuperQ:
@arvenil Ok, I had to move the `log_slow_filter` flag into the exporter code.  This makes more sense to me anyway. :smile:

README.md
```diff
@@ -76,7 +76,8 @@ Name                                       | Description
 -------------------------------------------|--------------------------------------------------------------------------------------------------
 config.my-cnf                              | Path to .my.cnf file to read MySQL credentials from. (default: `~/.my.cnf`)
 log.level                                  | Logging verbosity (default: info)
-log_slow_filter                            | Add a log_slow_filter to avoid exessive MySQL slow logging.  NOTE: Not supported by Oracle MySQL.
+exporter.lock_wait_timeout                 | Set a lock_wait_timeout on the connection to avoid long metadata locking. (default: 2 seconds)
+exporter.log_slow_filter                   | Add a log_slow_filter to avoid exessive MySQL slow logging.  NOTE: Not supported by Oracle MySQL.
```
arvenil: Just a bonus question, what's the purpose of this flag in the exporter? To overwrite global `log_slow_filter` or to avoid query logging of exporter own queries?

This was a bit unclear to me from above description but after I found https://github.com/prometheus/mysqld_exporter/pull/60 the sentence "to avoid slow query logging of scrapes" makes it more clear.

SuperQ: It's a session flag to avoid the exporter logging its own queries.  Yes, I think your description of the flag is better.

SuperQ:
@arvenil Let me know if there's any other issues.

PR #263, opened 2018-01-31
Fix lock_wait_timeout after #259

arvenil:
Passing `exporterLockTimeout` to `lock_wait_timeout=%d` results in large integer instead of expected default `2` or whatever user passes. This is because `exporterLockTimeout` is a pointer so it results in passing memory address rather of `exporterLockTimeout` than value of `exporterLockTimeout`. Getting value with `*` solves the issue.

```diff
-	dsnParams := []string{fmt.Sprintf(timeoutParam, exporterLockTimeout)}
+	dsnParams := []string{fmt.Sprintf(timeoutParam, *exporterLockTimeout)}
```

Error I was getting:
```
Error 1231: Variable 'lock_wait_timeout' can't be set to the value of '842351053576'
```

### DIFF
--- a/collector/exporter.go
+++ b/collector/exporter.go
@@ -32,11 +32,11 @@ const (
 var (
 	exporterLockTimeout = kingpin.Flag(
 		"exporter.lock_wait_timeout",
-		"Set the MySQL session lock_wait_timeout to avoid stuck metadata locks",
+		"Set a lock_wait_timeout on the connection to avoid long metadata locking.",
 	).Default("2").Int()
 	slowLogFilter = kingpin.Flag(
 		"exporter.log_slow_filter",
-		"Add a log_slow_filter to avoid exessive MySQL slow logging.  NOTE: Not supported by Oracle MySQL.",
+		"Add a log_slow_filter to avoid slow query logging of scrapes. NOTE: Not supported by Oracle MySQL.",
 	).Default("false").Bool()
 
 	scrapeDurationDesc = prometheus.NewDesc(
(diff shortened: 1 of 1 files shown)


collector/exporter.go
```diff
@@ -32,11 +32,11 @@ const (
 var (
 	exporterLockTimeout = kingpin.Flag(
 		"exporter.lock_wait_timeout",
-		"Set the MySQL session lock_wait_timeout to avoid stuck metadata locks",
+		"Set a lock_wait_timeout on the connection to avoid long metadata locking.",
 	).Default("2").Int()
 	slowLogFilter = kingpin.Flag(
 		"exporter.log_slow_filter",
-		"Add a log_slow_filter to avoid exessive MySQL slow logging.  NOTE: Not supported by Oracle MySQL.",
+		"Add a log_slow_filter to avoid slow query logging of scrapes. NOTE: Not supported by Oracle MySQL.",
```
arvenil: Just to make it in sync with what's in README.

collector/exporter.go
```diff
@@ -88,17 +88,18 @@ type Exporter struct {
 // New returns a new MySQL exporter for the provided DSN.
 func New(dsn string, collect Collect) *Exporter {
 	// Setup extra params for the DSN, default to having a lock timeout.
-	dsnParams := []string{fmt.Sprintf(timeoutParam, exporterLockTimeout)}
+	dsnParams := []string{fmt.Sprintf(timeoutParam, *exporterLockTimeout)}
```
arvenil: The actual fix, pass value with `*` instead of address.

collector/exporter.go
```diff
@@ -88,17 +88,18 @@ type Exporter struct {
...
 	}
 
 	if strings.Contains(dsn, "?") {
-		dsn = dsn + "&" + strings.Join(dsnParams, "&")
+		dsn = dsn + "&"
 	} else {
-		dsn = dsn + "?" + strings.Join(dsnParams, "&")
+		dsn = dsn + "?"
 	}
+	dsn += strings.Join(dsnParams, "&")
```
arvenil: Just to drop logic dupe.

PR #275, opened 2018-04-16
Add innodb compression statistic

adivinho:
https://dev.mysql.com/doc/refman/5.5/en/innodb-cmp-table.html
https://dev.mysql.com/doc/refman/5.5/en/innodb-cmpmem-table.html


README.md
```diff
@@ -58,6 +58,8 @@ collect.global_variables                               | 5.1           | Collect
 collect.info_schema.clientstats                        | 5.5           | If running with userstat=1, set to true to collect client statistics.
 collect.info_schema.innodb_metrics                     | 5.6           | Collect metrics from information_schema.innodb_metrics.
 collect.info_schema.innodb_tablespaces                 | 5.7           | Collect metrics from information_schema.innodb_sys_tablespaces.
+collect.info_schema.innodb_cmp                	       | 5.5           | Collect metrics from information_schema.innodb_cmp.
+collect.info_schema.innodb_cmpmem              	       | 5.5           | Collect metrics from information_schema.innodb_cmpmem.
```
SuperQ: The spacing seems off here.

SuperQ: Maybe let's give this a nicer description, like `Collect InnoDB buffer pool compression metrics from information_schema.innodb_cmpmem`.

collector/info_schema_innodb_cmp.go
```diff
@@ -0,0 +1,98 @@
...
+
+package collector
+
+import (
+	"database/sql"
+	"fmt"
+	"strings"
+
+	"github.com/prometheus/client_golang/prometheus"
+	//	"github.com/prometheus/common/log"
```
SuperQ: Remove commented out code.

collector/info_schema_innodb_cmp.go
```diff
@@ -0,0 +1,98 @@
...
+)
+
+const innodbCmpQuery = `SELECT * FROM information_schema.INNODB_CMP`
```
SuperQ: We prefer to use explicit column names in the exporter to avoid having to assume the column list.

SuperQ: Also, for consistency, we can use the lowercase table name here.

collector/info_schema_innodb_cmp.go
```diff
@@ -0,0 +1,98 @@
...
+
+var (
+	// Map known client-statistics values to types. Unknown types will be mapped as
+	// untyped.
+	informationSchemaInnodbCmpTypes = map[string]struct {
```
SuperQ: It's not necessary to use a struct like this.  Take a look at the [innodb_metrics](https://github.com/prometheus/mysqld_exporter/blob/master/collector/info_schema_innodb_metrics.go#L21-L43) for a good example.

This code isn't wrong, but I would like to keep a consistent simple style between all of the collectors.  Because we prefer explicit column names, the code is simplified.

collector/info_schema_innodb_cmp.go
adivinho: The query's rewritten

collector/info_schema_innodb_cmp.go
adivinho: Code was rewritten as in the pointed example.

collector/info_schema_innodb_cmp.go
```diff
@@ -0,0 +1,96 @@
...
+// ScrapeInnodbCmp collects from `information_schema.innodb_cmp`.
+func ScrapeInnodbCmp(db *sql.DB, ch chan<- prometheus.Metric) error {
+
+	informationSchemaInnodbCmpRows, err := db.Query(innodbCmpQuery)
+	if err != nil {
+		return err
+	}
+	defer informationSchemaInnodbCmpRows.Close()
+
+	// The client column is assumed to be column[0], while all other data is assumed to be coerceable to float64.
```
SuperQ: I think this documentation is out of date with the refactoring.  Part of the reason I prefer the [explicit mapping](https://github.com/prometheus/mysqld_exporter/blob/master/collector/info_schema_innodb_metrics.go#L65-L67) rather than automatic mapping of columns is that there's less need for documentation like the code makes it clear what columns are mapped to what values.

adivinho: Mapping parts were rewritten.

collector/exporter.go
```diff
@@ -68,6 +68,8 @@ type Collect struct {
 	UserStat                bool
 	ClientStat              bool
 	TableStat               bool
+       InnodbCmp               bool
```
SuperQ: I think this indenting is off, please run `make format`.

adivinho: Done.

SuperQ:
Thanks!

PR #276, opened 2018-04-20
Update build

SuperQ:
* Update to circleci 2.0
* Update Makefile to include new tests, style check, etc.

Signed-off-by: Ben Kochie <superq@gmail.com>


krasi-georgiev:
Aren't we going to use the common makefile ?

https://github.com/prometheus/prometheus/blob/master/Makefile.common

SuperQ:
Yea, I suppose I could do that.

krasi-georgiev:
it could be a good test run. open an issue if there is a blocker so we can adjust.

SuperQ:
@krasi-georgiev Ok, this should be ready for review.

PR #265, opened 2018-02-01
Introduce Scraper interface

arvenil:
# Introduces Scraper interface:
```go
type Scraper interface {
	Scrape(db *sql.DB, ch chan<- prometheus.Metric) error
	Name() string
	Help() string
}
```

# Changes:
- Decouple exporter from scrapes. Pass scrapes instead of configuration to exporter. This should also allow easier extending in forks with new scrapes.
  ```patch
  -func New(dsn string, collect Collect) *Exporter {
  +func New(dsn string, scrapers []Scraper) *Exporter {
  ```
- Run in parallel all scrapes (still using single db connection).
  ```go
  wg := &sync.WaitGroup{}
  defer wg.Wait()
  for _, scraper := range e.scrapers {
  	wg.Add(1)
  	go func(scraper Scraper) {
  		defer wg.Done()
  		label := "collect." + scraper.Name()
  		scrapeTime := time.Now()
  		if err := scraper.Scrape(db, ch); err != nil {
  			log.Errorln("Error scraping for "+label+":", err)
  			e.scrapeErrors.WithLabelValues(label).Inc()
  			e.error.Set(1)
  		}
  		ch <- prometheus.MustNewConstMetric(scrapeDurationDesc, prometheus.GaugeValue, time.Since(scrapeTime).Seconds(), label)
  	}(scraper)
   }
  ```
- Writing new scrape means just implementing `collector.Scraper` interface e.g. like this:
  ```go
  // ScrapeEngineTokudbStatus scrapes from `SHOW ENGINE TOKUDB STATUS`.
  type ScrapeEngineTokudbStatus struct{}
  
  // Name of the Scraper.
  func (ScrapeEngineTokudbStatus) Name() string {
  	return "engine_tokudb_status"
  }
  
  // Help returns additional information about Scraper.
  func (ScrapeEngineTokudbStatus) Help() string {
  	return "Collect from SHOW ENGINE TOKUDB STATUS"
  }
  
  // Scrape collects data.
  func (ScrapeEngineTokudbStatus) Scrape(db *sql.DB, ch chan<- prometheus.Metric) error {
  	...
  	return nil
  }
  ```
  and adding it in `main.go` to the list of available `scrapers` with information if it should be enabled by default:
  ```go
  // scrapers lists all possible collection methods and if they should be enabled by default.
  var scrapers = map[collector.Scraper]bool{
  	collector.ScrapeGlobalStatus{}:                true,
  	collector.ScrapeGlobalVariables{}:             true,
[...]

PR #280, opened 2018-05-03
Allow to keep connections between scrapes plus various config options.

arvenil:
New options:
* `exporter.global-conn-pool`: Use global connection pool instead of creating new pool for each http request. Disabled by default - keeps original behavior.
* `exporter.max-open-conns`: Maximum number of open connections to the database (https://golang.org/pkg/database/sql/#DB.SetMaxOpenConns). By default `1` - keeps original behavior.
* `exporter.max-idle-conns`: Maximum number of connections in the idle connection pool (https://golang.org/pkg/database/sql/#DB.SetMaxIdleConns). By default `1` - keeps original behavior.
* `exporter.conn-max-lifetime`: Maximum amount of time a connection may be reused (https://golang.org/pkg/database/sql/#DB.SetConnMaxLifetime). By default `60s` - keeps original behavior.
* `collect.all`: Collect all metrics. Unrelated change but really handy especially for debug purposes.
* This covers https://github.com/prometheus/mysqld_exporter/pull/253 except parallel processing of userstats.

I think this could settle the issue with global connection pool vs connection pool per scrape. Also it gives full control over the number of connections and with https://github.com/prometheus/mysqld_exporter/pull/265 it improves parallel processing of scrapers. We also had some requests for that so it's already in https://github.com/percona/mysqld_exporter

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ Example format for flags for version <= 0.10.0:
 Name                                                   | MySQL Version | Description
 -------------------------------------------------------|---------------|------------------------------------------------------------------------------------
 collect.auto_increment.columns                         | 5.1           | Collect auto_increment columns and max values from information_schema.
-collect.binlog_size                                    | 5.1           | Collect the current size of all registered binlog files
+collect.binlog_size                                    | 5.1           | Collect the current size of all registered binlog files.
 collect.engine_innodb_status                           | 5.1           | Collect from SHOW ENGINE INNODB STATUS.
 collect.engine_tokudb_status                           | 5.6           | Collect from SHOW ENGINE TOKUDB STATUS.
 collect.global_status                                  | 5.1           | Collect from SHOW GLOBAL STATUS (Enabled by default)
@@ -82,7 +82,7 @@ collect.slave_status                                   | 5.1           | Collect
 collect.heartbeat                                      | 5.1           | Collect from [heartbeat](#heartbeat).
 collect.heartbeat.database                             | 5.1           | Database from where to collect heartbeat data. (default: heartbeat)
 collect.heartbeat.table                                | 5.1           | Table from where to collect heartbeat data. (default: heartbeat)
-
+collect.all                                            | -             | Collect all metrics.
 
 ### General Flags
 Name                                       | Description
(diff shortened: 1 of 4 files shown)


arvenil:
@SuperQ What do you think about this?

SuperQ:
Sorry, I was at KubeCon and very busy.

I think the option to increase the collection pool is a good idea, but I very much prefer to not have a global connection pool.  Part of the idea is that the scrape specifically tests the creation and connection workflow.

Having a global persistent pool has too much potential for abuse by users who do not realize implications of turning it on.

SuperQ:
I understand you want to disable it by default, but I do not want to integrate the feature.  I'm happy to accept the rest of this feature, but I really, really, recommend not supporting global connection pools no matter how much users think they need it.  I can not compromise best practices here.

arvenil:
What if:
* We have global, shared, connection pool for all metrics except `up`.
* `up` metric gets it's own scrapper, it always opens and closes connection, and users are able to turn it off if they are concerned with too many connections made to db. Or they will be able to also scrape it at different interval so users will be able to scrape metrics with `1s` interval but check `up` metric every e.g. `60s` which should has lower performance impact.

SuperQ:
Having `up` be a separate collector, with a special implementation, seems a lot of overhead. I would prefer to keep this exporter reasonably simple to use and understand.

Opening and closing MySQL connections is absurdly cheap from my observation, I've used connection threadpools with Percona server and have never seen this be an issue.

README.md
```diff
@@ -82,7 +82,7 @@ collect.slave_status                                   | 5.1           | Collect
 collect.heartbeat                                      | 5.1           | Collect from [heartbeat](#heartbeat).
 collect.heartbeat.database                             | 5.1           | Database from where to collect heartbeat data. (default: heartbeat)
 collect.heartbeat.table                                | 5.1           | Table from where to collect heartbeat data. (default: heartbeat)
-
+collect.all                                            | -             | Collect all metrics.
```
SuperQ: I don't see the point of this flag.  We collect all enabled collectors by default.

arvenil: Maybe wording is bad. This enables all collectors. By default only 4 is enabled. To enable all other you need to pass them one by one. It's fine, but for testing it's easier to just turn one flag. https://github.com/prometheus/mysqld_exporter/blob/master/mysqld_exporter.go#L37-L41

SuperQ: I don't think this is a feature we want to support.

PR #283, opened 2018-05-17
Ignore boolean keys in my.cnf.

arvenil:
Below is valid MySQL config however mysqld_exporter fails to read it,
because of the boolean keys.
Config:
```
[client]
user = root
password = abc123

[mysql]
skip-auto-rehash
```

For reference: https://jira.percona.com/browse/PMM-2469


arvenil:
What blocks it from merging?

SuperQ:
Sorry, got distracted. :grin:

PR #287, opened 2018-06-01
Add endpoint for mysql health state

carlo-blohm:
I want to use the endpoint for consul to check if the mysql is up. 
The mysql_exporter is knowing this, why do not export it as an endpoint.

### DIFF
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -90,7 +90,7 @@ func init() {
 	prometheus.MustRegister(version.NewCollector("mysqld_exporter"))
 }
 
-func newHandler(scrapers []collector.Scraper) http.HandlerFunc {
+func newHandler(scrapers []collector.Scraper, health *collector.MysqlHealth) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		filteredScrapers := scrapers
 		params := r.URL.Query()["collect[]"]
@@ -112,7 +112,8 @@ func newHandler(scrapers []collector.Scraper) http.HandlerFunc {
 		}
 
 		registry := prometheus.NewRegistry()
-		registry.MustRegister(collector.New(dsn, filteredScrapers))
+		exporter := collector.New(dsn, filteredScrapers)
+		registry.MustRegister(exporter)
 
 		gatherers := prometheus.Gatherers{
 			prometheus.DefaultGatherer,
@@ -121,10 +122,13 @@ func newHandler(scrapers []collector.Scraper) http.HandlerFunc {
 		// Delegate http serving to Prometheus client library, which will call collector.Collect.
 		h := promhttp.HandlerFor(gatherers, promhttp.HandlerOpts{})
 		h.ServeHTTP(w, r)
+		health.Set(exporter.MysqlHealth())
 	}
 }
 
 func main() {
+	health := collector.MysqlHealth{}
+
 	// Generate ON/OFF flags for all scrapers.
 	scraperFlags := map[collector.Scraper]*bool{}
 	for scraper, enabledByDefault := range scrapers {
@@ -178,7 +182,18 @@ func main() {
 			enabledScrapers = append(enabledScrapers, scraper)
 		}
 	}
-	http.HandleFunc(*metricPath, prometheus.InstrumentHandlerFunc("metrics", newHandler(enabledScrapers)))
+	http.HandleFunc(*metricPath, prometheus.InstrumentHandlerFunc("metrics", newHandler(enabledScrapers, &health)))
+	// Add an endpoint for mysql health state.
+	http.HandleFunc("/-/mysqld_healthy", func(w http.ResponseWriter, r *http.Request) {
+		if err := health.Get(); err != nil {
+			w.WriteHeader(http.StatusServiceUnavailable)
+			w.Write([]byte("critical: mysql_up"))
+			fmt.Fprint(w, err)
+		} else {
+			w.Write([]byte("ok: mysql_up"))
+		}
+	})
+
 	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
 		w.Write(landingPage)
 	})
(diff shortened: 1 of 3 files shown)


mysqld_exporter.go
```diff
@@ -178,6 +178,16 @@ func main() {
 		}
 	}
 	http.HandleFunc(*metricPath, prometheus.InstrumentHandlerFunc("metrics", newHandler(enabledScrapers)))
+        // add an endpoint for mysql health state
+	http.HandleFunc("/mysqlhealth", func(w http.ResponseWriter, r *http.Request) {
```
brian-brazil: /-/healthy is our standard health endpoint: https://github.com/prometheus/prometheus/blob/master/web/web.go#L308

mysqld_exporter.go
```diff
@@ -178,6 +178,17 @@ func main() {
 		}
 	}
 	http.HandleFunc(*metricPath, prometheus.InstrumentHandlerFunc("metrics", newHandler(enabledScrapers)))
+	// add an endpoint for mysql health state
+	http.HandleFunc("/-/healthy", func(w http.ResponseWriter, r *http.Request) {
+
+		if collector.GetmysqlHealth() {
+			w.Write([]byte("ok: mysqldup"))
```
SuperQ: To match the metric we provide[0], maybe call it `mysql_up`

[0]: https://github.com/prometheus/mysqld_exporter/blob/master/collector/exporter.go#L100-L104

SuperQ:
Fantastic.  I had considered writing such a feature, but had not had the time.

SuperQ:
What do you think of `http.StatusServiceUnavailable` vs `http. StatusInternalServerError`? Either works, but maybe unavailable makes more sense.

carlo-blohm:
I think it is to big to fit in the mysql_exporter and for a consul check it is to much for me. We are alerting on the metrics, the basic check is only to see if the process is working or not.
About the scrape interval, for now it's ok for me, but I do not know how to trigger a scrape on the healthy endpoint before it's sends the data.

SuperQ:
I think we could fit it in, but yes it's a lot more code.

I think this might be an OK start, and we could iterate on it. My only real concern now is that this code is not thread safe.  Take a look at https://notes.shichao.io/gopl/ch9/

carlo-blohm:
@SuperQ hope you like my approach, I am not sure if it's the best.

carlo-blohm:
Who is willing to merge or approve this?

SuperQ:
@juliusv Can you take a look at this?

juliusv:
Sorry, I missed this. I'm not a mysqld_exporter maintainer, but some observations:

- The semantics of `/-/healthy` in the other components that have this handler (Prometheus, Alertmanager, Pushgateway) are such that they inform about the health of the process itself, instead of proxying third-party backend health information. I'm not sure we should deviate from that here, or maybe it should be a new kind of endpoint that is specific to the backend, like `/-/mysqld_healthy`. @brian-brazil What do you think, since this question can come up for other exporters as well?
- For simply getting and setting a value, a goroutine+channel combo is uncommon and a bit clunky, just use a `sync.Mutex`. Also, the current solution is still not concurrency-safe, as reads of `mh.state` are not protected against concurrent writes. It only ensures that writes are serialized.
- Avoid global state - the health boolean should just be a field in the `Exporter` (together with a mutex that protects it). The exporter shouldn't update any global state.
- It might be a nice touch to store the `err` value instead of a boolean. Then the healthy handler could include error string details in the HTTP body if `err != nil`.
- Comment style in Prometheus repos is sentence-style: start comments capitalized, end with period.

brian-brazil:
The /-/healthy endpoint should only relate to the health of the exporter itself, and it is generally unwise to talk to any remote systems when implementing such an endpoint. The database being down does not mean that the exporter needs to be restarted.

carlo-blohm:
Hey @juliusv , @brian-brazil I changed following
- endpoint is now /-/mysqld_healthy
- using `sync.Mutex`
- try to use a less global approach
- use the `error` to show the mysqld error
- fixed the comments to make them a sentence

Do you like it and who can or is willing to merge?

SuperQ (approved):
LGTM.

SuperQ:
Looks ok to me, @brian-brazil / @juliusv?

mysqld_exporter.go
```diff
@@ -178,7 +182,18 @@ func main() {
...
 		}
 	}
-	http.HandleFunc(*metricPath, prometheus.InstrumentHandlerFunc("metrics", newHandler(enabledScrapers)))
+	http.HandleFunc(*metricPath, prometheus.InstrumentHandlerFunc("metrics", newHandler(enabledScrapers, &health)))
+	// Add an endpoint for mysql health state.
+	http.HandleFunc("/-/mysqld_healthy", func(w http.ResponseWriter, r *http.Request) {
+		if err := health.Get(); err != nil {
+			w.WriteHeader(http.StatusServiceUnavailable)
+			w.Write([]byte("critical: mysql_up"))
+			fmt.Fprint(w, err)
```
juliusv: The above three lines can be combined to:

```go
http.Error(w, fmt.Sprintf("critical: mysql_up: %v", err), http.StatusServiceUnavailable)
```

(currently I think there would be no space or newline before the `err` message)

juliusv:
Hmm, thinking about this further (sorry): the way the health information is currently retrieved from the `Exporter` is getting a bit convoluted. Also, it requires a scrape to happen before `/-/mysqld_healthy` has any information on whether MySQLd is healthy or not (it will actually always look healthy initially due to the `error` zero value being `nil`). What if nobody is currently scraping the exporter? Even if a scrape happened a long time ago, if Prometheus now stops scraping, it will report good health forever. Should the health check maybe be run in a separate background goroutine that is completely independent from normal scrapes?

SuperQ:
@juliusv This was my first concern. In my previous implementation of a mysqld healthy, it did a completely separate database ping from the rest of the metrics. We implemented this at SoundCloud as a separate process, but I had plans to integrate the code here.

But I haven't had the time to do this re-implementation.

ArthurSens:
Hi folks, I'm taking the liberty of closing this PR, as it's been stale for a few years. The idea is still good, so a new PR would be very welcome, as long as we address the concerns already expressed here.

Issue #269, opened 2018-03-12
Add collector for exporter metrics


SuperQ:
In order to correctly collect exporter-wide metrics, we need to register a separate collector.

SuperQ:
Fixed with #292.

Issue #245, opened 2017-11-11
collect parameter commit broke metric counter(s)?


frimik:
I was basing another prometheus exporter on this and noticed that things weren't working as intended for me ... I went back in history a bit and found that 8b92fe8a393d3f3d48b2f3bad2195d68ed8a893f broke at least one global metric counter: 

```# HELP mysql_exporter_scrapes_total Total number of times MySQL was scraped for metrics.
# TYPE mysql_exporter_scrapes_total counter
mysql_exporter_scrapes_total 2
```

On HEAD it's forever stuck on 2 (it gets init and counts to 2 on every scrape).

Sorry, edit; it's #235 that broke this. Messed up my history a bit.

frimik:
Sorry, it's #235 that broke this. Messed up my history a bit.

SuperQ:
Fixed with #292.

PR #292, opened 2018-06-27
PMM-2612: Scrape MySQL database only once and fix scrape counter.

arvenil:
There are two issues with scrapping I would like to address:
* `mysql_exporter_scrapes_total` never increases
  ```
  $ curl --silent localhost:9104/metrics | grep scrapes_total | grep -v '#'
  mysql_exporter_scrapes_total 2
  $ curl --silent localhost:9104/metrics | grep scrapes_total | grep -v '#'
  mysql_exporter_scrapes_total 2
  $ curl --silent localhost:9104/metrics | grep scrapes_total | grep -v '#'
  mysql_exporter_scrapes_total 2
  ```
  https://github.com/prometheus/mysqld_exporter/pull/235#discussion_r173890280
  
* `mysqld_exporter` every request scrapes source database twice - this means double load for MySQL database itself. The scrape is done twice because `Describe()` also runs `Collect()`. Additionally, every request we register new collector so every request we call `Describe()` and `Collect()`. This is different than in other exporters where exporter is registered only once so `Describe()` is called only on first http request but not on subsequent. This was also true for this exporter up to this change https://github.com/prometheus/mysqld_exporter/pull/235/files#r198540277
  In our fork we fixed first issue already, but this resulted in:
  ```
  $ curl --silent localhost:9104/metrics | grep scrapes_total | grep -v '#'
  mysql_exporter_scrapes_total 2
  $ curl --silent localhost:9104/metrics | grep scrapes_total | grep -v '#'
  mysql_exporter_scrapes_total 4
  $ curl --silent localhost:9104/metrics | grep scrapes_total | grep -v '#'
  mysql_exporter_scrapes_total 6
  ```
  Which clearly shows problem that database is now scrapped twice for every request.

This PR fixes both issues so it results in:
  ```
  $ curl --silent localhost:9104/metrics | grep scrapes_total | grep -v '#'
  mysql_exporter_scrapes_total 1
  $ curl --silent localhost:9104/metrics | grep scrapes_total | grep -v '#'
  mysql_exporter_scrapes_total 2
  $ curl --silent localhost:9104/metrics | grep scrapes_total | grep -v '#'
  mysql_exporter_scrapes_total 3
  ```

And internally database is indeed scrapped only once per http request.

### DIFF
--- a/collector/exporter.go
+++ b/collector/exporter.go
@@ -85,7 +85,7 @@ func (e *Exporter) Describe(ch chan<- *prometheus.Desc) {
 	ch <- e.metrics.TotalScrapes.Desc()
 	ch <- e.metrics.Error.Desc()
 	e.metrics.ScrapeErrors.Describe(ch)
-	ch <- e.metrics.MySQLUP.Desc()
+	ch <- e.metrics.MySQLUp.Desc()
 }
 
 // Collect implements prometheus.Collector.
@@ -95,7 +95,7 @@ func (e *Exporter) Collect(ch chan<- prometheus.Metric) {
 	ch <- e.metrics.TotalScrapes
 	ch <- e.metrics.Error
 	e.metrics.ScrapeErrors.Collect(ch)
-	ch <- e.metrics.MySQLUP
+	ch <- e.metrics.MySQLUp
 }
 
 func (e *Exporter) scrape(ch chan<- prometheus.Metric) {
@@ -120,13 +120,13 @@ func (e *Exporter) scrape(ch chan<- prometheus.Metric) {
 	isUpRows, err := db.Query(upQuery)
 	if err != nil {
 		log.Errorln("Error pinging mysqld:", err)
-		e.metrics.MySQLUP.Set(0)
+		e.metrics.MySQLUp.Set(0)
 		e.metrics.Error.Set(1)
 		return
 	}
 	isUpRows.Close()
 
-	e.metrics.MySQLUP.Set(1)
+	e.metrics.MySQLUp.Set(1)
 
 	ch <- prometheus.MustNewConstMetric(scrapeDurationDesc, prometheus.GaugeValue, time.Since(scrapeTime).Seconds(), "connection")
 
@@ -153,7 +153,7 @@ type Metrics struct {
 	TotalScrapes prometheus.Counter
 	ScrapeErrors *prometheus.CounterVec
 	Error        prometheus.Gauge
-	MySQLUP      prometheus.Gauge
+	MySQLUp      prometheus.Gauge
 }
 
 // NewMetrics creates new Metrics instance.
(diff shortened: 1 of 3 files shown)


collector/exporter.go
```diff
@@ -79,79 +76,37 @@ func New(dsn string, scrapers []Scraper) *Exporter {
...
 	e.scrape(ch)
 
-	ch <- e.totalScrapes
-	ch <- e.error
-	e.scrapeErrors.Collect(ch)
-	ch <- e.mysqldUp
+	ch <- e.metrics.TotalScrapes
+	ch <- e.metrics.Error
+	e.metrics.ScrapeErrors.Collect(ch)
+	ch <- e.metrics.MySQLUP
```
juliusv: Since `UP` is not an initialism (https://github.com/golang/go/wiki/CodeReviewComments#initialisms):

`MySQLUP` -> `MySQLUp`

collector/exporter.go
```diff
@@ -79,79 +76,37 @@ func New(dsn string, scrapers []Scraper) *Exporter {
...
-		close(doneCh)
-	}()
-
-	e.Collect(metricCh)
-	close(metricCh)
-	<-doneCh
+	ch <- e.metrics.TotalScrapes.Desc()
+	ch <- e.metrics.Error.Desc()
+	e.metrics.ScrapeErrors.Describe(ch)
+	ch <- e.metrics.MySQLUP.Desc()
```
juliusv: So we basically don't even try to register most of the metrics descs for this collector anymore? That's probably fine as it was best-effort anyway and we only have this one collector, but was surprised it wasn't explicitly mentioned.

arvenil: I've actually made a comment about it but looks like I forgot to press "Send" ;)
Yes. I think it doesn't make sense but your expertise here guys is highly required. To be honest I don't see when registering those descriptions could be useful, maybe if you write your own app, then you register all metrics first, they get cached, so you can just focus on sending metrics. But for standalone exporter where we explicitly say we can't determine all metrics I don't see the reason to do that. Any idea what the lack of registration of descriptions could impact? I didn't looked into implementation details so again, your knowledge here could be useful ;)

Also registering those metrics made sense as long as collector was registered once, on binary start, so first http request was calling `Describe()` but next request were using cached descriptions and `Describe()` was not called. Now we register collector every scrape, so `Describe` is called every http request anyway, cached descriptions are not re-used.

Initially I thought my idea was ugly hack, but I checked later and node_exporter does the same: https://github.com/prometheus/node_exporter/blob/master/collector/collector.go#L106-L110 They also in similar way register collector every http request https://github.com/prometheus/node_exporter/blob/master/node_exporter.go#L47

juliusv: Yeah, it's fine in this case.

juliusv:
:+1:

arvenil:
@SuperQ This closes also https://github.com/prometheus/mysqld_exporter/issues/245 and https://github.com/prometheus/mysqld_exporter/issues/269 ?

PR #300, opened 2018-07-06
Update collect[] param documentation

SuperQ:
Update the `collect[]` param documentation to match the `node_exporter`
docs. This avoids new user copy-pasta issues.


brian-brazil:
:+1:

PR #301, opened 2018-07-14
Refactor generation of data source name

Mattias-:
I have the intention to contribute to this project so that the data source configuration
optionally can be provided in multiple environment variables rather than one, `DATA_SOURCE_NAME`.
To achieve that I figured that it would make sense to get some refactoring in first.

### DIFF
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -67,7 +67,7 @@ var scrapers = map[collector.Scraper]bool{
 	collector.ScrapeSlaveHosts{}:                      false,
 }
 
-func parseMycnf(config interface{}) (string, error) {
+func parseMycnf(config interface{}) (*mysql.Config, error) {
 	mysqlConfig := mysql.NewConfig()
 	opts := ini.LoadOptions{
 		// MySQL ini file can have boolean keys.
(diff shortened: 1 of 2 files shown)


SuperQ:
I don't see the point of making it more complicated with multiple ENV vars. Can you explain more about what you're trying to solve?

Right now the DATA_SOURCE_NAME is a literal string supported by the upstream Go library. This makes it simple to configure and well documented in how it works.

Mattias-:
@SuperQ Basically I'm looking for way to provide a password independently from other configuration.

For example see the postgres_exporter
https://github.com/wrouesnel/postgres_exporter#environment-variables

SuperQ:
@Mattias- So, I thought about this some more, and I thinking we may want to fully remove the `DATA_SOURCE_NAME` and go to an entirely composed connection string for the Go driver.

This would solve issues like https://github.com/prometheus/mysqld_exporter/issues/303, where characters, like '?', in the password cause unintended side effects.

We can also better sanitize the input of protocol, host, and port. Plus set sane defaults since most users don't need to interact with those. I'm considering switching the default to unix socket.

I'd like to discuss this in an issue before we go too far on the code.

Mattias-:
@SuperQ Absolutely, I'll close this and open an issue

PR #323, opened 2018-08-30
Abort on timeout.

arvenil:
# Abort data collection: 
* when connection gets closed
* on prometheus scrape timeout `X-Prometheus-Scrape-Timeout-Seconds`
  * give exporter time to finish and send data with `--timeout-offset`. Default `0.1` second.

Tested manually. Returns metrics even if scraping was canceled. Example errors:
```console
ERRO[0014] Error scraping for collect.global_status: context deadline exceeded  source="exporter.go:104"
ERRO[0014] Error scraping for collect.info_schema.innodb_metrics: context deadline exceeded  source="exporter.go:104"
```

Based on https://github.com/prometheus/blackbox_exporter/blob/master/main.go#L74-L83

Let me know what you think.

### DIFF
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
@@ -8,6 +9,8 @@ import (
 	"net/http"
 	"os"
 	"path"
+	"strconv"
+	"time"
 
 	"github.com/go-sql-driver/mysql"
 	"github.com/prometheus/client_golang/prometheus"
@@ -29,6 +32,10 @@ var (
 		"web.telemetry-path",
 		"Path under which to expose metrics.",
 	).Default("/metrics").String()
+	timeoutOffset = kingpin.Flag(
+		"timeout-offset",
+		"Offset to subtract from timeout in seconds.",
+	).Default("0.5").Float64()
 	configMycnf = kingpin.Flag(
 		"config.my-cnf",
 		"Path to .my.cnf file to read MySQL credentials from.",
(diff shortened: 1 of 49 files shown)


brian-brazil:
In the blackbox exporter we use a flag-based constant offset, we should
probably be consistent.

arvenil:
Indeed. Thanks for pointing this out, I will change it.

arvenil:
Fixed.

mysqld_exporter.go
```diff
@@ -29,6 +32,10 @@ var (
 		"web.telemetry-path",
 		"Path under which to expose metrics.",
 	).Default("/metrics").String()
+	timeoutOffset = kingpin.Flag(
+		"timeout-offset",
+		"Offset to subtract from timeout in seconds.",
+	).Default("0.5").Float64()
```
SuperQ: Should we make this `0.1` if people are going to be scraping with 1s intervals for some collectors?

arvenil: This is why originally I went with percentage of timeout. But I think 0.5 makes more sense. It means, the collector has `0.5s` to stop collecting and send data collected so far to prometheus server. So it's more like a constant that depends on tcp connection and how long it takes for exporter to stop collecting.

But I'm simply ok with any version that works.

arvenil: What do you think @SuperQ ? Should I change it to `0.1`? It's not a big issue for us for now, we are mostly interested in any form of aborting data collection after timeout - so let me know if you would like any changes here.

arvenil:
Tested manually. Returns metrics even if scraping was canceled. Example errors:
```console
ERRO[0014] Error scraping for collect.global_status: context deadline exceeded  source="exporter.go:104"
ERRO[0014] Error scraping for collect.info_schema.innodb_metrics: context deadline exceeded  source="exporter.go:104"
```

mysqld_exporter.go
SuperQ: How about we split the difference, make it `0.25`. I think 250ms should be enough cleanup and transmit time for most use cases.

arvenil: Makes perfect sense. Done.

arvenil:
@SuperQ Worth merging or should I close it?

SuperQ:
@arvenil Sorry, busy month, it just fell off my review TODO list. I'll take a look at it on Monday.

Issue #241, opened 2017-10-25
Tag queries


AlekSi:
Query tags (comments) will help users to track them from the MySQL logs back to the mysqld_exporter. For example:
```
SELECT /* mysqld_exporter:performance_schema.events_waits_summary_global_by_event_name */ 
	  EVENT_NAME, COUNT_STAR, SUM_TIMER_WAIT
	  FROM performance_schema.events_waits_summary_global_by_event_name

SELECT /* mysqld_exporter:performance_schema.file_summary_by_event_name */
	    EVENT_NAME,
	    COUNT_READ, SUM_TIMER_READ, SUM_NUMBER_OF_BYTES_READ,
	    COUNT_WRITE, SUM_TIMER_WRITE, SUM_NUMBER_OF_BYTES_WRITE,
	    COUNT_MISC, SUM_TIMER_MISC
	  FROM performance_schema.file_summary_by_event_name
```

We could work on it if you are interested.

SuperQ:
LGTM, good idea.  The style question is, before `SELECT`, after like you propose, or after the whole statement.  I'm not sure if there is any consistency in the industry for this.

AlekSi:
That's the only format I'm aware of. The rationale for it is that it has nice sorting properties (all `SELECT`s are grouped, the ones with comments are grouped, etc.) and also comment is visible in relatively narrow console windows. My opinion may be one-sided though.

SuperQ:
The down-side is having it in the front takes up a lot of room in the default 250 char limit of 
 `collect.perf_schema.eventsstatements.limit`.

Maybe it would be good enough to just log `/* mysqld_exporter */`.  There's no big duplication of queries coming from the exporter, so it should be obvious which collector is sending the query.

AlekSi:
Well, it is obvious for mysqld_exporter developers, but not really obvious for other users like database administrators. Having collector name in the comment makes it clear what flag should be used to disable it without asking (and waiting for) developers or reading source code. That was my original reason to propose that feature.

AlekSi:
Bump. I propose to settle with `Scraper.Name()` with `mysqld_exporter:` prefix: `mysqld_exporter:perf_schema.eventswaits`, etc. That makes it clear that (1) this query comes from `mysqld_exporter`, and (2) what flag should be used to disable this scraper.

SuperQ:
@AlekSi That sounds fine to me.

PR #331, opened 2018-09-05
[WIP] Add query tags.

AlekSi:
Refs #241.

@SuperQ PTAL. Before I went too far – does it looks good?

### DIFF
--- a/collector/exporter.go
+++ b/collector/exporter.go
@@ -37,6 +37,10 @@ var (
 		"exporter.log_slow_filter",
 		"Add a log_slow_filter to avoid slow query logging of scrapes. NOTE: Not supported by Oracle MySQL.",
 	).Default("false").Bool()
+	tagQueries = kingpin.Flag(
+		"exporter.tag-queries",
+		"TODO",
+	).Default("false").Bool()
 )
 
 // Metric descriptors.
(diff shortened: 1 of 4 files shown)


collector/exporter.go
```diff
@@ -37,6 +37,10 @@ var (
 		"exporter.log_slow_filter",
 		"Add a log_slow_filter to avoid slow query logging of scrapes. NOTE: Not supported by Oracle MySQL.",
 	).Default("false").Bool()
+	tagQueries = kingpin.Flag(
+		"exporter.tag-queries",
+		"TODO",
+	).Default("false").Bool()
```
SuperQ: I don't think we need a flag for this, let's just add the feature so it's always on. :smile:

utdrmac: This flag should allow the enduser to pass N number of tags. Tags should not be static, and should be a Key/Value pair so that monitoring systems (such as Grafana) can filter on the tag and value.

ArthurSens:
Thank you for prototyping query tags and adding the accompanying test coverage. The exporter has changed substantially since this WIP began, so the branch can no longer be merged as-is. We are closing it as part of the backlog cleanup, while preserving the implementation and discussion as useful prior work. If query tags are pursued again, we would welcome a fresh proposal or PR based on the current collector implementation.

PR #333, opened 2018-09-11
add metrics to monitor processes by user and host

peterloeffler:
Added 2 new flags --collect.info_schema.processlist.processes_by_user and --collect.info_schema.processlist.processes_by_host which are needed in addition to --collect.info_schema.processlist.
By using them you get 2 new metrics:
- mysql_info_schema_processes_by_user{src_user=""}
- mysql_info_schema_processes_by_host{src_host=""}

Found it very useful to have this metrics and thought it could be interesting to have it in place for many people.

Closes: https://github.com/prometheus/mysqld_exporter/issues/334

### DIFF
--- a/collector/info_schema_processlist.go
+++ b/collector/info_schema_processlist.go
@@ -32,6 +32,14 @@ var (
 		"collect.info_schema.processlist.min_time",
 		"Minimum time a thread must be in each state to be counted",
 	).Default("0").Int()
+	processesDisableByUserFlag = kingpin.Flag(
+		"collect.info_schema.processlist.disable_processes_by_user",
+		"Disable collecting the number of processes by user",
+	).Default("false").Bool()
+	processesDisableByHostFlag = kingpin.Flag(
+		"collect.info_schema.processlist.disable_processes_by_host",
+		"Disable collecting the number of processes by host",
+	).Default("false").Bool()
 )
 
 // Metric descriptors.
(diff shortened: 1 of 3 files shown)


collector/info_schema_processlist.go
```diff
@@ -121,6 +147,52 @@ var (
 	}
 )
 
+func processesByUser(db *sql.DB, ch chan<- prometheus.Metric, query string, i int) error {
```
SuperQ: What is the `i int` for? It seems unused.

SuperQ: Also, I don't think it's necessary to pass the package scope `query string`.

peterloeffler:
Only executing one query sounds like a good idea.
I'm not sure if my go skills are good enough to implement this properly.

SuperQ:
Let's document this feature request in an issue.

SuperQ:
Let me know if #334 is what you're looking for.

peterloeffler:
Sounds perfect!

SuperQ:
One more request, please add a CHANGELOG entry.

peterloeffler:
> Nice, looks pretty good. What do you think about keeping a couple bool flags to disable host/user metrics. We can leave them on by default, but some people might have a lot of users/hosts where the cardinality would be too high, but they may still want command/state metrics.

Good idea. Is it ok that the additional flags are not listet when calling --help?

SuperQ:
I would probably list them in `--help`, like the other collector tunables are.

peterloeffler:
Added the flags + a hopefully proper changelog entry.

peterloeffler:
What do you think? Can we merge it? Anything I can do?

SuperQ:
Sorry, got lost in my review pile.

collector/info_schema_processlist.go
```diff
@@ -26,6 +32,14 @@ var (
 		"collect.info_schema.processlist.min_time",
 		"Minimum time a thread must be in each state to be counted",
 	).Default("0").Int()
+	processesDisableByUserFlag = kingpin.Flag(
+		"collect.info_schema.processlist.disable_processes_by_user",
```
SuperQ: I prefer positive flags when possible, to avoid confusing double negatives.

Let's use `collect.info_schema.processlist.processes_by_user`, with a default of true.  Same for by_host.

This way the disable flag will be `--no-collect.info_schema.processlist.processes_by_user`.

peterloeffler:
No problem. Just wanted to know...
Flags are positive now.

peterloeffler:
Very nice! Thx a lot!

PR #336, opened 2018-09-24
Use GLOBAL to prevent mysql deadlock

codytaylor:
@SuperQ 
In mysql 5.7.22 we encountered an issue where the entire process would completely lock up. We were able to solve it by modifying this one file and also removing the '--collect.binlog_size' param from the mysqld exporter.

Here is the bug report: https://bugs.mysql.com/bug.php?id=92108

### DIFF
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -15,7 +15,7 @@ const (
 	// Math constant for picoseconds to seconds.
 	picoSeconds = 1e12
 	// Query to check whether user/table/client stats are enabled.
-	userstatCheckQuery = `SHOW VARIABLES WHERE Variable_Name='userstat'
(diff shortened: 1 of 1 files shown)


SuperQ:
Very interesting, seems reasonable to me.

SuperQ:
Please sign your commit with `git commit -s --amend`.

SuperQ:
Hrm, it looks like your commit history is a little strange now.

codytaylor:
Yeah, trying out a new tool called sublime merge. Looks like it did a pull/merge before I pushed the signed commit. No further changes though. 

Need me to figure out how to remove that?

SuperQ:
Yes, it would be nice. It seems like the commit history is making the DCO bot confused.

codytaylor:
Looks good now

PR #341, opened 2018-10-25
add scraping of mysql.user

peterloeffler:
Hi!

Would it make sense to add information from mysql.users?

Why we want to have this:
We want to have an alert based on something like `min by(hostname,src_user)(predict_linear(mysql_info_schema_processes_by_user[10m], 600)) > min by(hostname,src_user)(mysql_mysql_max_user_connections)` being aware of the fact that there can be multiple user/host combinations in the user table with different max_user_connections limits.
So we would be able to predict if a single user will use all his connections in a particular time frame.

No problem if there has to be some additional work to be done. Just wanted to ask if there is a chance that the pull request will be merged before spending too much time.

### DIFF
--- a/collector/mysql_user.go
+++ b/collector/mysql_user.go
@@ -1,0 +1,251 @@
+// Copyright 2018 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Scrape `mysql.user`.
+
+package collector
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	"strings"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"gopkg.in/alecthomas/kingpin.v2"
+)
+
+const mysqlUserQuery = `
+		  SELECT
+		    user,
+		    host,
+		    Select_priv,
+		    Insert_priv,
+		    Update_priv,
+		    Delete_priv,
+		    Create_priv,
+		    Drop_priv,
+		    Reload_priv,
+		    Shutdown_priv,
+		    Process_priv,
+		    File_priv,
+		    Grant_priv,
+		    References_priv,
+		    Index_priv,
+		    Alter_priv,
+		    Show_db_priv,
+		    Super_priv,
+		    Create_tmp_table_priv,
+		    Lock_tables_priv,
+		    Execute_priv,
+		    Repl_slave_priv,
+		    Repl_client_priv,
+		    Create_view_priv,
+		    Show_view_priv,
+		    Create_routine_priv,
+		    Alter_routine_priv,
+		    Create_user_priv,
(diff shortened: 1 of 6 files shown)


SuperQ:
Interesting idea. The whole CASE selection is a bit wonky, I think we would prefer to do this kind of matching in PromQL since we have the mysql global variables metric. You can do something like `mysql_mysql_max_user_connections > 0 or mysql_global_variables_max_user_connections`

peterloeffler:
I completely agree. Was not aware that this is possible in PromQL.
Because they are also INTs I would also add metrics for:
max_questions
max_updates
max_connections
What do you say?

SuperQ:
Yes, if we're going to pull data from the users table, might as well add those extra metrics.

I would also suggest label names `mysql_user` and `mysql_host` to make them less ambiguous.

peterloeffler:
When we use `mysql_user` I would also change the `src_user` lable from the `mysql_info_schema_processes_by_user` metric.

Providing the user privileges as `0` and `1` would be too much information, right?

SuperQ:
Agreed, `src_user` and `src_host` should be updated in collector/info_schema_processlist.go.

SuperQ:
Exposing the privilege bitmap would be a lot of data, maybe have a feature flag for it, off by default?

peterloeffler:
OK. Would you prefer to do the Y/N to 0/1 thing in SQL or in GO?

peterloeffler:
> Agreed, `src_user` and `src_host` should be updated in collector/info_schema_processlist.go.

changing `src_user` to `mysql_user` sounds ok but `src_host` to `mysql_host` would make no sense in collector/info_schema_processlist.go because it describes the source host i would say.

SuperQ:
I think we need to have non-matching label names between the two. In the processlist, it's more like `client_host`, but the user spec is more like `hostmask`, because it could be `%`. In order to use it sanely, you'll have to aggregate the host label away and just match on users most of the time.

peterloeffler:
If anybody is interested: the alerting rule could look like `min by(hostname,mysql_user)(predict_linear(mysql_info_schema_processes_by_user[10m], 600)) > min by(hostname,mysql_user)(mysql_mysql_max_user_connections > 0 or (mysql_global_variables_max_user_connections > 0 or mysql_global_variables_max_connections))`
Don't forget to use the 2 flags --collect.mysql.user and --collect.info_schema.processlist.

peterloeffler:
Do I have to fix the sign off issues too for the pull request to be accepted or is it fine the way it is?

collector/mysql_user.go
```diff
@@ -0,0 +1,362 @@
...
+
+	"github.com/prometheus/client_golang/prometheus"
+	"gopkg.in/alecthomas/kingpin.v2"
+)
+
+const mysqlUserQuery = `
+		  SELECT
+		    user,
+		    host,
+		    IF( Select_priv = 'N', 0, 1 ) AS select_priv,
```
SuperQ: It's better to handle this in the code, rather than create a larger messy SELECT statement. There's a `parseStatus()` helper function in collector.go. You could create a `parsePrivilege()` function to deal with the Y/N here.

peterloeffler:
Sorry! I'm too stupid to fix this signoff thing because of too many merge confilcts.
I think the cleanest solution would be if I delete my fork and recreate it.
Fine for you?

SuperQ:
Yea, re-fork, branch, and patch is totally fine. Just open a new PR when you're ready.

peterloeffler:
See follow up on pull request #343

SuperQ:
Closed in favor of https://github.com/prometheus/mysqld_exporter/pull/343

Issue #345, opened 2018-11-14
Update to go-sql-driver/mysql 1.4.1


foosinn:
Hey,

go-sql-driver/mysql just released 1.4.1. Is it possible to rebuild the official package against the latest version?

https://github.com/go-sql-driver/mysql/releases/tag/v1.4.1

## Version 1.4.1 (2018-11-14)

Bugfixes:

 - Fix TIME format for binary columns (#818)
 - Fix handling of empty auth plugin names (#835)
 - Fix caching_sha2_password with empty password (#826)
 - Fix canceled context broke mysqlConn (#862)
 - Fix OldAuthSwitchRequest support (#870)
 - Fix Auth Response packet for cleartext password (#887)

SuperQ:
I'm traveling right now, but there's a few good reasons to cut a new release soon.

PR #346, opened 2018-11-14
Update to go-sql-driver/mysql 1.4.1

foosinn:
Updated go-sql-driver/mysql. Automated tests and a short manual test look good.

I am not a go programmer so i would appreciate a quick sanity check before merging. 

Thanks!

Closes: https://github.com/prometheus/mysqld_exporter/issues/345

### DIFF
--- a/vendor/vendor.json
+++ b/vendor/vendor.json
@@ -31,12 +31,12 @@
 			"revisionTime": "2016-08-04T10:47:26Z"
 		},
 		{
-			"checksumSHA1": "kxzfnN+IPF0sViI9iiUnI7aZVWM=",
+			"checksumSHA1": "yg01DVQJjQgMeXf9C7Fpl2Yo5+Y=",
 			"path": "github.com/go-sql-driver/mysql",
-			"revision": "d523deb1b23d913de5bdada721a6071e71283618",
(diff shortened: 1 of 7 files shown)


vendor/vendor.json
```diff
@@ -31,10 +31,10 @@
...
 		{
-			"checksumSHA1": "kxzfnN+IPF0sViI9iiUnI7aZVWM=",
+			"checksumSHA1": "ANMYkuezaW7Pl/JJBiEs0KLzTOg=",
 			"path": "github.com/go-sql-driver/mysql",
-			"revision": "d523deb1b23d913de5bdada721a6071e71283618",
-			"revisionTime": "2018-06-03T12:45:54Z",
+			"revision": "369b5d6e5e8e108ed4ae2f2b1607d444b3807dfb",
+			"revisionTime": "2018-11-13T02:38:49Z",
 			"version": "v1.4",
 			"versionExact": "v1.4.0"
```
SuperQ: I think you need to run this with `govendor fetch github.com/go-sql-driver/mysql@v1.4`. This should correctly update all the vendor metadata.

foosinn: oh, you are right. seems like i updated to the latest master commit. i fixed it and pushed the changes.

thanks for pointing that out.

PR #347, opened 2018-11-15
Add perf_schema.eventsstatementssum collector

alexey-medvedchikov:
This PR adds collector with aggregated perf_schema.eventstatements metrics (when it makes sense).

The reason why I did this is that we have a lot of MySQL servers with different SQL queries and multiplication X by Y gives us an enormous number of series. We would like to monitor this values still, but it requires either complex setup of Prometheus servers with the federation, either switching off this collector on most of the servers.

Having this collector enables allows us to have an overview of all performance events, and thanks to `collect[]` parameter feature easily enable extra verbose collector on problematic ones.

Basically, this code is copied and modified `collector/perf_schema_events_statements.go` one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,12 +38,12 @@ Running using ~/.my.cnf:
     ./mysqld_exporter <flags>
 
 Example format for flags for version > 0.10.0:
-  
+
     --collect.auto_increment.columns
     --no-collect.auto_increment.columns
-  
+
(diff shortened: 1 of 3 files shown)


SuperQ:
This needs a rebase.

SuperQ:
Sorry, I didn't notice this had been re-based.

PR #352, opened 2018-11-28
*: support Go modules

simonpasquier:


go.mod
```diff
@@ -0,0 +1,36 @@
...
+	github.com/go-sql-driver/mysql v1.4.1
+	github.com/gogo/protobuf v1.1.1 // indirect
+	github.com/gopherjs/gopherjs v0.0.0-20170609002610-dc374d327045 // indirect
+	github.com/jtolds/gls v4.2.1+incompatible // indirect
+	github.com/kr/pretty v0.1.0 // indirect
+	github.com/matttproud/golang_protobuf_extensions v1.0.1 // indirect
+	github.com/onsi/ginkgo v1.7.0 // indirect
+	github.com/onsi/gomega v1.4.3 // indirect
+	github.com/pmezard/go-difflib v1.0.0 // indirect
+	github.com/prometheus/client_golang v0.9.0
```
SuperQ: Looks like this is pinned incorrectly. It should be `v0.9` so it locks at `v0.9.1`

PR #363, opened 2019-02-14
Add scraping Query Response Time Total

adivinho:
It is usefull in order to draw a histogram Query Load by Response Average Time

screenshot_20190118_145045

### DIFF
--- a/collector/info_schema_query_response_time.go
+++ b/collector/info_schema_query_response_time.go
@@ -53,6 +53,24 @@ var (
 			[]string{}, nil,
 		),
 	}
+
+	infoSchemaQueryResponseTimeTotalDescs = [3]*prometheus.Desc{
+		prometheus.NewDesc(
+			prometheus.BuildFQName(namespace, informationSchema, "query_response_time_microseconds_total"),
+			"The total execution time in microseconds of all queries by duration they took to execute.",
+			[]string{}, nil,
+		),
+		prometheus.NewDesc(
+			prometheus.BuildFQName(namespace, informationSchema, "read_query_response_time_microseconds_total"),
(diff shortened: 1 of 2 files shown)


SuperQ:
No, this isn't necessary. For histograms we only care about the counts. We roll up the count and total into `_count` and `_sum` metrics.

For the heatmap display, you only need the counts per bucket. The totals per bucket don't help with this.

SuperQ:
I think what you might be missing is the Grafana setting for how to display the data from Prometheus.

Under the "Axes" tab, change the "Data format" to "Time Series Buckets". The Grafana default of "Time Series" is incorrect for displaying Prometheus histograms.

adivinho:
> No, this isn't necessary. For histograms we only care about the counts. We roll up the count and total into `_count` and `_sum` metrics.
> 
> For the heatmap display, you only need the counts per bucket. The totals per bucket don't help with this.

Let me shown an example when new metrics are useful.
It was occurred one long-running query.
screenshot_20190215_125148

And this query can be easily noticed on the first heatmap (mysql_info_schema_query_response_time_microseconds_total_bucket) but it's absolutely missed on the second one (mysql_info_schema_query_response_time_seconds_bucket).
 
screenshot_20190215_122842

SuperQ:
Like I said previously, this is possible with the existing histogram data. What you are proposing is an invalid histogram. Histogram buckets contain the count of values within the bucket, not the value change within the bucket.

Your query should be:

```
sum by (le) (rate(mysql_info_schema_query_response_time_seconds_bucket[$__interval]))
```

Make sure to set a reasonable min step (3-4x your scrape interval) and use "Time Series Buckets" display mode.

PR #366, opened 2019-02-22
Collect metrics from replication_applier_status_by_worker

alantang888:
https://dev.mysql.com/doc/refman/8.0/en/replication-applier-status-by-worker-table.html
Also calculate replication lag by `NOW()-APPLYING_TRANSACTION_ORIGINAL_COMMIT_TIMESTAMP`

@SuperQ


collector/perf_schema_replication_applier_status_by_worker.go
```diff
@@ -0,0 +1,223 @@
...
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const perfReplicationApplierStatsByWorkerQuery = `
+	SELECT CHANNEL_NAME, WORKER_ID, 
+		LAST_APPLIED_TRANSACTION_ORIGINAL_COMMIT_TIMESTAMP, LAST_APPLIED_TRANSACTION_IMMEDIATE_COMMIT_TIMESTAMP, LAST_APPLIED_TRANSACTION_START_APPLY_TIMESTAMP, LAST_APPLIED_TRANSACTION_END_APPLY_TIMESTAMP,
+		APPLYING_TRANSACTION_ORIGINAL_COMMIT_TIMESTAMP, APPLYING_TRANSACTION_IMMEDIATE_COMMIT_TIMESTAMP, APPLYING_TRANSACTION_START_APPLY_TIMESTAMP,
+		(NOW()-APPLYING_TRANSACTION_ORIGINAL_COMMIT_TIMESTAMP) as REPLICAION_LAG
```
SuperQ: We don't pre-calculate things with the exporter in general. We allow users to calculate this themselves with `time() - timestamp_metric` in PromQL.

collector/perf_schema_replication_applier_status_by_worker.go
```diff
@@ -0,0 +1,223 @@
...
+    FROM performance_schema.replication_applier_status_by_worker
+	`
+
+// Metric descriptors.
+var (
+	performanceSchemaReplicationApplierStatsByWorkerLastAppliedTransactionOriginalCommitTimestampDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, performanceSchema, "last_applied_transaction_original_commit_timestamp"),
```
SuperQ: Please name these metrics with their unit, `_seconds`.

collector/perf_schema_replication_applier_status_by_worker.go
```diff
@@ -0,0 +1,223 @@
...
+	"context"
+	"database/sql"
+	"strconv"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const perfReplicationApplierStatsByWorkerQuery = `
+	SELECT CHANNEL_NAME, WORKER_ID,
```
SuperQ: To make this easier to read and avoid long lines, let's format it with  a column on each line. See [this example](https://github.com/prometheus/mysqld_exporter/blob/master/collector/perf_schema_table_lock_waits.go#L25-L50)

collector/perf_schema_replication_applier_status_by_worker.go
alantang888: Done

collector/perf_schema_replication_applier_status_by_worker.go
```diff
@@ -0,0 +1,207 @@
+// Copyright 2018 The Prometheus Authors
```
SuperQ: This should be 2019. :smile:

collector/perf_schema_replication_applier_status_by_worker.go
```diff
@@ -0,0 +1,207 @@
...
+// Metric descriptors.
+var (
+	performanceSchemaReplicationApplierStatsByWorkerLastAppliedTransactionOriginalCommitSecondDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, performanceSchema, "last_applied_transaction_original_commit_second"),
+		"A timestamp shows when the last transaction applied by this worker was committed on the original master.",
+		[]string{"channel_name", "member_id"}, nil,
+	)
+
+	performanceSchemaReplicationApplierStatsByWorkerLastAppliedTransactionImmediateCommitSecondDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, performanceSchema, "last_applied_transaction_immediate_commit_second"),
```
SuperQ: Minor typo, these should be plural, `_seconds`.

SuperQ: This is why we usually call this `_time_seconds` or `_timestamp_seconds`.

collector/perf_schema_replication_applier_status_by_worker.go
```diff
@@ -0,0 +1,207 @@
...
+		} else {
+			lastAppliedTransactionImmediateCommitSecond = 0
+		}
+		ch <- prometheus.MustNewConstMetric(
+			performanceSchemaReplicationApplierStatsByWorkerLastAppliedTransactionImmediateCommitSecondDesc,
+			prometheus.GaugeValue, lastAppliedTransactionImmediateCommitSecond, channelName, workerId,
+		)
+
+		if lastAppliedTransactionStartApplyTime.Nanosecond() != 0 {
+			lastAppliedTransactionStartApplySecond = float64(lastAppliedTransactionStartApplyTime.Unix())
```
SuperQ: Does MySQL return a more accurate time here? We typically use `float64(timeVal.UnixNano()) / 1e9` to get sub-second precision out of timestamps.

collector/perf_schema_replication_applier_status_by_worker.go
alantang888: is `_timestamp_seconds` more fit for this case?

SuperQ: Yes, I would say `_timestamp_seconds` is the best for this case.

SuperQ:
Looks mostly good, thanks for all the changes. I'm traveling today so I'll have to finish up the review when I'm back online.

alantang888:
Thx a lot.

alantang888:
Hi @SuperQ Is this PR ok?

alantang888:
@SuperQ May I know this PR need to modify or anything I need to do?

SuperQ:
Sorry, I've been traveling, got behind in my review TODO list.

Two items:
* Please add a `[FEATURE]` item to the unreleased section of the CHANGELOG.md.
* Please add a test.  You can see other tests like `perf_schema_index_io_waits_test.go`.

alantang888:
Thanks. I'm working on it.

alantang888:
Added test case and change to use `IsZero()` to determine is it a `0` Time. `Nanosecond()` is just the nano second. Not current in nano second form.

alantang888:
@SuperQ Added feature item on CHANGELOG.md and test case. Also fixed a bug. Hope it is ok.

SuperQ:
It looks like this feature now requires changes to the connection DSN string. We should make this requirement transparent to the user.

There are several options for doing this.
1. Update the DSN string to include `parseTime`.
2. Update the DSN string automatically when this collector is enabled.
3. Handle the time parsing internally.

I would prefer option 1 or 3, to provide consistency.

alantang888:
@SuperQ  I handle time parse instead of change DSN. Because I don't know is there any collector already parse by itself. I afraid if I add parseTime on DSN will make that fail.
But this also have risk if someone/user add `parseTime=true` on DSN.

SuperQ:
No users should be modifying the DSN that way.

nirgilboa:
Hi :) I think this may not be compatible with 5.7 as the schema is different.
Maybe just update the readme that this collector is supported on 8.0 ?

PR #369, opened 2019-03-04
Innodb buffer pool mapping

maeserichar:
This is a port of a pull request made in the percona fork of this exporter:

https://github.com/percona/mysqld_exporter/pull/20

Description of the problem in https://jira.percona.com/browse/PMM-1453

As a quick summary, the stats for Innodb_buffer_pool_pages_old and Innodb_buffer_pool_pages_total	should be returned as gauges, instead of counters.

@SuperQ


collector/global_status.go
```diff
@@ -139,7 +139,7 @@ func (ScrapeGlobalStatus) Scrape(ctx context.Context, db *sql.DB, ch chan<- prom
 				)
 			case "innodb_buffer_pool_pages":
 				switch match[2] {
-				case "data", "dirty", "free", "misc":
+				case "data", "dirty", "free", "misc", "old", "total":
```
SuperQ: We typically avoid including the "total" values, as they break the ability to do `sum(foo)` on metrics.

If the sum of all the cases is not the same as "total", we should expose a separate metric that contains the total.

maeserichar: The total is a sum of data, free and misc, so in this case is not the sum of all the others. 

When you said "a separate metric that contains the total", are you talking about the innodb_buffer_pool_pages metric tagged with "total"  or the equivalent of `sum(innodb_buffer_pool_pages)`?

maeserichar: Ok, been reading a bit about this and I think what you mean was avoid using "total" as a label. So, maybe with something like the following  we can avoid that situation:

```suggestion
				case "data", "dirty", "free", "misc", "old":
					ch <- prometheus.MustNewConstMetric(
						globalBufferPoolPagesDesc, prometheus.GaugeValue, floatVal, match[2],
					)
				case "total":
					ch <- prometheus.MustNewConstMetric(
						globalBufferPoolPagesDesc, prometheus.GaugeValue, floatVal, "total_pages",
					)
```

Not a fan of it, as there is no unified "format" for the labels, but should work. Another option would be adding the same suffix to all the labels for this metric, what do you think?

SuperQ: You can just leave out the total. We left it out of the list intentionally. It's unnecessary as it can be calculated with `sum(innodb_buffer_pool_pages)`.

maeserichar: I think that would give a different value from the one collected from mysql, as `sum(innodb_buffer_pool_pages)` would include `dirty`value,  while the `total` collected from mysql is only the sum of data, free and misc.

SuperQ: Whatever the correct labels are to equal total is what we should use here. Other things that are unrelated should be split out as separate metrics.

maeserichar: Changes applied, thanks!

SuperQ:
Oh, would you please update the PR description and add a CHANGELOG entry.

```
* [CHANGE] Update innodb buffer pool mappings #369 
```

maeserichar:
Changelog updated.

When are you expecting to release a new version?

SuperQ:
Sadly, this exporter is lower on my priority list, and I've been putting off making a release. I really need to do one soon. :frowning: 

I'm also working a plan to onboard some new maintainers, so features/releases will be more regular.

PR #372, opened 2019-03-08
Group table stats by regex substitution

maeserichar:
In our system, we are sharding data adding a suffix to the table names, so for us would be great to have the metrics from the info_schema_tablestats collector grouped by the table suffix, as some "logic" tables in our case have thousands of shards. 

So the idea is provide a regex with capture groups and a pattern that using those groups, build the "logical" name of the table, and group the stats by the name.

@SuperQ

### DIFF
--- a/collector/info_schema_tablestats_filtered.go
+++ b/collector/info_schema_tablestats_filtered.go
@@ -1,0 +1,175 @@
+// Copyright 2018 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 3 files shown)


brian-brazil:
This sounds like a case for metric_relabel_configs.

maeserichar:
In our case we prefer to avoid generating the metrics (and transmitting them through the network), as we are talking about a quite big number of them.

SuperQ:
* I agree with @brian-brazil, this seems out of scope for the exporter.
* Even if it is in scope, having this as a separate collector seems unnecessary, the existing `info_schema_schemastats.go` collector should be extended.

maeserichar:
Well, thanks for your time anyway :-)

PR #381, opened 2019-04-26
fix info_schema_processlist collector, binlog dump may cause ‘write to net‘ a huge number 

wuxu92:
when binlog dump thread in 'writing to net' state, the value of metric `'mysql_info_schema_threads_seconds_by_user{state='writing to net'}' `can be a very huge number, which firing meaningless alerts.

move binlog dump cmd check before nomrmal state checker.

### DIFF
--- a/collector/info_schema_processlist.go
+++ b/collector/info_schema_processlist.go
@@ -135,8 +135,8 @@ var (
 		"other":                     uint32(0),
 	}
 	threadStateMapping = map[string]string{
-		"user sleep":     "idle",
-		"creating index": "altering table",
+		"user sleep":                               "idle",
+		"creating index":                           "altering table",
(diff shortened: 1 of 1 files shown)


SuperQ:
That seems like a problem with the alert, not with the exporter. I don't see anything invalid about this.

wuxu92:
> That seems like a problem with the alert, not with the exporter. I don't see anything invalid about this.

IMHO, it would be better to add **all** binlog dump threads to `mysql_info_schema_threads_seconds_by_user{state='replication master'}`, rather than `mysql_info_schema_threads_seconds_by_user{state='writing to net'}` even this binlog dump thread in `writing to net` state. for this thead's time can be very huge and will shadow all user's 'writing to net' thread count

SuperQ:
Fixed with #603.

PR #395, opened 2019-06-11
Remove deprecated usage of prometheus/client_golang

beorn7:
Note that the next release of client_golang will remove all deprecated
features.

Note that this _will_ change the metrics about the exposition itself.


beorn7:
Got all the tests green. Failures were caused by flakiness not related to this change.

PR #412, opened 2019-07-22
Add support to monitor multiple instances at the same time

roman-vynar:
Hello,

This adds an ability to monitor multiple MySQL instances at once by a single exporter instance. I know this PR may come against some practices but I am sure it might be useful for some users.

This is really useful when you need to monitor RDS instances from a single box and you don't want to run a ton of mysqld_exporters and do a client config on nginx with multiple ports/paths and then Prometheus service discovery with the same. There is no way to pass a DSN via query params and it's not a good thing anyway.

Features:
* You can specify multiple DSN endpoints in `DATA_SOURCE_NAME` env var as comma-separated values.
* When multiple endpoints are configured, each metric gets a new label `endpoint={"hostname:port"}`.
* When 1 endpoint is configured, the exporter works as normal w/o extra labels keeping backward-compatibility
* Each metric description is built with whether `newDesc()` or `newDescLabels()` func which is now consistent across all scrapers (that's why there are ton of changes). 
* Tested to to monitor global_status and global_variables of 6 RDS instances at once
* The tests work too :)

P.S. postgres_exporter supports multiple instances, redis_exporter removed this but now there is an issue when you have the different passwords (of course, you are) and can't really monitor multiple instances, haproxy_exporter has a PR for that and now this :)

Thanks!


SuperQ:
This moves discovery out of Prometheus and into the exporter. We _really_ do not want to do this here.

The latest version of the redis_exporter follows the correct pattern. It allows the target host to be selected via a target URL parameter, similar to blackbox_exporter and snmp_exporter.

This allows for keeping the discovery in Prometheus, but makes it possible to run the exporter against several remote databases that you don't have host access to.

roman-vynar:
I understand about service discovery however it does not solve the security issue.

redis_exporter allows to send an URL via query params but assumes the password is the same for all redis instances. That's not what people do on production, i.e. using the same password across all database instances.

On mysql: the same thing. I can make it so we pass DSN via query args but how about passwords? Prometheus does the service discovery but it can't do password discovery, moreover sending passwords from server to client is also not a good thing from security perspective.

If we do another way - keeping only passwords aside, say mysqld_exporter will have some sort of password-to-instance map but the actual address to scrape comes from Prometheus then we are returning to the original problem: we still have to maintain the service "discovery" thing on the client side but in this case password-to-instance mapping.

roman-vynar:
Also Prometheus does not perform RDS discovery and this is MySQL case in particular.

SuperQ:
The proposal for the redis_exporter was to have passwords be kept in a mapping config file, similar to how we do it in the snmp_exporter.

Handling RDS discovery is out of scope for the exporter.

SuperQ:
This is probably not how we want to implement proxy scraping of multiple targets. With this method, there is no way to parallelize the scrapes, so you end up having one scrape to many targets. This doesn't scale well. By using a `target=` URL param like is done in the redis_exporter or snmp_exporter, we can process multiple scrapes simultaneously.

With this method, one slow scrape can ruin the scrape of several targets.

PR #417, opened 2019-07-30
Add option to skip tls hostname verification

### LINKED ISSUE #348: Unable to use skip-verify or equivalent when custom SSL is used
### Host operating system: output of `uname -a`
Centos 7.3
### mysqld_exporter version: output of `mysqld_exporter --version`
``` 
INFO[0000] Starting mysqld_exporter (version=0.11.0, branch=HEAD, revision=5d7179615695a61ecc3b5bf90a2a7c76a9592cdd)  source="mysqld_exporter.go:206"
```
### MySQL server version
5.7
### mysqld_exporter command line flags
/usr/local/bin/mysqld_exporter --config.my-cnf=$HOME/.my.cnf
### What did you do that produced an error?
Using a custom SSL configuration, if the SSL cert is invalid, there is no way to ignore "skip-verify". This is important because some cloud providers SSL certs for Mysql connections do not include ips in them.
### What did you expect to see?
An ability to connect using a custom cert and use "skip-verify" as documented here: https://golang.org/src/crypto/tls/common.go?s=12984:21448#L334. 
Also according to mysql documentation, it doesnt normally check SSL certs when connecting unless specified to https://dev.mysql.com/doc/refman/5.7/en/encrypted-connection-options.html#option_general_ssl-mode
  
### What did you see instead?
An error saying
```
Error pinging mysqld: x509: cannot validate certificate for X.X.X.X because it doesn't contain any IP SANs" source="exporter.go:119
```

markcorwin-iex:
This is useful in cases where the provided certs do not include hostname
information, such as GCP's SQL instances.

Closes #348

### DIFF
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -53,6 +53,10 @@ var (
 		"config.my-cnf",
 		"Path to .my.cnf file to read MySQL credentials from.",
 	).Default(path.Join(os.Getenv("HOME"), ".my.cnf")).String()
+	tlsSkipVerify = kingpin.Flag(
+		"tls.skip-verify",
+		"Ignore hostname verification when using a tls connection.",
+	).Bool()
 	dsn string
 )
 
@@ -150,6 +154,10 @@ func customizeTLS(sslCA string, sslCert string, sslKey string) error {
 		}
 		certPairs = append(certPairs, keypair)
 		tlsCfg.Certificates = certPairs
+
+		if *tlsSkipVerify {
+			tlsCfg.InsecureSkipVerify = true
+		}
 	}
 	mysql.RegisterTLSConfig("custom", &tlsCfg)
 	return nil
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -53,9 +53,9 @@ var (
 		"config.my-cnf",
 		"Path to .my.cnf file to read MySQL credentials from.",
 	).Default(path.Join(os.Getenv("HOME"), ".my.cnf")).String()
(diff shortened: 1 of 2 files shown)


mysqld_exporter.go
```diff
@@ -53,6 +53,10 @@ var (
 		"config.my-cnf",
 		"Path to .my.cnf file to read MySQL credentials from.",
 	).Default(path.Join(os.Getenv("HOME"), ".my.cnf")).String()
+	tlsSkipVerify = kingpin.Flag(
+		"tls.skip-verify",
+		"Ignore hostname verification when using a tls connection.",
```
SuperQ: This does more than ignore the hostname verification. It ignores all TLS errors.

mysqld_exporter.go
```diff
@@ -53,6 +53,10 @@ var (
 		"config.my-cnf",
 		"Path to .my.cnf file to read MySQL credentials from.",
 	).Default(path.Join(os.Getenv("HOME"), ".my.cnf")).String()
+	tlsSkipVerify = kingpin.Flag(
+		"tls.skip-verify",
```
SuperQ: This should be `tls.insecure-skip-verify` to match how we use this elsewhere.

SuperQ:
Please add a `[FEATURE]` entry to the changelog file.

markcorwin-iex:
Pushing separate commits just for ease of tracking but will squash once done/approved.

markcorwin-iex:
@SuperQ no rush on this but just wanted to bring changes to the forefront.

Issue #413, opened 2019-07-26
no binlog metrics in mysql 8.x


r0bj:
### mysqld_exporter version: output of `mysqld_exporter --version`
```
mysqld_exporter, version 0.11.0 (branch: HEAD, revision: 5d7179615695a61ecc3b5bf90a2a7c76a9592cdd)
  build user:       root@3d3ff666b0e4
  build date:       20180629-15:00:35
  go version:       go1.10.3
```

### MySQL server version
8.0.17

### mysqld_exporter command line flags

```
mysqld_exporter --collect.binlog_size
```

### What did you expect to see?
Metrics `mysql_binlog_*` are available.

### What did you see instead?
Metrics `mysql_binlog_*` are not available.

It may be because in mysql 8.x `SHOW BINARY LOGS` returns 3 columns instead of 2 in previous versions.

PR #419, opened 2019-08-05
fix binlog metrics on mysql 8.x

r0bj:
Mysql 8.x returns 3 columns during `SHOW BINARY LOGS` query instead of 2 in older mysql.

It fixes https://github.com/prometheus/mysqld_exporter/issues/413
It's based on @SuperQ comment in https://github.com/prometheus/mysqld_exporter/pull/390

### DIFF
--- a/collector/binlog.go
+++ b/collector/binlog.go
@@ -18,6 +18,7 @@ package collector
 import (
 	"context"
 	"database/sql"
+	"fmt"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


r0bj:
@SuperQ Those tests seems unstable, every time different one fails, some with `Failed to connect to github.com port 443: Connection timed out`.

SuperQ:
I'm going to adjust the Travis test setup to only run the integration test, rather than a full test as we do that in CircleCI.

https://github.com/prometheus/mysqld_exporter/pull/420

SuperQ:
Hrm, that didn't help at all. I'm guessing there's something wrong with Travis right now. It's usually reliable.

SuperQ:
Retried the tests, looks like the Travis instability is fixed.

PR #425, opened 2019-08-15
add a api named scrape for different mysql instance

soul12345:
Signed-off-by: soul12345 <735205371@qq.com>

### DIFF
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -160,6 +160,16 @@ func customizeTLS(sslCA string, sslCert string, sslKey string) error {
 	return nil
 }
 
+func getdsnByRequest(r *http.Request)string{
+	user := r.URL.Query().Get("user")
+	password := r.URL.Query().Get("pass")
+	host := r.URL.Query().Get("host")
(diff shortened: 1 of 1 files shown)


mysqld_exporter.go
```diff
@@ -160,6 +160,16 @@ func customizeTLS(sslCA string, sslCert string, sslKey string) error {
 	return nil
 }
 
+func getdsnByRequest(r *http.Request)string{
+	user := r.URL.Query().Get("user")
+	password := r.URL.Query().Get("pass")
```
SuperQ: Passing credentials over URL params is insecure. This is not allowed.

soul12345: thx for reply. How about  write credentials in config , url will carry a parameter key  for point credentials

SuperQ: Yes, that's more reasonable. This is how we do things with the snmp_exporter.

SuperQ:
There's no need to open and close PRs. Simply push new commits to the existing PR to make changes.

PR #435, opened 2019-09-20
Add collector for AWS Aurora information_schema.replica_host_status

sburke-at-ziprecruiter:
This adds a collector for metrics in Amazon Aurora's information_schema.replica_host_status table, that are suitable for Prometheus:

mysql_info_schema_replica_host_cpu_percent The CPU usage as a percentage.
mysql_info_schema_replica_host_lag_seconds The replica lag in seconds.
mysql_info_schema_replica_host_log_stream_speed The log stream speed in kilobytes per second.
mysql_info_schema_replica_host_replay_latency_seconds The current replay latency in seconds.
mysql_info_schema_replica_host_slave_latency_seconds The master - slave latency in seconds.

### DIFF
--- a/collector/info_schema_replica_host.go
+++ b/collector/info_schema_replica_host.go
@@ -1,0 +1,156 @@
+// Copyright 2018 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Scrape `information_schema.replica_host_status`.
+
+package collector
+
+import (
+	"context"
+	"database/sql"
+
+	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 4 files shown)


collector/info_schema_replica_host.go
```diff
@@ -0,0 +1,156 @@
...
+
+// Version of MySQL from which scraper is available.
+func (ScrapeReplicaHost) Version() float64 {
+	return 5.6
+}
+
+// Scrape collects data from database connection and sends it over channel as prometheus metric.
+func (ScrapeReplicaHost) Scrape(ctx context.Context, db *sql.DB, ch chan<- prometheus.Metric) error {
+	var count int64
+	err := db.QueryRowContext(ctx, replicaHostCheckQuery).Scan(&count)
```
SuperQ: Rather than have a separate check query, we can look into the `err`.

Something like this could be added to the replicaHostQuery
```go
if mysqlErr, ok := err.(*mysql.MySQLError); ok { // Now the error number is accessible directly
  // Check for 1109 Unknown table error.
  if mysqlErr.Number == 1109 {
    ...
```

collector/info_schema_replica_host.go
```diff
@@ -0,0 +1,156 @@
+// Copyright 2018 The Prometheus Authors
```
SuperQ:
```suggestion
// Copyright 2020 The Prometheus Authors
```

collector/info_schema_replica_host.go
sburke-at-ziprecruiter: Thanks! I've implemented your suggestions in a second commit.

PR #449, opened 2020-01-27
[WIP] Remove "check" queries

SuperQ:
Use the MySQL error code to determine if a feature is unavailable. This
reduces the number of queries per scrape.

Signed-off-by: Ben Kochie <superq@gmail.com>

### DIFF
--- a/collector/info_schema_tablestats.go
+++ b/collector/info_schema_tablestats.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
+	_ "github.com/go-sql-driver/mysql"
 	"github.com/prometheus/client_golang/prometheus"
 )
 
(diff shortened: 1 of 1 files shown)


collector/info_schema_tablestats.go
```diff
@@ -73,19 +74,14 @@ func (ScrapeTableStat) Version() float64 {
...
-	}
-	if varVal == "OFF" {
-		level.Debug(logger).Log("msg", "MySQL variable is OFF.", "var", varName)
-		return nil
-	}
-
 	informationSchemaTableStatisticsRows, err := db.QueryContext(ctx, tableStatQuery)
 	if err != nil {
+		if mysqlErr, ok := err.(*mysql.MySQLError); ok {
+			if mysqlErr.Number == 1109 {
```
Andor, line 80: Maybe it is better to not use "magic" constants? I.e https://github.com/VividCortex/mysqlerr or https://github.com/Zhwt/mysqlerr could be used to match error code, if you don't mind yet another dependency.

Issue #452, opened 2020-02-22
Scrape multiple MySQL servers from a single exporter instance


zoni:
At Castor, we run a somewhat non-uniform MySQL infrastructure:

- Some MySQL servers we have direct control over where we can run mysqld_exporter on the host itself.
- Some MySQL servers we have no access to, but we have a Prometheus server in the same network capable of reaching it that we do have control over.
- And some of these are really Azure MySQL and/or AWS RDS instances, which we'd still like to monitor with mysqld_exporter instead of relying on the cloud provider's native metrics so that we have a uniform set of metrics for our databases.

The current mysqld_exporter model of "one server per instance" is consistent with the typical Prometheus model, but it does present a usability problem. In our environment, this would require us to run multiple instances of the exporter on the same host on different ports and somehow encode this "MySQL server -> exporter port" mapping into the Prometheus scrape config.

If there was some way for us to tell mysqld_exporter which MySQL server to connect to at scrape time, we could greatly simplify all of this orchestration overhead (and as an added benefit, it would reduce memory usage from running multiple copies).

## Proposal

mysqld_exporter already supports reading from a `.my.cnf` file to get host, username, password and other connection details, but it's currently hard-coded to look for the `[client]` section.

Looking at the code, it appears to be relatively straightforward and non-invasive to optionally add an HTTP parameter (similar to how `collect[]` can be used to filter at scrape time already) to tell mysqld_exporter to get connection details under a key other than `client`.

We could then run a single mysqld_exporter instance on our prometheus servers, configured with a `.my.cnf` file containing connection details for all the hosts we're interested in under different keys.

Coupled with some relabeling (similar to the [strategy used with blackbox_exporter](https://github.com/prometheus/blackbox_exporter#prometheus-configuration)) we could then build a scrape configuration in Prometheus to scrape multiple MySQL servers using just a single mysqld_exporter instance.

I would be willing to make a PR for this, but would like to verify such a change stands a chance at being accepted before I invest time in trying to implement this.

SuperQ:
Adding a `target` parameter is something I've considered here, but haven't had time or motivation to implement.

Implementing a series of different hosts seems like it might not mesh well with the INI file that MySQL supports. It might be worth using a yaml file like the blackbox/snmp exporters to store multiple credentials keyed with a `module` param like those other exporters use.

So the exporter config would be something like:
```yaml
modules:
  default:
    username: exporter
    password: SomePass
  my_rds_account:
    username: rds_account
[...]

PR #454, opened 2020-03-05
Add target params to inject DSN from Prometheus

rrreeeyyy:
This makes it possible to overwrite `DATA_SOURCE_NAME` via `target` params like [blackbox_exporter](https://github.com/prometheus/blackbox_exporter).

## Motivation 

We want to monitor multiple MySQL databases (e.g. AWS RDS) with one mysqld_exporter.
Motivation is the same as https://github.com/prometheus/mysqld_exporter/pull/412 (We have 100+ RDSs, so we don't feel like managing 100+ mysqld_exporters).

And we also want to discover and monitor MySQL databases using Prometheus's service discovery.

## Prometheus configuration

Prometheus configuration is similar to the blackbox_exporter's one.

```
scrape_configs:
  - job_name: 'mysqld_exporter'
    static_configs:
      - targets:
        - user:password@tcp(my-mysql-network-1:3306)/
        - user:password@tcp(my-mysql-network-2:3306)/
    relabel_configs:
      - source_labels: [__address__]
        target_label: __param_target
      - source_labels: [__param_target]
        target_label: instance
      - target_label: __address__
        replacement: 127.0.0.1:9115  # The mysqld_exporter's real hostname:port.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,8 @@ The MySQL server's [data source name](http://en.wikipedia.org/wiki/Data_source_n
 must be set via the `DATA_SOURCE_NAME` environment variable.
 The format of this variable is described at https://github.com/go-sql-driver/mysql#dsn-data-source-name.
 
+`DATA_SOURCE_NAME` can overwrite via `target` query parameter.
+`http://localhost:9104/metrics?target=user:password@tcp(my-mysql-network:3306)/` will return metrics for a MySQL against `my-mysql-network:3306`.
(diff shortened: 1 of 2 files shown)


Andor:
Seems like related to https://github.com/prometheus/mysqld_exporter/issues/452 as well.

SuperQ:
Passing credentials over the URL is insecure. This will not be supported.

PR #459, opened 2020-03-31
Add collector for replication_group_members

nirgilboa:
Fixes #362 

The scraper sends a new Gauge metric with a static value of 1 and labels corresponding to columns of the replication_group_members table. I made this choice because the labels are very much static - if there's an idea for something more elegant I'm quite happy to change it :relaxed: 

Both 5.7 and 8.0 are supported, by way of a different query for each. The first query sent is used to determine whether the additional columns present in the 8.0 schema are there, and falls back to the 5.7 schema.

Hope this makes sense..

PR #462, opened 2020-04-09
Rewrite replication_group_member_stats collector

roman-vynar:
The existing `replication_group_member_stats` collector was written for MySQL 5.7.
With MySQL 8.x there are twice more columns. 
Especially interesting is `COUNT_TRANSACTIONS_REMOTE_IN_APPLIER_QUEUE` which is basically the replication delay (number of transactions) for a secondary member of Group Replication.

So what I did:
* support additional columns with MySQL 8.0
* backward compatible with MySQL 5.7
* if new columns are added in the future, the only change to be is to add one to the list
* no weird variable namings, compact code
* FIX: metric "transaction_in_queue" was Counter instead of Gauge
* BREAKING: renamed 3 old metrics starting with `mysql_perf_schema_transaction_` to start with `mysql_perf_schema_transactions_` to be consistent with column names
* BREAKING: exposing only server's own stats (matching MEMBER_ID with @@server_uuid), it does not make sense to expose other server stats to introduce confusing and you will be monitoring other servers separately anyway. Thus, member_id label is dropped.
* Add a test


nirgilboa:
LGTM
Also ran it against my staging cluster and seems to be working correctly

SuperQ:
Thanks. Mostly LGTM.

I would prefer to drop the `_count_` redundancy. Yes, it's in the column names, but we try and fix bad style in Prometheus. We prefer making them easier to understand for Prometheus users.

roman-vynar:
Ok, done. Updated PR description with the following:

BREAKING: renamed 3 old metrics starting with `mysql_perf_schema_transaction_` to start with `mysql_perf_schema_transactions_` to be consistent with column names.

collector/perf_schema_replication_group_member_stats.go
```diff
@@ -16,42 +16,51 @@ package collector
...
+	// In MySQL 5.7 these are the 4 first columns available. In MySQL 8.x all 8.
+	perfReplicationGroupMemberStats = map[string]struct {
+		vtype prometheus.ValueType
+		desc  *prometheus.Desc
+	}{
+		"COUNT_TRANSACTIONS_IN_QUEUE": {prometheus.GaugeValue,
+			prometheus.NewDesc(prometheus.BuildFQName(namespace, performanceSchema, "transactions_in_queue"),
+				"The number of transactions in the queue pending conflict detection checks.", nil, nil)},
+		"COUNT_TRANSACTIONS_CHECKED": {prometheus.CounterValue,
+			prometheus.NewDesc(prometheus.BuildFQName(namespace, performanceSchema, "transactions_checked"),
```
SuperQ, line 40: Sorry, another naming nitpick. :)

`CounterValue` metrics should be named `_total`.

roman-vynar:
Done.

Also if you can take a look to this other PR related to Group Replication https://github.com/prometheus/mysqld_exporter/pull/459

PR #465, opened 2020-04-23
Add performance_schema.status_by_host gauge

romuald:
This commit aims at adding a gauge measuring the contents of the
`performance_schema.status_by_host` table

Gauge name is `mysql_perf_schema_status_by_host`,
with labels `host` and `variable_name`

### DIFF
--- a/collector/perf_schema_status_by_host.go
+++ b/collector/perf_schema_status_by_host.go
@@ -1,0 +1,87 @@
+// Copyright 2020 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Scrape `performance_schema.status_by_host` column information.
+
+package collector
+
+import (
+	"context"
+	"database/sql"
+
+	"github.com/go-kit/kit/log"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const perfStatusByHostQuery = `
+	SELECT HOST, VARIABLE_NAME, VARIABLE_VALUE
+	FROM performance_schema.status_by_host
+	WHERE HOST IS NOT NULL;
+	`
+
+// Metric descriptors.
+var (
+	performanceSchemaStatusByHostDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, performanceSchema, "status_by_host"),
+		"The status variable value for a given host.",
+		[]string{"host", "variable_name"}, nil,
+	)
+)
+
+// ScrapePerfStatusByHost collects status_by_host table information.
+type ScrapePerfStatusByHost struct{}
+
+// Name of the Scraper. Should be unique.
+func (ScrapePerfStatusByHost) Name() string {
+	return "perf_schema.status_by_host"
+}
+
+// Help describes the role of the Scraper.
+func (ScrapePerfStatusByHost) Help() string {
+	return "Collect status_by_host gauges in performance_schema"
+}
+
+// Version of MySQL from which scraper is available.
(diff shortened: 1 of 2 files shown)


romuald:
Please note that this is my first PR on some Go code in general. I've simply copied one of the other modules, trying to rename things properly. 
I'm also unaware of some possible naming conventions used in the project

Tested locally with a Percona server 5.7

romuald:
poke @SuperQ because I forgot to read the CONTRIBUTING before opening the PR

SuperQ:
Can you describe more about what this table is useful for in a monitoring context?

romuald:
@SuperQ in our case, we had a some SQL clients doing "bad" queries and increasing the `read_rnd_next` handlers (`mysql_global_status_handlers_total{handler="read_rnd_next"}`)

Monitoring this table helped us to narrow the host / process behaving poorly

Did you mean to add this context in the commit or files too?

collector/perf_schema_status_by_host.go
```diff
@@ -0,0 +1,87 @@
...
+
+// Scrape `performance_schema.status_by_host` column information.
+
+package collector
+
+import (
+	"context"
+	"database/sql"
+
+	"github.com/go-kit/kit/log"
```
ArthurSens, line 22: we switched to another logging library

collector/perf_schema_status_by_host.go
```diff
@@ -0,0 +1,87 @@
...
+	`
+
+// Metric descriptors.
+var (
+	performanceSchemaStatusByHostDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, performanceSchema, "status_by_host"),
+		"The status variable value for a given host.",
+		[]string{"host", "variable_name"}, nil,
+	)
+)
```
ArthurSens, line 39: If I'm understanding this query correctly, it will produce one single metric with the following format

```
mysql_perf_schema_status_by_host{
  host="app1",
  variable_name="Bytes_received"
} 1000000

mysql_perf_schema_status_by_host{
  host="app1",
  variable_name="Handler_read_rnd_next"
} 5000

mysql_perf_schema_status_by_host{
  host="app1",
  variable_name="Questions"
} 800
```

Which could probably work, but it goes against our [naming conventions](https://prometheus.io/docs/practices/naming/#:~:text=...MUST%20refer%20to%20a%20single%20unit%20(e.g.%20do%20not%20mix%20seconds%20with%20milliseconds)%20and%20to%20a%20single%20quantity%20(e.g.%20do%20not%20mix%20request%20size%20with%20request%20duration).), mostly around one single metric having several different units.

A query like `sum by (host) (mysql_perf_schema_status_by_host)` would produce a result that makes absolutely no sense.

We should probably split this query into several different metrics, like one for each variable name you'd like to monitor.

collector/perf_schema_status_by_host.go
```diff
@@ -0,0 +1,87 @@
...
+
+import (
+	"context"
+	"database/sql"
+
+	"github.com/go-kit/kit/log"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const perfStatusByHostQuery = `
```
ArthurSens, line 26: Should we have control over MySQL versions? I'm not sure if this is available for everyone

collector/perf_schema_status_by_host.go
```diff
@@ -0,0 +1,87 @@
...
+	SELECT HOST, VARIABLE_NAME, VARIABLE_VALUE
```
ArthurSens, line 27: [Looking at docs](https://dev.mysql.com/doc/refman/8.4/en/performance-schema-hosts-table.html#:~:text=The%20hosts%20table%20contains%20a%20row%20for%20each%20host%20from%20which%20clients%20have%20connected%20to%20the%20MySQL%20server.), seems like host grows indefinitely as new hosts connect with the database. It can be truncated, but this exporter will retain already observed rows unless it is restarted.

This will eventually create high-cardinality metrics. It is good that this is disabled by default, but we definitely need to have proper documentation with warnings about high cardinality.

ArthurSens (commented):
Hi, sorry for the super-delayed review. I'll close it since it has been a very long time without activity and the codebase has evolved a lot in the last 6 years, but I think it is still a good change and would be awesome if a new PR is opened, if you're still willing to contribute.

Here's a review in case you consider opening a new PR:

ArthurSens:
That said, closing the PR. Please open a new one if you still have the bandwidth!

PR #471, opened 2020-05-26
Support heartbeats in UTC

koorgoo:
`pt-heartbeat` supports UTC for heartbeat timestamps.

> The heartbeat table requires at least one row. If you manually create the heartbeat table, then you must insert a row by doing:
> 
> ```sql
> INSERT INTO heartbeat (ts, server_id) VALUES (NOW(), N);
> ```
> or if using `--utc`:
> ```sql
> INSERT INTO heartbeat (ts, server_id) VALUES (UTC_TIMESTAMP(), N);
> ```

https://www.percona.com/doc/percona-toolkit/LATEST/pt-heartbeat.html#cmdoption-pt-heartbeat-utc

When I started playing with `mysqld_exporter` I noticed 7200 second gap between heartbeat metrics. I understood that MySQL server was using the system timezone (https://time.is/CET), but the heartbeat timestamps were in UTC. 

This patch allows configuring the heartbeat collector to be consistent in such cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,7 @@ collect.slave_hosts                                          | 5.1           | C
 collect.heartbeat                                            | 5.1           | Collect from [heartbeat](#heartbeat).
 collect.heartbeat.database                                   | 5.1           | Database from where to collect heartbeat data. (default: heartbeat)
 collect.heartbeat.table                                      | 5.1           | Table from where to collect heartbeat data. (default: heartbeat)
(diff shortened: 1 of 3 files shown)


SuperQ:
I can fix the codespell problem in master.

koorgoo:
@SuperQ could you say whether a new version will be released?

v0.12.1 was released on 29 Jul 2019. And some work has been done from that time.

SuperQ:
I can't say for sure, but I can take a look at making a new release soon.

koorgoo:
Thank you. That's good enough for me.

PR #473, opened 2020-05-28
Fixed output value of wsrep_cluster_status

adivinho:
There is a misprint in checking cluster status values.
One of statues is never checked. 

`node1:root@localhost> SHOW STATUS LIKE 'wsrep_cluster_status';
+----------------------+-------------+
| Variable_name        | Value       |
+----------------------+-------------+
| wsrep_cluster_status | non-Primary |
+----------------------+-------------+
1 row in set (0,00 sec)

$ curl -s -k http://localhost:42001/metrics | grep mysql_global_status_wsrep_cluster_status
$`


SuperQ:
Thanks, Would you mind re-basing this on master to pick up the build fixes?

Also, update the CHANGELOG.md with this:

```
* [BUGFIX] Fixed output value of wsrep_cluster_status #473
```

collector/collector.go
```diff
@@ -52,11 +52,11 @@ func parseStatus(data sql.RawBytes) (float64, bool) {
 	if bytes.Equal(data, []byte("Connecting")) {
 		return 0, true
 	}
-	// SHOW GLOBAL STATUS like 'wsrep_cluster_status' can return "Primary" or "Non-Primary"/"Disconnected"
+	// SHOW GLOBAL STATUS like 'wsrep_cluster_status' can return "Primary" or "non-Primary"/"Disconnected"
 	if bytes.Equal(data, []byte("Primary")) {
 		return 1, true
 	}
-	if bytes.Equal(data, []byte("Non-Primary")) || bytes.Equal(data, []byte("Disconnected")) {
+	if bytes.Equal(data, []byte("non-Primary")) || bytes.Equal(data, []byte("Disconnected")) {
```
SuperQ, line 59: Looking at the [documentation](https://galeracluster.com/library/documentation/monitoring-cluster.html)

> The node should only return a value of Primary. Any other value indicates that the node is part of a nonoperational component. This occurs in cases of multiple membership changes that result in a loss of quorum or in cases of split-brain situations.

I think we should change this block to be an `else`. Anything that doesn't match "Primary" should return 0.

SuperQ: CC @rtreffer, what do you think?

SuperQ: Oh, nevermind, we have to do it this way otherwise it short-circuits the function. :frowning_face:

SuperQ, line 59: I think this may have changed between versions, let's make it case insensitive.

```suggestion
	if strings.EqualFold(string(data), "non-Primary") || bytes.Equal(data, []byte("Disconnected")) {
```

SuperQ:
It looks like the rebase you did was not fast-forward. This caused all of the upstream changes to be added to this PR.

I would recommend resetting your fork and using ["For and Branch Workflow"](https://blog.scottlowe.org/2015/01/27/using-fork-branch-git-workflow/) to avoid funky git history.

SuperQ:
```
collector/collector.go:59:5: undefined: strings
```

"strings" needs to be added to the import list.

adivinho:
> "strings" needs to be added to the import list.
Done.

PR #478, opened 2020-06-15
Add tls.insecure-skip-verify flag to README.md

Gremash:
This flag isn't documented. I needed it to be able to connect to a Google CloudSQL instance that requires SSL. The flag was add to the code in Aug. 2019.


SuperQ:
It looks like your change is missing DCO sign-off. Please amend the commit with `git commit -s --amend`.

PR #184, opened 2017-02-21
Add an example of heartbeat metrics for slave lag

SuperQ:
Heartbeat[0] metrics allow for similar alerting symantics to "seconds
behind master" metrics.
* Add a recording rule to provide a similar gauge of seconds behind.
* Add an alert with the same style.

[0]: https://github.com/prometheus/mysqld_exporter/pull/183

PR #479, opened 2020-06-17
Add heartbeat lag metric

koorgoo:
Heartbeats are used to measure replication delay and alert on that.

While starting using `mysqld_exporter` I decided to create a recording rule to calculate an actual lag. But later I realized that the approach is completely incorrect - having multiple tiers of aggregation the recorded metric had "strange" values - it's unclear which samples were used in calculation.

```yaml
- record: mysql_heartbeat_lag_seconds
  expr: >
    mysql_heartbeat_now_timestamp_seconds - 
    mysql_heartbeat_stored_timestamp_seconds
```

mysqld_exporter_repl

_The recorded metric at the top. A common metric at the bottom._

I suggest adding another metric `mysql_heartbeat_lag_seconds` which is calculated by the collector at each scrape so that the metric could be processed anyhow without correctness loss.

The metric can be used in an alerting rule right away. For example:

```yam
- alert: HighReplicationDelay
  expr: quantile_over_time(0.8, mysql_heartbeat_lag_seconds[5m]) > 5
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,9 @@ Changes related to `replication_group_member_stats` collector:
 * renamed 3 metrics starting with `mysql_perf_schema_transaction_` to start with `mysql_perf_schema_transactions_` to be consistent with column names
 * exposing only server's own stats by matching MEMBER_ID with @@server_uuid resulting "member_id" label to be dropped.
 
+Changes related to `heartbeat` collector:
+* metric "mysql_heartbeat_lag_seconds" replaced "mysql_heartbeat_now_timestamp_seconds" #479
+
 ## 0.12.1 / 2019-07-10
 
 ### Changes:
--- a/collector/heartbeat.go
+++ b/collector/heartbeat.go
@@ -63,6 +63,11 @@ var (
 		"Timestamp of the current server.",
 		[]string{"server_id"}, nil,
 	)
+	HeartbeatLagDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, heartbeat, "lag_seconds"),
+		"Heartbeat lag of the current server.",
+		[]string{"server_id"}, nil,
+	)
 )
 
 // ScrapeHeartbeat scrapes from the heartbeat table.
@@ -138,6 +143,12 @@ func (ScrapeHeartbeat) Scrape(ctx context.Context, db *sql.DB, ch chan<- prometh
 			HeartbeatStoredDesc,
 			prometheus.GaugeValue,
 			tsFloatVal,
+			serverId,
+		)
+		ch <- prometheus.MustNewConstMetric(
+			HeartbeatLagDesc,
+			prometheus.GaugeValue,
+			nowFloatVal-tsFloatVal,
 			serverId,
 		)
 	}
--- a/collector/heartbeat.go
+++ b/collector/heartbeat.go
@@ -29,11 +29,11 @@ import (
 const (
 	// heartbeat is the Metric subsystem we use.
 	heartbeat = "heartbeat"
-	// heartbeatQuery is the query used to fetch the stored and current
-	// timestamps. %s will be replaced by the database and table name.
+	// heartbeatQuery is the query used to fetch the stored timestamp and lag
+	// seconds. %s will be replaced by the database and table name.
 	// The second column allows gets the server timestamp at the exact same
 	// time the query is run.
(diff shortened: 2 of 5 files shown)


koorgoo:
It is fair to say that "strange" values happened because of [sample_limit](https://www.robustperception.io/using-sample_limit-to-avoid-overload) - some samples of either `mysql_heartbeat_now_timestamp_seconds` or `mysql_heartbeat_stored_timestamp_seconds` were dropped that caused wrong calculations. 

Nevertheless I would prefer to have no data instead of misleading values. So I am sure this PR is still relevant.

SuperQ:
Sorry, but we try and avoid exactly this kind of metric in exporters. Pre-computed metrics that can be derived.

koorgoo:
Nobody said it will be easy...

I was thinking about refusing work on this PR. But I came up with a few arguments why it might be appropriate to accept the precomputed metric this time.

1. The heartbeat collector is disabled by default. If you have enabled it then you need to know about replication delay and act on it.
2. `mysql_heartbeat_(now|stored)_timestamp_seconds` can be helpful. For example, the metrics helped me to understand the 7200 seconds delta and #471 was open.
3. Although the metrics can be helpful individually they will be always used together for alerting. Always in the same expression.

In #184 @SuperQ declared `mysql_heartbeat_lag_seconds` proving that neither `now_timestamp` nor `stored_timestamp` is useable on its own. Consequently every user of the collector has to repeat the same step to start getting the benefit from heartbeats - subtract those metrics.

> Don't repeat yourself.

My belief in `mysql_heartbeat_lag_seconds` as an independent metric is so strong, that I will take the risk of asking people from #183 what they think. 

@roidelapluie @brian-brazil @grypyrg, please, give a feedback.

roidelapluie:
Thanks for pinging.

I think that this is unnecessary. This is trivial and costs even so few that it can also be done live in grafana-without recording rules. Two extra features were backed into prometheus since 2017 that makes this even more unnecessary: isolation (the two metrics will always come from the same scrape) and subqueries (you don't need the recording rules all the time).

In addition, your original problem was related to sample_limit, not by anything wrong with the calculation. Having that extra metric would not have helped.

koorgoo:
Thanks for mentioning isolation and subqueries. I will take my time to learn about them. 

I agree that a 3rd metric is excessive because other two can express it.

Can we consider dropping `now_timestamp_seconds` in favour of `stored_timestamp_seconds` & `lag_seconds` then? 

Pros:

- `now_timestamp_seconds` can still be calculated. Although it's useless on its own anyway.
- `lag_seconds` is present and available out of the box for alerting on.
  - A great step to a better user experience I care about the most. Because I am a such user.

Cons:

- Breaking backwards compatibility

roidelapluie:
I still do not see an advantage with this, also this is the expected behaviour. If we accept it, why should we refuse to pre-process all the mysql status values too?

koorgoo:
> Dima: it might be appropriate to accept the precomputed metric this time.
> Julien: I still do not see an advantage with this

`mysql_heartbeat_lag_seconds`, a metric used for alerting on replication delay, will be at hands out of the box. No need to add a recording rule, no need to type the same expression again and again.

> why should we refuse to pre-process all the mysql status values too?

If you mean `mysql_slave_lag_seconds = mysql_slave_status_seconds_behind_master - mysql_slave_status_sql_delay`, unreliable `seconds_behind_master` was clearly explained in https://github.com/prometheus/mysqld_exporter/pull/183#issuecomment-281150958.

--

I think the whole MR is about user experience. 
If everyone should subtract `now_timestamp` and `stored_timestamp` to get needed `lag`, why don't we export `lag` directly?!

After all the knowledge about `lag` is the only purpose of the collector.

roidelapluie:
mysql_heartbeat_now_timestamp_seconds > tell me if the MySQL server is on time and if the ouput I have for the lag is relevant
    mysql_heartbeat_stored_timestamp_seconds > tells me if replication & pt-heartbeat runs

If I only get the lag, it is difficult to make the difference between if replication is slow or stopped.
without the now_timestamp, it is difficult to know if the server is on time.

koorgoo:
> without the now_timestamp, it is difficult to know if the server is on time.

As you can see in the latest changes I suggest exporting 2 metrics:

- `mysql_heartbeat_stored_timestamp_seconds`
- `mysql_heartbeat_lag_seconds`

With the equation (`now_timestamp` - **`stored_timestamp`** = **`lag`**) knowing 2 variables we can calculate the third.

roidelapluie:
So we are back to the original issue :)

koorgoo:
Sad to contemplate that the PR has received no support. Of course, this happens.
Thanks for the feedback.

Issue #458, opened 2020-03-20
Connection fails when the password contains `%` or `?` characters


erolatreston:
MySQL is created using the `mysql_native_password` plugin.
If the password contains `%` or `?` characters, authentication seems to fail.

### Host operating system: output of `uname -a`
Linux foo.example.com 3.10.0-1062.12.1.el7.x86_64 #1 SMP Tue Feb 4 23:02:59 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

### mysqld_exporter version: output of `mysqld_exporter --version`
mysqld_exporter, version 0.11.0 (branch: HEAD, revision: 5d7179615695a61ecc3b5bf90a2a7c76a9592cdd)
  build user:       root@3d3ff666b0e4
  build date:       20180629-15:00:35
  go version:       go1.10.3

### MySQL server version
5.7.29

### mysqld_exporter command line flags
Environment=DATA_SOURCE_NAME=exp:Abc123!#%^&*()_+[;.?|@(localhost:3306)/

### What did you do that produced an error?
change password to contain `%` or `?`

### What did you expect to see?
connection established

### What did you see instead?
connection failed

gnufree:
Your password setting is too complex. You can consider removing some special symbols.

Issue #376, opened 2019-03-29
mysql passwords with # and ;


aliensmily:
### Host operating system: output of `uname -a`
Linux gisu665 4.4.175-94.79-default #1 SMP Thu Feb 21 16:03:55 UTC 2019 (047a6d3) x86_64 x86_64 x86_64 GNU/Linux
=> SUSE Linux Enterprise Server 12 (x86_64) VERSION = 12 PATCHLEVEL = 3

### mysqld_exporter version: output of `mysqld_exporter --version`
mysqld_exporter, version 0.11.0 (branch: HEAD, revision: 5d7179615695a61ecc3b5bf90a2a7c76a9592cdd)
  build user:       root@3d3ff666b0e4
  build date:       20180629-15:00:35
  go version:       go1.10.3

### MySQL server version
8.0.13 MySQL Community Server

### mysqld_exporter command line flags
/opt/mysql/product/exporters/mysqld_exporter --config.my-cnf=/home/mysqladm/.myMOND.cnf

cat /home/mysqladm/.myMOND.cnf
```
# Password in "
# Password contains ; and #
[client]
socket=/opt/mysql/admin/mond/config/mysql.sock
user=MON_NOCMONITOR
password=****
```

### What did you do that produced an error?
Starting the exporter with systemctl.

### What did you expect to see?
The option file is working fine when using mysql on shell.

```
mysqladm@host0815:~> mysql --defaults-file=/home/mysqladm/.myMOND.cnf --print-defaults
mysql would have been started with the following arguments:
--socket=/opt/mysql/admin/mond/config/mysql.sock --user=MON_NOCMONITOR --password=*****
mysqladm@host0815:~> mysql --defaults-file=/home/mysqladm/.myMOND.cnf
Welcome to the MySQL monitor.  Commands end with ; or \g.
Your MySQL connection id is 2294
Server version: 8.0.13 MySQL Community Server - GPL

Copyright (c) 2000, 2018, Oracle and/or its affiliates. All rights reserved.

Oracle is a registered trademark of Oracle Corporation and/or its
affiliates. Other names may be trademarks of their respective
owners.

Type 'help;' or '\h' for help. Type '\c' to clear the current input statement.

(MON_NOCMONITOR@localhost) [(none)]> \q
```

### What did you see instead?
[...]

Issue #303, opened 2018-07-17
Error 1049 Unknown database '&lock_wait_timeout=2' 


wk60807430:
### Host operating system: output of `uname -a`
localhost
### mysqld_exporter version: output of `mysqld_exporter --version`

0.11.0.linux-amd64
### MySQL server version
5.7.22
### mysqld_exporter command line flags

./mysqld_exporter --config.my-cnf=".my.cnf"  

### What did you do that produced an error?
i want to start mysqld_exporter
### What did you expect to see?
run
### What did you see instead?
ERRO[0012] Error pinging mysqld: Error 1049: Unknown database '&lock_wait_timeout=2'  source="exporter.go:119"

cat .my.cnf
[client]
host=localhost
port= 3306
user=root
password=*******

wk60807430:
its so wired pls help me

wk60807430:
yes man its have a '?' in it

SuperQ:
I thought so, that's a bug in the current way we handle building the connection string.

For now, I would recommend creating a separate user for the exporter that does not have a `?` in the password.

wk60807430:
ok，tks

nabadger:
@SuperQ would you consider re-opening this? 

This also impacts passwords that are generated too - it doesn't seem reasonable to force users to generate passwords without `?` 
:)

MahaGamal:
I hit the same issue with password have `@` after changing it working fine

pengxiaobo1234:
great!

PR #484, opened 2020-06-26
Fix handling of DSN extra-argument handling.

nabadger:
I believe this addresses the following issues

#458  #303  #376 

```
DATA_SOURCE_NAME="my-user:my-password-with&?-chars@(127.0.0.1:3306)/ ./mysqld_exporter
```

This fix will just look for `?` by checking only the chars that appear after the final `/` .

If there is no `/` in the DSN, then it's already being treated as an invalid DSN

```
invalid DSN: missing the slash separating the database name
```


nabadger:
Will re-open soon, messed something up...

PR #485, opened 2020-06-26
Fix handling of DSN extra arguments

nabadger:
I believe this addresses the following issues

#458  #303  #376 

```
DATA_SOURCE_NAME="my-user:my-password-with&?-chars@(127.0.0.1:3306)/ ./mysqld_exporter
```

This fix will just look for `?` by checking only the chars that appear after the final `/` .

If there is no `/` in the DSN, then it's already being treated as an invalid DSN

```
invalid DSN: missing the slash separating the database name
```

### DIFF
--- a/collector/exporter.go
+++ b/collector/exporter.go
@@ -92,7 +92,10 @@ func New(ctx context.Context, dsn string, metrics Metrics, scrapers []Scraper, l
 		dsnParams = append(dsnParams, sessionSettingsParam)
 	}
 
-	if strings.Contains(dsn, "?") {
+	dsnParts := strings.Split(dsn, "/")
+	lastDsnPart := dsnParts[len(dsnParts)-1]
+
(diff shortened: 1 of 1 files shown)


nabadger:
Marking as WIP as i spotted an issue:
```
mysqld_exporter: error: unrecognized log format "logger:stderr?json=true", try --help
```

```
      containers:
      - args:
        - /bin/mysqld_exporter
        - --log.format=logger:stderr?json=true
```

nabadger:
Makes sense - we are actually writing the `.my.cnf` file, it's just that the exporter takes the contents and parses them out into a DSN.

SuperQ:
The whole DSN thing in the Go mysql driver is not fun.

SuperQ:
We've reworked the exporter to use `mysql.Config()` to configure the DSN. The old `DATA_SOURCE_NAME` env var has been removed.

PR #504, opened 2020-10-14
Support for scraping multiple mysqld hosts.

samitpal:
Discussion at https://github.com/prometheus/mysqld_exporter/issues/452#issuecomment-701893399

Short summary:
This adds a new flag called `export-multi-hosts` and allows for passing a yaml file which contains the config related to this mode. Note that this change does not make any changes to the way mysqld_exporter works and as a result it would continue to use the ini format config when working in the non export-multi-hosts mode.

### DIFF
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -60,16 +60,16 @@ var (
 		"tls.insecure-skip-verify",
 		"Ignore certificate and server verification when using a tls connection.",
 	).Bool()
-    multiHostExporter = kingpin.Flag(
-        "export-multi-hosts",
-        "Setting it to true enables scraping multiple mysql hosts.",
-    ).Default("false").Bool()
-    multiHostExporterConfigFile = kingpin.Flag(
-        "config-multi-hosts",
-        "Path to yaml config file to fetch mysql client info. Used when export-multi-hosts is true.",
-    ).Default("config-multi.yml").String()
+	multiHostExporter = kingpin.Flag(
+		"export-multi-hosts",
+		"Setting it to true enables scraping multiple mysql hosts.",
+	).Default("false").Bool()
+	multiHostExporterConfigFile = kingpin.Flag(
+		"config-multi-hosts",
+		"Path to yaml config file to fetch mysql client info. Used when export-multi-hosts is true.",
+	).Default("config-multi.yml").String()
 	dsn string
-    configMulti *multiHostExporterConfig
+	configMulti *multiHostExporterConfig
 )
 
 // scrapers lists all possible collection methods and if they should be enabled by default.
@@ -223,18 +223,18 @@ func newHandler(metrics collector.Metrics, scrapers []collector.Scraper, logger 
 
 		registry := prometheus.NewRegistry()
 
-        if *multiHostExporter {
-            target := r.URL.Query().Get("target")
-            if target == "" {
-	            http.Error(w, "Target parameter is missing", http.StatusBadRequest)
-		        return
-	        }
-            var err error
-            if dsn, err = configMulti.formDSN(target); err != nil {
-                http.Error(w, err.Error(), http.StatusInternalServerError)
-                return
-            }
-        }
+		if *multiHostExporter {
+			target := r.URL.Query().Get("target")
+			if target == "" {
+				http.Error(w, "Target parameter is missing", http.StatusBadRequest)
+				return
+			}
+			var err error
+			if dsn, err = configMulti.formDSN(target); err != nil {
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+				return
+			}
+		}
 		registry.MustRegister(collector.New(ctx, dsn, metrics, filteredScrapers, logger))
 
 		gatherers := prometheus.Gatherers{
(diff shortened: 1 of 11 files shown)


SuperQ:
My only question is, because mysql users are used to the INI style of my.cnf, maybe we should stick to INI parsing. I would prefer to implement one kind of config for the exporter. Having multiple ways to configure gets confusing for users.

I also think `port` should come from the target URL param, not the config.

I'm thinking a config something along the lines of:
```yaml
clients:
  default:
    user: foo
    password: foo123
  other_server:
    user: bar
    password: bar123
```

Or in INI style
```
[client]
user = foo
password = foo123
[client.other_server]
user = bar
password = bar123
```

samitpal:
Makes sense. Given that the currently supported format is INI, I will try to modify my change to use INI instead of yaml.

mysqld_exporter.go
```diff
@@ -264,11 +286,24 @@ func main() {
 	level.Info(logger).Log("msg", "Starting msqyld_exporter", "version", version.Info())
 	level.Info(logger).Log("msg", "Build context", version.BuildContext())
 
-	dsn = os.Getenv("DATA_SOURCE_NAME")
-	if len(dsn) == 0 {
+	if !*multiHostExporter {
```
SuperQ, line 289: I have been wanting to drop the `DATA_SOURCE_NAME` mode for this exporter, as it complicates things and has a number of bugs (special chars in passwords for example).

I also would prefer to have one config parsing mode.

SuperQ:
It looks like some of the commits are missing DCO.

mysqld_exporter.go
SuperQ: Maybe we can start with this pattern, but I think I may want to refactor it in short order.

samitpal: - Tried removing the `DATA_SOURCE_NAME` env variable.  Trying to fix the tests.
- With regards to 2 modes of operation, perhaps we could run in this way with a deprecation notice about the old mode? This  might give the users some time for familiarization. And then the old mode can be removed in a future release? But i do agree that running in one mode of operation is always better.

SuperQ: We should make it so that the new mode (multi-target) also supports one default case with no `target` URL param.

One thing I also think we need to do is make it so the auth part is specified by a 'module' param. This way the user/pass required for remote servers doesn't need to be repeated for each target.

This would allow for a config that only contains one default username and password to work for as many targets, as long as they all support the same password. It avoids the need to adjust the exporter config in the case of adding new targets.

samitpal: - Not sure about the support for default case with no target URL param.. Is that to handle the localhost condition?
- As far as the auth question is concerned I think that support is there in this change already. Basically we check for the target specific client in the config failing which we try the parent client (the `client` section). So if the auth is same across the fleet  one can just set up the config as follows. Perhaps it needs to be documented better in the README.

```
[client]
user = foo
password = foo123
```

RichiH:
YAML would have the advantage of being well-established in Prometheus-land; the question is if users are more likely to be DBAs exposing data or Prometheus people grabbing data from a lot of different places. I would suspect the latter, but don't have data, of course.

SuperQ:
@RichiH Supporting the MySQL-style INI file was an often, and now implemented, requested feature. I think we want to stick with it.

SuperQ:
Here's a proposal on how the final config (one config to rule them all) would look like.

```
[client]
user = foo
password = foo123
target = hostname:3306
[client.module1]
user = bar
password = bar123
```

The `target` key is only allowed in the default case and is ignored/rejected by additional modules.

* `/metrics` URL would be a backwards equivalent of `/mysql?module=default`.
* `/mysql?module=default&target=some-remote:3306` would use the default user/password but override the target.
* `/mysql?module=module1` is invalid, as it doesn't contain a target.
* `/mysql?module=module1&target=some-remote:3306` is a typical scrape of a remote.

I'm also thinking about an option to make `/metrics` not return mysql data, just exporter data. This would make it easier to separate exporter data from target data. Easier to monitor the health of just the exporter.

piotr1212:
Any progress on this? Would be very nice to have this.

yum-dev:
When is this being merged? We are waiting for this too.

rpsadarangani:
Hi, When this will get merge, we are also waiting for this.

SuperQ (changes requested):
This is not ready for merge, as the requested changes in https://github.com/prometheus/mysqld_exporter/pull/504#issuecomment-714592229 need to be addressed.

samitpal:
For our internal use case my implementation is sufficient. I won't be able to work on the proposal as outlined at #504 (comment). If any one else wants to take this forward, please feel free to go ahead.

SuperQ:
Alright, if I can find some help to work on this, we could merge it and build on it. But as-is, this needs work before it could be released.

Dims-lab:
Is it possible to use prometheus exporter when mysql is running on a different port then 3306? When i change the /etc/.mysqld_exporter.cnf file for looking at another port. it still tries to connect with port 3306. changes i have made are.

[client]
user=mysqld_exporter
password=apassword
host=localhost:portnumber

level=error msg="Error pinging mysqld: dial tcp: lookup localhost:45654:3306:

ning1875:
I have an improved version here, transformed into a probe, which supports http incoming parameters like blackbox_exporter . https://github.com/ning1875/mysqld_exporter/releases/tag/v0.12.1

SuperQ:
This was completed in #651

PR #505, opened 2020-10-15
fix log output of collect[] params

ishii1648:
The following codes always output 'msg="collect[] params" params="unsupported value type"'.
```
level.Debug(logger).Log("msg", "collect[] params", "params", params)
```

This "params" variable is slice, so should be split to string.

This is example log output by modified code.
```
msg="collect[] params" params=info_schema.innodb_cmp+nfo_schema.query_response_time
```

### DIFF
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"path"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/go-kit/kit/log"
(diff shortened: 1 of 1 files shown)


mysqld_exporter.go
```diff
@@ -195,7 +196,7 @@ func newHandler(metrics collector.Metrics, scrapers []collector.Scraper, logger
 				r = r.WithContext(ctx)
 			}
 		}
-		level.Debug(logger).Log("msg", "collect[] params", "params", params)
+		level.Debug(logger).Log("msg", "collect[] params", "params", strings.Join(params, "+"))
```
SuperQ, line 199: I think a comma would be better.

```suggestion
		level.Debug(logger).Log("msg", "collect[] params", "params", strings.Join(params, ","))
```

SuperQ:
Thanks, minor change request.

mysqld_exporter.go
ishii1648: Thanks. I got it!

SuperQ (approved):
LGTM

PR #506, opened 2020-10-21
Added cluster and namespace variables to the dashboard. 

bergquist:
I'm not really happy with it. I want to split it up into two different dashboards. One to get a global/cluster overview and for digging into certain servers. But its better!


bergquist:
ping @tomwilkie

bergquist:
Thinking a bit more about this, I dont think we should introduce cluster to this dashboard. Ill build two instead and people can choose which of them suits their needs. This will only be confusing folk users with few databases.

PR #512, opened 2020-11-02
mixin: fix redundant alert

oiooj:
There are 2 `MySQLReplicationLag` alerts.  I think it's just a typo.


SuperQ (approved):
Good catch, Thanks!

PR #515, opened 2020-11-11
Feat add ps memory collector

dafydd-t:
This PR adds a collector for `performance_schema.memory_summary_global_by_event_name`, creating three metrics as described in #75 .

- The collector is disabled by default
- Include a flag to remove the `memory/` prefix from each event label for this metric.
- Tested locally against MySQL 5.6, 5.7 and 8.
- 5.6 doesn't have this table, so upon scrape, the exporter logs: 
```
level=error ts=2020-11-11T17:06:26.873Z caller=exporter.go:174 msg="Error from scraper" scraper=perf_schema.memory_events err="Error 1146: Table 'performance_schema.memory_summary_global_by_event_name' doesn't exist"
```
- Running against mysql:8 and mysql:5.7 with performance_schema=OFF, the exporter logs no errors and does not export any memory_events metrics, but does expose `mysql_exporter_collector_duration_seconds ` for the collector.

- Add new collector flags to the README. Also included missing flag for `file_instances.remove_prefix`.

Issue #75, opened 2016-01-29
Implement Memory information capture from performance_schema on MySQL 5.7


PeterZaitsev:
Note this table only exists on MySQL 5.7 so graceful handling of its absence is recommended 

mysql> select \* from memory_summary_global_by_event_name where count_alloc>0 order by current_number_of_bytes_used desc limit 2 \G
**************************\* 1. row ***************************
                  EVENT_NAME: memory/innodb/buf_buf_pool
                 COUNT_ALLOC: 192
                  COUNT_FREE: 0
   SUM_NUMBER_OF_BYTES_ALLOC: 26398949376
    SUM_NUMBER_OF_BYTES_FREE: 0
              LOW_COUNT_USED: 0
          CURRENT_COUNT_USED: 192
             HIGH_COUNT_USED: 192
    LOW_NUMBER_OF_BYTES_USED: 0
CURRENT_NUMBER_OF_BYTES_USED: 26398949376
   HIGH_NUMBER_OF_BYTES_USED: 26398949376
**************************\* 2. row ***************************
                  EVENT_NAME: memory/innodb/hash0hash
                 COUNT_ALLOC: 82
                  COUNT_FREE: 6
   SUM_NUMBER_OF_BYTES_ALLOC: 1976975584
    SUM_NUMBER_OF_BYTES_FREE: 1223839920
              LOW_COUNT_USED: 0
          CURRENT_COUNT_USED: 76
             HIGH_COUNT_USED: 76
    LOW_NUMBER_OF_BYTES_USED: 0
CURRENT_NUMBER_OF_BYTES_USED: 753135664
   HIGH_NUMBER_OF_BYTES_USED: 1161079744
2 rows in set (0.01 sec)

There are a lot of columns here.  I think most interesting are ones related to the memory usage not number of allocation calls:

CURRENT_NUMBER_OF_BYTES_USED: 753135664
   HIGH_NUMBER_OF_BYTES_USED: 1161079744

Which correspond to current usage and highest usage ever.  Former is helpful for catching high memory usage by stored procedures or other areas where allocation is done for short term.

dafydd-t:
I'd like to take on this work. We'd like to export some of this data so we can better visualise our MySQL 5.7 memory usage.

### Proposed work

Handle case where this table doesn't exist, so the exporter is still compatible with earlier MySQL versions.

Use the following SQL query, with some explanations below:
```sql
[...]

PR #517, opened 2020-11-19
bugfix: allow perf_schema.memory summary current_bytes to be negative

dafydd-t:
Related to https://github.com/prometheus/mysqld_exporter/issues/75 and PR https://github.com/prometheus/mysqld_exporter/pull/515

It is possible to end up with a negative value for `CURRENT_NUMBER_OF_BYTES_USED` - this can happen if you enabled instrumentation for performance_schema memory at some point after startup time. 

Currently, if this is the case, the collector emits the following error when scraped, and stops scraping rows after the first row with a negative value for this column:

```
caller=exporter.go:174 msg="Error from scraper" scraper=perf_schema.memory_events err="sql: Scan error on column index 3, name \"CURRENT_NUMBER_OF_BYTES_USED\": converting driver.Value type []uint8 (\"-3214848\") to a uint64: invalid syntax"
```

This change makes the `currentBytes` variable signed, so it can accept negative values and should allow the collector to scrape all the rows of the query results.

the `memory_events_used_bytes ` is already a gauge, so it can be negative.

### DIFF
--- a/collector/perf_schema_memory_events.go
+++ b/collector/perf_schema_memory_events.go
@@ -90,7 +90,7 @@ func (ScrapePerfMemoryEvents) Scrape(ctx context.Context, db *sql.DB, ch chan<- 
 		eventName    string
 		bytesAlloc   uint64
 		bytesFree    uint64
-		currentBytes uint64
+		currentBytes int64
 	)
 
 	for perfSchemaMemoryEventsRows.Next() {
--- a/collector/perf_schema_memory_events_test.go
+++ b/collector/perf_schema_memory_events_test.go
@@ -47,6 +47,7 @@ func TestScrapePerfMemoryEvents(t *testing.T) {
 
 	rows := sqlmock.NewRows(columns).
 		AddRow("memory/innodb/event1", "1001", "500", "501").
+		AddRow("memory/performance_schema/event1", "6000", "7", "-83904").
 		AddRow("memory/innodb/event2", "2002", "1000", "1002").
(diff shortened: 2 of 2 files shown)


dafydd-t:
@SuperQ I'm pinging you as you were really helpful with the initial PR for this feature

SuperQ:
Yup, it looks like the columns is a `bigint(20)`.

```
CREATE TABLE `memory_summary_global_by_event_name` (
  `EVENT_NAME` varchar(128) NOT NULL,
  `COUNT_ALLOC` bigint(20) unsigned NOT NULL,
  `COUNT_FREE` bigint(20) unsigned NOT NULL,
  `SUM_NUMBER_OF_BYTES_ALLOC` bigint(20) unsigned NOT NULL,
  `SUM_NUMBER_OF_BYTES_FREE` bigint(20) unsigned NOT NULL,
  `LOW_COUNT_USED` bigint(20) NOT NULL,
  `CURRENT_COUNT_USED` bigint(20) NOT NULL,
  `HIGH_COUNT_USED` bigint(20) NOT NULL,
  `LOW_NUMBER_OF_BYTES_USED` bigint(20) NOT NULL,
  `CURRENT_NUMBER_OF_BYTES_USED` bigint(20) NOT NULL,
  `HIGH_NUMBER_OF_BYTES_USED` bigint(20) NOT NULL
) ENGINE=PERFORMANCE_SCHEMA DEFAULT CHARSET=utf8
```

SuperQ (approved):
LGTM

PR #520, opened 2020-12-29
Add unix socket support for metric exposition

ArcticB:
Hello again @SuperQ,

After adding UNIX socket support on the Mtail project, I offer you the same implementation for the Mysqld exporter. I have been playing with it for a few weeks now and I really would like to include it in my monitoring system. For the same reasons as mentioned in [this PR](https://github.com/google/mtail/pull/327), I think this feature can be useful to many people also on this tool.

Thanks

### DIFF
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -19,6 +19,7 @@ import (
 	"crypto/x509"
 	"fmt"
 	"io/ioutil"
+	"net"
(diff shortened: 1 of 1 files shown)


SuperQ:
Sorry, this has been discussed many times in the past. We do not intend to support unix sockets in any official exporters.

PR #526, opened 2021-01-28
Extend tests to validate metric name and help field

guidoiaquinti:
### Description
Extend `collector/collector_test.go` by adding validations to the `fqName`  and `help` fields.

### Why
While working on a PR to extend `collector/global_variables.go` I've introduced a breaking change not spotted by the current tests.

Note: tests are going to be more verbose but I consider this 👍 as a change in a metric name should be considered an incompatible API change. 

### Tests
Tests are passing.


guidoiaquinti:
@SuperQ please let me know if you are interested in this and I'll rebase. Otherwise I'll simply close it.

Issue #389, opened 2019-05-22
Feature request: expose metrics from mysql.innodb_index_stats


guidoiaquinti:
It could be nice to have `mysqld_exporter` exposing [InnoDB index stats metrics](https://dev.mysql.com/doc/refman/5.6/en/innodb-persistent-stats.html#innodb-persistent-stats-tables).

Example:
```
mysql> SELECT * FROM innodb_table_stats \G
*************************** 1. row ***************************
           database_name: sakila
              table_name: actor
             last_update: 2014-05-28 16:16:44
                  n_rows: 200
    clustered_index_size: 1
sum_of_other_index_sizes: 1
```
```
mysql> SELECT * FROM innodb_index_stats \G
*************************** 1. row ***************************
   database_name: sakila
      table_name: actor
      index_name: PRIMARY
     last_update: 2014-05-28 16:16:44
       stat_name: n_diff_pfx01
      stat_value: 200
     sample_size: 1
     ...
```

matlegit:
Hi @guidoiaquinti , do you know if this was ever implemented? We're interested in table index size as well

guidoiaquinti:
From what I can see in the codebase the answer is: no

pgporada:
@matlegit, I've put up a PR which should provide these stats. Here's hoping it gets reviewed in a timely manner!

PR #529, opened 2021-02-05
Collect innodb table and index stats

pgporada:
Hi,

We found this information useful in troubleshooting some database performance issues at my job. I hope this information can be useful for others who run multi-TB sized databases with tons of indices.

Information regarding these metrics can be found at 

- https://mariadb.com/kb/en/mysqlinnodb_index_stats/
- https://mariadb.com/kb/en/mysqlinnodb_table_stats/

The metric descriptions where mostly copy pasted from https://dev.mysql.com/doc/refman/5.6/en/innodb-persistent-stats.html

Using a test mariadb container, the new metrics will appear as
```
mysql_mysql_innodb_index_stats{database_name="mysql",index_name="PRIMARY",stat_name="n_diff_pfx01",table_name="gtid_slave_pos"} 0
mysql_mysql_innodb_index_stats{database_name="mysql",index_name="PRIMARY",stat_name="n_diff_pfx02",table_name="gtid_slave_pos"} 0
mysql_mysql_innodb_index_stats{database_name="mysql",index_name="PRIMARY",stat_name="n_leaf_pages",table_name="gtid_slave_pos"} 1
mysql_mysql_innodb_index_stats{database_name="mysql",index_name="PRIMARY",stat_name="size",table_name="gtid_slave_pos"} 1
mysql_mysql_innodb_table_stats_clustered_index_size{database_name="mysql",table_name="gtid_slave_pos"} 1
mysql_mysql_innodb_table_stats_n_rows{database_name="mysql",table_name="gtid_slave_pos"} 0
mysql_mysql_innodb_table_stats_sum_of_other_index_sizes{database_name="mysql",table_name="gtid_slave_pos"} 0
```

### DIFF
--- a/collector/mysql_innodb_index_stats.go
+++ b/collector/mysql_innodb_index_stats.go
@@ -1,0 +1,106 @@
+// Copyright 2018 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Scrape `mysql.innodb_index_stats`.
+
+package collector
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+
+	"github.com/go-kit/kit/log"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const mysqlIndexStatsQuery = `
+		  SELECT
+		    database_name,
+		    table_name,
+		    index_name,
+            stat_name,
+            stat_value
+		  FROM mysql.innodb_index_stats
+		`
+
+var (
+	indexStatLabelNames = []string{"database_name", "table_name", "index_name", "stat_name"}
+)
+
+// Metric descriptors.
+var (
+	indexStatsValueDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, mysql, "innodb_index_stats_stat_value"),
+		"Stores data related to particular InnoDB Persistent Statistics.",
+		indexStatLabelNames, nil)
+)
+
+type ScrapeMysqlIndexStat struct{}
+
+// Name of the Scraper. Should be unique.
+func (ScrapeMysqlIndexStat) Name() string {
+	return "mysql.innodb_index_stats"
+}
+
+// Help describes the role of the Scraper.
+func (ScrapeMysqlIndexStat) Help() string {
+	return "Collect data from mysql.innodb_index_stats"
+}
+
+// Version of MySQL from which scraper is available.
+func (ScrapeMysqlIndexStat) Version() float64 {
+	return 5.1
+}
+
+// Scrape collects data from database connection and sends it over channel as prometheus metric.
+func (ScrapeMysqlIndexStat) Scrape(ctx context.Context, db *sql.DB, ch chan<- prometheus.Metric, logger log.Logger) error {
+	var (
+		indexStatRows *sql.Rows
+		err           error
+	)
+	indexStatsQuery := fmt.Sprint(mysqlIndexStatsQuery)
+	indexStatRows, err = db.QueryContext(ctx, indexStatsQuery)
+	if err != nil {
+		return err
+	}
+	defer indexStatRows.Close()
+
+	var (
+		database_name string
+		table_name    string
+		index_name    string
+		stat_name     string
+		stat_value    uint32
+	)
+
+	for indexStatRows.Next() {
+		err = indexStatRows.Scan(
+			&database_name,
+			&table_name,
+			&index_name,
+			&stat_name,
+			&stat_value,
+		)
+
+		if err != nil {
+			return err
+		}
+
+		ch <- prometheus.MustNewConstMetric(indexStatsValueDesc, prometheus.GaugeValue, float64(stat_value), database_name, table_name, index_name, stat_name)
+	}
+
+	return nil
+}
+
+var _ Scraper = ScrapeMysqlIndexStat{}
--- a/collector/mysql_innodb_index_stats.go
+++ b/collector/mysql_innodb_index_stats.go
@@ -42,7 +42,7 @@ var (
 var (
 	indexStatsValueDesc = prometheus.NewDesc(
 		prometheus.BuildFQName(namespace, mysql, "innodb_index_stats_stat_value"),
-		"Stores data related to particular InnoDB Persistent Statistics.",
+		"Size of the InnoDB index in bytes.",
 		indexStatLabelNames, nil)
 )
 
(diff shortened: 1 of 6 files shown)


pgporada:
Fixes #389

guidoiaquinti:
Nice work! 👏  

In a future iteration, what do you think about adding CLI flags to be able to specify which schema and table to watch? In a big deployment exporting metrics for all database/tables can be expensive.

pgporada:
I think that would be wise to add to this PR. When I first started on this work I new 0 golang. I now know 1 golang and can probably suss it out.

mdkent:
@pgporada I'm curious how you're updating these stats, are you calling ANALYZE TABLE on a regular cadence or do you just have enough row churn that they get regular updates?

Per https://mariadb.com/kb/en/innodb-system-variables/#innodb_stats_auto_recalc you need ~10% of the rows to get touched for a stats refresh.

pgporada:
@mdkent 
The [Boulder CA](https://github.com/letsencrypt/boulder/) software has very high row churn when updating certificate status information for 200+ million active TLS certificates.

collector/mysql_innodb_index_stats.go
```diff
@@ -0,0 +1,106 @@
+// Copyright 2018 The Prometheus Authors
```
SuperQ, line 1: This should be 2021.

collector/mysql_innodb_index_stats.go
```diff
@@ -0,0 +1,106 @@
...
+		`
+
+var (
+	indexStatLabelNames = []string{"database_name", "table_name", "index_name", "stat_name"}
+)
+
+// Metric descriptors.
+var (
+	indexStatsValueDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, mysql, "innodb_index_stats_stat_value"),
```
SuperQ, line 44: This is going to make metrics named like `mysql_mysql_...`. I think we want `prometheus.BuildFQName(namespace, innodb, "index_stats_stat_value")`

SuperQ (commented):
The problem with these generic stats dumps is that they're not useful without some extended documentation/help.

In Prometheus, we tend to avoid generic data dumps. Rather, we create opinionated names/mappings so that the end users doesn't have as much work to do in order to understand the meaning of the metrics.

pgporada:
> The problem with these generic stats dumps is that they're not useful without some extended documentation/help.
@SuperQ Where should I do that, in the README.md or in the specific .go files?

The mysql-server:5.7 broken test seems unrelated to this change fwiw.

collector/mysql_innodb_index_stats.go
```diff
@@ -0,0 +1,106 @@
...
+	indexStatLabelNames = []string{"database_name", "table_name", "index_name", "stat_name"}
+)
+
+// Metric descriptors.
+var (
+	indexStatsValueDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, innodb, "innodb_index_stats_stat_value"),
+		"Size of the InnoDB index in bytes.",
+		indexStatLabelNames, nil)
+)
```
ArthurSens, line 47: I believe we need to adjust the metric here to comply with [Prometheus conventions](https://prometheus.io/docs/practices/naming/), primarily by ensuring each metric has a single unit of measure.

Bytes isn't really what is returned here, since all stats are mixed up. A query like `sum(mysql_engine_innodb_innodb_index_stats_stat_value)` returns something that makes absolutely no sense.

If we break this down into separate metrics, like, for example:

```
mysql_innodb_index_size_bytes{
  schema="shop", table="orders", index="PRIMARY"
} 1638400

mysql_innodb_index_leaf_size_bytes{
  schema="shop", table="orders", index="PRIMARY"
} 1310720

mysql_innodb_index_distinct_values{
  schema="shop", table="orders", index="customer_created_at",
  prefix_columns="customer_id"
} 5000

mysql_innodb_index_distinct_values{
  schema="shop", table="orders", index="customer_created_at",
  prefix_columns="customer_id,created_at"
} 12000
```

A `sum` would sum index bytes with index bytes, leaf bytes with leaf bytes and we guarantee that queries have meaningful results.

collector/mysql_innodb_index_stats.go
```diff
@@ -0,0 +1,106 @@
...
+		return err
+	}
+	defer indexStatRows.Close()
+
+	var (
+		database_name string
+		table_name    string
+		index_name    string
+		stat_name     string
+		stat_value    uint32
```
ArthurSens, line 84: should we use unit64 to ensure this still works for big numbers? [Docs](https://mariadb.com/docs/server/reference/system-tables/the-mysql-database-tables/mysql-innodb_index_stats#:~:text=stat_value,Estimated%20statistic%20value.) shows that this row is a BIGINT

collector/mysql_innodb_index_stats.go
```diff
@@ -0,0 +1,106 @@
...
+// Scrape `mysql.innodb_index_stats`.
+
+package collector
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+
+	"github.com/go-kit/kit/log"
```
ArthurSens, line 23: we use a different logging library nowadays

collector/mysql_innodb_index_stats.go
```diff
@@ -0,0 +1,106 @@
...
+		`
+
+var (
+	indexStatLabelNames = []string{"database_name", "table_name", "index_name", "stat_name"}
+)
+
+// Metric descriptors.
+var (
+	indexStatsValueDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, innodb, "innodb_index_stats_stat_value"),
```
ArthurSens, line 44:
```suggestion
		prometheus.BuildFQName(namespace, innodb, "index_stats_stat_value"),
```

seems awkward to have repeated `innodb` in the query name

collector/mysql_innodb_table_stats.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+const mysqlTableStatsQuery = `
+		  SELECT
+		    database_name,
+		    table_name,
+		    n_rows,
+            clustered_index_size,
+            sum_of_other_index_sizes
+		  FROM mysql.innodb_table_stats
```
ArthurSens, line 34: This query could be massive, depending on the database size. Maybe we can add a configurable filter so we can collect stats from a smaller set of tables?

collector/mysql_innodb_index_stats.go
```diff
@@ -0,0 +1,106 @@
...
+)
+
+const mysqlIndexStatsQuery = `
+		  SELECT
+		    database_name,
+		    table_name,
+		    index_name,
+            stat_name,
+            stat_value
+		  FROM mysql.innodb_index_stats
```
ArthurSens, line 34: This query could be massive, depending on the database size. Maybe we can add a configurable filter so we can collect stats from a smaller set of indexes?

ArthurSens: Also, could we include the [`last_update` timestamp](https://dev.mysql.com/doc/refman/8.4/en/innodb-persistent-stats.html#:~:text=last_update,updated%20this%20row) as a separate metric?

Looking at the docs, stats updates happen very sporadically, so understanding how old the stats information is

ArthurSens (commented):
Hey 👋, sorry for the super-delayed review. The codebase has changed a lot in the last few years, so a rebase will definitely be needed here 😓 

It would be awesome if the new collector could be tested against the integration tests we have for several different MySQL flavors

PR #537, opened 2021-03-23
support for scraping multiple mysql host

caiwc:


SuperQ:
We will be supporting multi-target using the [multi-target exporter pattern](https://prometheus.io/docs/guides/multi-target-exporter/).

ning1875:
I have an improved version here, transformed into a probe, which supports http incoming parameters like blackbox_exporter .  https://github.com/ning1875/mysqld_exporter/releases/tag/v0.12.1

PR #539, opened 2021-03-25
Support authenticating using mTLS client cert and no password

tirsen:


SuperQ:
Nice. I still want to completely re-write the config handler here. But I think this is a step in the right direction.

tirsen:
Ping? If it looks good are you okay with merging?

SuperQ (approved):
LGTM

Yes, I think this is fine for now, re-writing the config handling is still a TODO item.

PR #542, opened 2021-04-02
$host changed by $instance

herveleclerc:
$host changed by $instance in 'MySQL Handlers' and MySQL Transaction Handlers

$host variables does not exist.

Cheers


SuperQ:
This was fixed in #580

PR #551, opened 2021-05-10
Update README.md

wonderg:
User changes should be followed by FLUSH PRIVILEGES;


wonderg:
i don't understand how to sigh off this correctly. Nothing did work for me.

SuperQ:
This is unnecessary when using `USER` and `GRANT`. See the [MySQL documentation](https://dev.mysql.com/doc/refman/8.0/en/privilege-changes.html).

PR #564, opened 2021-07-28
fix: log message without any key

greut:
> {"(go=go1.16.4, user=root@e2043849cb1f, date=20210531-07:30:16)":"(MISSING)","caller":"mysqld_exporter.go:278","level":"info","msg":"Build context","ts":"2021-07-27T07:51:05.601Z"}

This is the log we see and it doesn't please Kibana.


SuperQ (approved):
LGTM, Thanks!

greut:
@SuperQ anything else might be needed?

SuperQ:
I need to fix the mixin, but that's a separate issue.

PR #566, opened 2021-08-06
transform into a prober just like the blackbox_exporter

ning1875:


SuperQ:
Taking the DSN from the URL is insecure and not something we can use here.

Issue #422, opened 2019-08-13
Exporter logging errors when running SHOW SLAVE HOSTS on mysql 5.5


matejzero:
### Host operating system: output of `uname -a`
CentOS 7 latest (3.10.0-862.11.6.el7.x86_64)

### mysqld_exporter version: output of `mysqld_exporter --version`
`mysqld_exporter, version 0.12.1 (branch: HEAD, revision: 48667bf7c3b438b5e93b259f3d17b70a7c9aff96)`

### MySQL server version
Percona MySQL 5.5.61-rel38

### mysqld_exporter command line flags
`/usr/bin/mysqld_exporter --web.listen-address=:9113 --config.my-cnf=/var/lib/mysqld_exporter/my.cnf --collect.binlog_size --collect.engine_innodb_status --collect.info_schema.processlist --collect.info_schema.query_response_time --collect.info_schema.clientstats --collect.info_schema.userstats --collect.slave_status --collect.slave_hosts --no-collect.info_schema.tables`

### What did you do that produced an error?
Ran exporter on a mysql master running version 5.5.

### What did you expect to see?
No errors in the logs.

### What did you see instead?
Errors in the logs regarding `SHOW SLAVE HOSTS`.
```
Aug 13 21:06:35 server mysqld_exporter[5076]: time="2019-08-13T21:06:35+02:00" level=error msg="Error scraping for collect.slave_hosts: sql: expected 4 destination arguments in Scan, not 5" source="exporter.go:171"
```

If I ran `SHOW SLAVE HOSTS` on my server, this is the output:
```
mysql> SHOW SLAVE HOSTS;
+-----------+---------------+------+-----------+
| Server_id | Host          | Port | Master_id |
+-----------+---------------+------+-----------+
|        13 | sql2.domain.tld | 3306 |        11 |
|        12 | sql3.domain.tld | 3306 |        11 |
+-----------+---------------+------+-----------+
```

According to MySQL documentation, 5.6 does have `Slave_UUID` column, but 5.5 doesn't.

I think this is the code where it fails: https://github.com/prometheus/mysqld_exporter/blob/master/collector/slave_hosts.go#L87

I also see that there are no tests for 4 column output.

betanummeric:
[...]

PR #577, opened 2021-09-13
fix mysql_slave_hosts_info for mysql 5.5 and mariadb 10.5, add unit test

betanummeric:
Fixes https://github.com/prometheus/mysqld_exporter/issues/422

The `show slave hosts` command on mysql 5.5 and mariadb 10.5 returns neither a `Slave_UUID` nor a `Rpl_recovery_rank` column. Set the `slave_uuid` metric label to `""` in this case.

The unit test is copied from `TestScrapeSlaveHostsNewFormat`, with the mocked and expected rows amended. I also tested this with real MySQL 5.5.62, MySQL 5.7.33 and MariaDB 10.5.10 servers.

@SuperQ Can you please review this pull request? If I missed something, please let me know.

### DIFF
--- a/collector/slave_hosts.go
+++ b/collector/slave_hosts.go
@@ -80,25 +80,42 @@ func (ScrapeSlaveHosts) Scrape(ctx context.Context, db *sql.DB, ch chan<- promet
 	var masterId string
 	var slaveUuid string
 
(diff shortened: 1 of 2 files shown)


SuperQ (approved):
LGTM, Thanks.

PR #580, opened 2021-10-04
Update mixin to use `$__rate_interval`

rgeyer:
This PR changes to using `$__rate_interval` for promql queries containing a rate function call.

Since `$__rate_interval` should "always work" within reason, the queries which had a conditional fallback to a (potentially) larger timerange with and `irate` were also simplified to simply use `rate` with `$__rate_interval`.

Let me know if this can be merged, or needs more work @SuperQ (mentioning you per the contributing guidelines, thanks!).


SuperQ (approved):
LGTM, Thanks!

tomwilkie (approved):
LGTM too!

Did a quick grep and all I could find was the alerts and rules still doing a `rate[5m]`, which is fine.

Thanks @rgeyer (and @SuperQ for the heads up)

PR #584, opened 2021-10-20
Data source template should be labelled 'Data Source'

tomwilkie:
We're building a linter for Grafana dashboards and this came up; its super minor nit sorry.

Signed-off-by: Tom Wilkie <tom@grafana.com>


SuperQ (approved):
LGTM

PR #594, opened 2021-11-10
Add Custom Queries

spectro30:
Signed-off-by: Alif Biswas <alif@appscode.com>


SuperQ:
We do not intend to support custom queries in this exporter, as produces too many anti-pattern uses.

PR #442, opened 2019-11-04
add group command in collect.info_schema.processlist

hctech:
# HELP mysql_info_schema_processes_by_host The number of processes by host.
# TYPE mysql_info_schema_processes_by_host gauge
mysql_info_schema_processes_by_host{client_host="",command="Connect"} 6
mysql_info_schema_processes_by_host{client_host="10.11.0.9",command="Query"} 1
mysql_info_schema_processes_by_host{client_host="172.26.0.181",command="Sleep"} 5

# HELP mysql_info_schema_processes_by_user The number of processes by user.
# TYPE mysql_info_schema_processes_by_user gauge
mysql_info_schema_processes_by_user{command="Connect",mysql_user="system user"} 6
mysql_info_schema_processes_by_user{command="Query",mysql_user="root"} 1

PR #603, opened 2021-12-02
Rewrite processlist collector

roman-vynar:
Hello,

Rewritten processlist collector to include both Command and State columns and stop doing state remapping.

Couple of reasons for this:
* The existing collector has been last updated 3 years ago when we had no MySQL 8.0. Since that time we have got more process states and changes to the existing ones https://dev.mysql.com/doc/refman/8.0/en/general-thread-states.html
The old re-mapping is not accurate, somewhat outdated and requires constant maintaining. With this PR, we won't bother about thread state mapping and it will flexible for anyone to do what they want to do on the expression side.
* The existing collector does incorrect grouping by Command/State. For example, anything Command=Query is considered as State=executing which may include:
```
{command="query", state="checking_permissions"}
{command="query", state="executing"}
{command="query", state="reading_event_from_the_relay_log"}
{command="query", state="sending_to_client"}
{command="query", state="slave_has_read_all_relay_log_waiting_for_more_updates"}
{command="query", state="waiting_for_global_read_lock"}
{command="query", state="waiting_for_handler_commit"}
```
As you can see it is not always an execution phase of regular queries but a whole bunch of stuff that we may want to exclude for alerting and which is the normal thing, for say Group Replication waiting_for_handler_commit (it is a phase of group commit where it is waiting and not consuming CPU) as opposite to the user queries (consuming CPU).
It becomes very handy when we want to sum all queries and exclude some particular states, or just see on the graph how those state are distributed to decide what to do.
* There are other commands like "Connect", "Daemon", "Binlog dump" which we may want to see w/o being grouped/remapped into some states.

***Breaking changes:***
* Added "command" label to processlist "threads" and "seconds" metrics. No much increase in number of time-series.
* Stop doing remapping of the existing states. Just simple text normalization for command/state values.
* Metrics has been renamed to be easily found by name of the collector and for consistency that they belong to the same one:
```
mysql_info_schema_threads > mysql_info_schema_processlist_threads
mysql_info_schema_threads_seconds > mysql_info_schema_processlist_seconds
mysql_info_schema_processes_by_user > mysql_info_schema_processlist_processes_by_user
mysql_info_schema_processes_by_host > mysql_info_schema_processlist_processes_by_host
```
Also added a test.

There is a similar PR https://github.com/prometheus/mysqld_exporter/pull/442 but it just adds "command" and retains the state remapping. 
Another PR https://github.com/prometheus/mysqld_exporter/pull/381 complaining about remapping.

Sample output:
```# HELP mysql_info_schema_processlist_processes_by_host The number of processes by host.
# TYPE mysql_info_schema_processlist_processes_by_host gauge
mysql_info_schema_processlist_processes_by_host{client_host="10.0.7.154"} 8
mysql_info_schema_processlist_processes_by_host{client_host="10.0.7.224"} 1
mysql_info_schema_processlist_processes_by_host{client_host="10.0.7.234"} 24
mysql_info_schema_processlist_processes_by_host{client_host="blank"} 2
mysql_info_schema_processlist_processes_by_host{client_host="localhost"} 1
# HELP mysql_info_schema_processlist_processes_by_user The number of processes by user.
# TYPE mysql_info_schema_processlist_processes_by_user gauge
mysql_info_schema_processlist_processes_by_user{mysql_user="manager"} 20
mysql_info_schema_processlist_processes_by_user{mysql_user="store"} 1
mysql_info_schema_processlist_processes_by_user{mysql_user="event_scheduler"} 1
mysql_info_schema_processlist_processes_by_user{mysql_user="feedback"} 1
mysql_info_schema_processlist_processes_by_user{mysql_user="message"} 4
mysql_info_schema_processlist_processes_by_user{mysql_user="root"} 1
mysql_info_schema_processlist_processes_by_user{mysql_user="system user"} 2
# HELP mysql_info_schema_processlist_threads The number of threads split by current state.
# TYPE mysql_info_schema_processlist_threads gauge
mysql_info_schema_processlist_threads{command="connect",state="waiting_for_handler_commit"} 1
mysql_info_schema_processlist_threads{command="daemon",state="waiting_on_empty_queue"} 1
mysql_info_schema_processlist_threads{command="query",state="slave_has_read_all_relay_log_waiting_for_more_updates"} 1
mysql_info_schema_processlist_threads{command="sleep",state="blank"} 36
# HELP mysql_info_schema_processlist_seconds The number of seconds threads have used split by current state.
# TYPE mysql_info_schema_processlist_seconds gauge
mysql_info_schema_processlist_seconds{command="connect",state="waiting_for_handler_commit"} 7.339619e+06
mysql_info_schema_processlist_seconds{command="daemon",state="waiting_on_empty_queue"} 7.339619e+06
mysql_info_schema_processlist_seconds{command="query",state="slave_has_read_all_relay_log_waiting_for_more_updates"} 7.339619e+06
mysql_info_schema_processlist_seconds{command="sleep",state="blank"} 1030
```

### DIFF
--- a/collector/info_schema_processlist.go
+++ b/collector/info_schema_processlist.go
@@ -61,19 +61,19 @@ var (
 // Metric descriptors.
 var (
 	processlistCountDesc = prometheus.NewDesc(
-		prometheus.BuildFQName(namespace, informationSchema, "processlist_total"),
+		prometheus.BuildFQName(namespace, informationSchema, "processlist_threads"),
 		"The number of threads split by current state.",
 		[]string{"command", "state"}, nil)
 	processlistTimeDesc = prometheus.NewDesc(
-		prometheus.BuildFQName(namespace, informationSchema, "processlist_seconds_total"),
+		prometheus.BuildFQName(namespace, informationSchema, "processlist_seconds"),
 		"The number of seconds threads have used split by current state.",
 		[]string{"command", "state"}, nil)
 	processesByUserDesc = prometheus.NewDesc(
-		prometheus.BuildFQName(namespace, informationSchema, "processlist_by_user"),
+		prometheus.BuildFQName(namespace, informationSchema, "processlist_processes_by_user"),
 		"The number of processes by user.",
 		[]string{"mysql_user"}, nil)
 	processesByHostDesc = prometheus.NewDesc(
-		prometheus.BuildFQName(namespace, informationSchema, "processlist_by_host"),
+		prometheus.BuildFQName(namespace, informationSchema, "processlist_processes_by_host"),
 		"The number of processes by host.",
 		[]string{"client_host"}, nil)
 )
--- a/collector/info_schema_processlist.go
+++ b/collector/info_schema_processlist.go
@@ -19,6 +19,8 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"reflect"
+	"sort"
 	"strings"
 
 	"github.com/go-kit/log"
@@ -30,16 +32,15 @@ const infoSchemaProcesslistQuery = `
 		  SELECT
 		    user,
 		    SUBSTRING_INDEX(host, ':', 1) AS host,
-		    COALESCE(command,'') AS command,
-		    COALESCE(state,'') AS state,
-		    count(*) AS processes,
-		    sum(time) AS seconds
+		    COALESCE(command, '') AS command,
+		    COALESCE(state, '') AS state,
+		    COUNT(*) AS processes,
+		    SUM(time) AS seconds
 		  FROM information_schema.processlist
 		  WHERE ID != connection_id()
 		    AND TIME >= %d
-		  GROUP BY user,SUBSTRING_INDEX(host, ':', 1),command,state
-		  ORDER BY null
-		`
+		  GROUP BY user, SUBSTRING_INDEX(host, ':', 1), command, state
+	`
 
 // Tunable flags.
 var (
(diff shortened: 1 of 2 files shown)


SuperQ:
Thanks!

One question, this drops the `ORDER BY null`, I seem to remember this was a small performance optimization. Possibly for a very old version of MySQL.

roman-vynar:
Yea, looks like it's from some very old MySQL. Normally, it shouldn't do anything useful but being present :)

collector/info_schema_processlist.go
```diff
@@ -189,83 +109,102 @@ func (ScrapeProcesslist) Scrape(ctx context.Context, db *sql.DB, ch chan<- prome
...
-		realState := deriveThreadState(command, state)
-		stateCounts[realState] += processes
-		stateTime[realState] += time
-		hostCount[host] = hostCount[host] + processes
-		userCount[user] = userCount[user] + processes
-	}
+		command = sanitizeState(command)
+		state = sanitizeState(state)
+		if host == "" {
+			host = "blank"
```
SuperQ, line 133: I think we would prefer to keep an empty string here, or `"unknown"` rather than `"blank"`.

collector/info_schema_processlist_test.go
```diff
@@ -0,0 +1,77 @@
...
+	"testing"
+
+	"github.com/DATA-DOG/go-sqlmock"
+	"github.com/go-kit/log"
+	"github.com/prometheus/client_golang/prometheus"
+	dto "github.com/prometheus/client_model/go"
+	"github.com/smartystreets/goconvey/convey"
+)
+
+func TestScrapeProcesslist(t *testing.T) {
```
SuperQ, line 28: In order for this to work, you need to parse the command line flags.

Something like this:
```go
  _, err := kingpin.CommandLine.Parse([]string{"--collect.info_schema.processlist", ""})
  if err != nil {
    t.Fatal(err)
  }
```

collector/info_schema_processlist_test.go
```diff
@@ -0,0 +1,77 @@
+// Copyright 2018 The Prometheus Authors
```
SuperQ, line 1: Nit
```suggestion
// Copyright 2021 The Prometheus Authors
```

SuperQ (changes requested):

collector/info_schema_processlist_test.go
roman-vynar: Cool, I was trying to find how to pass this but didn't have a luck. Thanks!

SuperQ:
Looks like there are some unrelated mixin errors.

collector/info_schema_processlist.go
```diff
@@ -60,104 +61,23 @@ var (
 // Metric descriptors.
 var (
 	processlistCountDesc = prometheus.NewDesc(
-		prometheus.BuildFQName(namespace, informationSchema, "threads"),
-		"The number of threads (connections) split by current state.",
-		[]string{"state"}, nil)
+		prometheus.BuildFQName(namespace, informationSchema, "processlist_total"),
```
SuperQ, line 64: These are gauges, so they should not end in `_total`. The original metric name here is correct. Maybe `processlist_threads`?

collector/info_schema_processlist.go
```diff
@@ -60,104 +61,23 @@ var (
...
+		"The number of threads split by current state.",
+		[]string{"command", "state"}, nil)
 	processlistTimeDesc = prometheus.NewDesc(
-		prometheus.BuildFQName(namespace, informationSchema, "threads_seconds"),
-		"The number of seconds threads (connections) have used split by current state.",
-		[]string{"state"}, nil)
+		prometheus.BuildFQName(namespace, informationSchema, "processlist_seconds_total"),
```
SuperQ, line 68: Same here, this is a gauge. Maybe `processlist_seconds`.

collector/info_schema_processlist.go
```diff
@@ -60,104 +61,23 @@ var (
...
+		"The number of seconds threads have used split by current state.",
+		[]string{"command", "state"}, nil)
 	processesByUserDesc = prometheus.NewDesc(
-		prometheus.BuildFQName(namespace, informationSchema, "processes_by_user"),
+		prometheus.BuildFQName(namespace, informationSchema, "processlist_by_user"),
```
SuperQ, line 72: `processlist_processes_by_user`?

SuperQ (commented):
There are some issues with the metric names.

roman-vynar:
You are right. I corrected the names and the description about breaking changes on the first post.

SuperQ (approved):
LGTM, Thanks!

PR #611, opened 2022-01-18
enableMgrAlert:Return different values according to different states …

yizhenlangyan:
…of performance_schema.replication_group_members column member_state state, and then alarm for MGR.


yizhenlangyan:
Performance_schema. replication_group_members MEMBER_STATE: returns 0 when not ONLINE. Configure rules in Prometheus for sending MGR non-online alarm.

SuperQ:

PR #615, opened 2022-02-15
set type of mysql_global_status_threads_* to gauge

aliorouji:
We are interested in these metrics:

- `mysql_global_status_threads_cached`
- `mysql_global_status_threads_connected`
- `mysql_global_status_threads_created`
- `mysql_global_status_threads_running`

which current version of `mysqld_exporter` does not assign a metric type to them. 

This PR will assign the correct metric type to them.


SuperQ:

collector/global_status_test.go
```diff
@@ -88,6 +92,10 @@ func TestScrapeGlobalStatus(t *testing.T) {
 		{labels: labelMap{"operation": "made_young"}, value: 15, metricType: dto.MetricType_COUNTER},
 		{labels: labelMap{"operation": "read"}, value: 8, metricType: dto.MetricType_COUNTER},
 		{labels: labelMap{"instrumentation": "users_lost"}, value: 9, metricType: dto.MetricType_COUNTER},
+		{labels: labelMap{"state": "cached"}, value: 7, metricType: dto.MetricType_GAUGE},
+		{labels: labelMap{"state": "connected"}, value: 18, metricType: dto.MetricType_GAUGE},
+		{labels: labelMap{"state": "created"}, value: 25, metricType: dto.MetricType_GAUGE},
```
ArthurSens, line 97: [Looking at docs](https://dev.mysql.com/doc/refman/8.4/en/connection-interfaces.html#:~:text=To%20monitor%20the%20number%20of%20threads%20in%20the%20cache%20and%20how%20many%20threads%20have%20been%20created%20because%20a%20thread%20could%20not%20be%20taken%20from%20the%20cache%2C%20check%20the%20Threads_cached%20and%20Threads_created%20status%20variables.), it seems like Threads_created is a number that only goes up, behaving like a Prometheus counter. This shouldn't be a gauge, I believe 🤔

ArthurSens (changes requested):
Sorry for the delayed review here. Seems like a good change, but we might need to update a few things first.

1. Although there's no conflicts, it mght be a good idea to do a rebase
2. Our mixins depend on the metrics being change, I think... could we double check?
3. Threads_created should be a counter

Issue #528, opened 2021-02-02
SPACE_TYPE Error while querying innodb_tablespaces


jzielke84:
### Host operating system: output of `uname -a`
Linux myhost01 4.9.0-14-amd64 #1 SMP Debian 4.9.246-2 (2020-12-17) x86_64 GNU/Linux

### mysqld_exporter version: output of `mysqld_exporter --version`
mysqld_exporter, version 0.12.1 (branch: HEAD, revision: 48667bf7c3b438b5e93b259f3d17b70a7c9aff96)
  build user:       root@0b3e56a7bc0a
  build date:       20190729-12:35:58
  go version:       go1.12.7

### MySQL server version
mysql  Ver 15.1 Distrib 10.1.48-MariaDB, for debian-linux-gnu (x86_64) using readline 5.2

### mysqld_exporter command line flags
            --collect.auto_increment.columns \
            --collect.binlog_size \
            --collect.engine_innodb_status \
            --collect.global_status \
            --collect.global_variables \
            --collect.info_schema.clientstats \
            --collect.info_schema.innodb_metrics \
            --collect.info_schema.innodb_tablespaces \
            --collect.info_schema.innodb_cmp \
            --collect.info_schema.innodb_cmpmem \
            --collect.info_schema.processlist \
            --collect.info_schema.tables \
            --collect.info_schema.tablestats \
            --collect.info_schema.schemastats \
            --collect.info_schema.userstats \
            --collect.perf_schema.eventsstatements \
            --collect.perf_schema.eventsstatementssum \
            --collect.perf_schema.eventswaits \
            --collect.perf_schema.file_events \
            --collect.perf_schema.file_instances \
            --collect.perf_schema.indexiowaits \
            --collect.perf_schema.tableiowaits \
            --collect.perf_schema.tablelocks

### What did you do that produced an error?
Starte the exporter

### What did you expect to see?
No errors

### What did you see instead?
mysql_exporter[6826]: time="2021-02-02T10:16:22+01:00" level=error msg="Error scraping for collect.info_schema.innodb_tablespaces: Error 1054: Unknown column 'SPACE_TYPE' in 'field list'" source="expor

Guzzi:
I observe the same error with 0.13.0.rc0 on mariadb 10.5.8
Diasabling info_schema.innodb_tablespaces makes the error line disappear.

PR #616, opened 2022-02-24
remove SPACE_TYPE from innodb_tablespaces (Closes: #528)

hggh:
The SPACE_TYPE column was introduced in MariaDB 10.2.1 and removed in MariaDB 10.5.0.


SuperQ:
What did this label provide in the previous versions? Are we sure that nobody depends on this in older versions?

Do we need to support both the old and new versions at the same time?

hggh:
The information inside SPACE_TYPE was General or Single. if you have the configuration option innodb_file_per_table the SPACE_TYPE was "Single". 

I'm not sure if anybody uses that information, since it was only present from 10.2.1 to 10.5; it was not so long in usage. 

But we can check the version and query that field if a MariaDB 10.2.1 to 10.5 is running?

ArthurSens:
Hi 👋 , looks like the problem was already solevd by #860, therefore I'm closing this one :)

Issue #617, opened 2022-02-28
Latest release v0.13.0 has CVE-2021-3538 (score 9.8) from dependency github.com/satori/go.uuid v1.2.0


srlobo:
Hi

mysqld_exporter showed up in a security scan because it contains github.com/satori/go.uuid@v1.2.0

https://nvd.nist.gov/vuln/detail/CVE-2021-3538

github.com/satori/go.uuid is used in `collector/slave_hosts.go`.

Would it be possible to make a patch release on docker hub containing an updated version of this module?

SuperQ:
It looks like the vulnerability is for generating UUIDs from that library. This collector only uses the library to parse UUID from MySQL. Therefore this collector / version is not vulnerable to [CVE-2021-3538](https://nvd.nist.gov/vuln/detail/CVE-2021-3538).

Please do not report results from vulnerability scanners without actively verifying that there is a vulnerability. These version-based scanners often produce false-positives.

srlobo:
Ok, but it's not a good idea to link to unmaintained code anyway (as told in the project discussion): https://github.com/satori/go.uuid/issues/73#issuecomment-905005880

They're suggesting change to https://github.com/gofrs/uuid

SuperQ:
PRs welcome.

PR #618, opened 2022-02-28
Change uuid lib to fix vulnerability

srlobo:
Change github.com/satori/go.uuid to github.com/gofrs/uuid. The former is not a
maintained library, and since it has a critical vulnerability CVE-2021-3538
there's an ongoing discussion proposing to change it for the later (https://github.com/satori/go.uuid/issues/73#issuecomment-905005880)

For reference: https://github.com/prometheus/mysqld_exporter/issues/617


SuperQ:

SuperQ:
The mixin failure is unrelated, will ignore it.

go.mod
```diff
@@ -4,11 +4,11 @@ require (
 	github.com/DATA-DOG/go-sqlmock v1.5.0
 	github.com/go-kit/log v0.2.0
 	github.com/go-sql-driver/mysql v1.6.0
+	github.com/gofrs/uuid v4.2.0+incompatible
```
SuperQ, line 7: So it seems like this fork dropped support Go modules, which seems a bit silly.

Maybe we should use `github.com/google/uuid` instead? The only difference is we need to use `uuid.Parse()` instead of `uuid.FromString()`.

srlobo: Thanks for the suggestion, indeed it seems a safer alternative. I will do it!

SuperQ:
Just needs a `go mod tidy`.

PR #621, opened 2022-03-04
enableMgrAlert:Return different values according to different states …

yizhenlangyan:
…of performance_schema.replication_group_members column member_state state, and then alarm for MGR. Performance_schema. replication_group_members MEMBER_STATE: returns 0 when not ONLINE. Configure rules in Prometheus for sending MGR non-online alarm.

Signed-off-by: jianshenghan <jianshenghan@163.com>


yizhenlangyan:
@SuperQ  Please check it. Thanks you!

yizhenlangyan:
@SuperQ  Does this code need to be merged? Is it necessary？ What should I do to merge the code?

ArthurSens:
Hi, sorry for the delayed review here. Per the [OpenMetrics spec](https://prometheus.io/docs/specs/om/open_metrics_spec/), info metrics must always have a value of 1. We want to maintain compliance with OpenMetrics, so this change doesn't make much sense.

If you want to alert when members aren't online, you can easily adjust your query without this change.

That said, I'm closing this PR

PR #625, opened 2022-03-29
Add loki logs dashboard to MySQL mixin (optional) 

v-zhuravlev:
This PR adds optional dashboard with loki logs to MySQL mixin. (off by default)

There are also dashboard links to quickly navigate between two main and logs dashboards.

image

image

- mixin: Conditional dashboard "MySQL Logs" that can be enabled/disabled by `enableLokiLogs` flag in config
- mixin: links to navigate between MySQL Logs and MySQL Overview with context
- mixin: renamed $datasource to $prometheus_datasource, added $loki_datasource
- mixin: Updated README with configuration snippets and sample screenshots

To enable logs support in MySQLd mixin, enable them in config.libsonnet first:

```
{
  _config+:: {
    enableLokiLogs: true,
  },
}

```

then run
```bash
$ make build


ArthurSens:
Hi, sorry for the delayed review here. The mixin project includes only the Prometheus datasource; adding Loki logs would create an empty panel for users without Loki.

Issue #633, opened 2022-05-06
Systemd $HOME variable


d4nnyx:
I'm trying to run mysqld_exporter as a systemd service. Problem is that exporter tries to find $HOME/.my.cnf, but systemd does not inherit this kind of environment variables. So only solution could by define `Environment=HOME=%h` in [Service] section of the service unit. But `%h`  is not dependent on `User=` settings, so I'm forced now to use `root` user for running exporter. 

Do you know some other way how to always get the right $HOME variable in systemd?

roidelapluie:
> $USER, $LOGNAME, $HOME, $SHELL
> User name (twice), home directory, and the login shell. The variables are set for the units that have User= set, which includes>  user systemd instances. See [passwd(5)](http://linux.die.net/man/5/passwd)

https://www.freedesktop.org/software/systemd/man/systemd.exec.html#Environment%20Variables%20Set%20or%20Propagated%20by%20the%20Service%20Manager

PR #635, opened 2022-05-06
Don't use HOME env in the my-cnf config

SuperQ:
Use a relative file as the default for the .my.cnf style configuration.

Fixes: https://github.com/prometheus/mysqld_exporter/issues/633

Signed-off-by: SuperQ <superq@gmail.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,11 @@
 
 BREAKING CHANGES:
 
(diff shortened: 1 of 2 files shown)


roidelapluie:
That's a pretty big & breaking change. I know we are allowed to do those changes, but can't we find a backwards compatible way?

PR #646, opened 2022-07-03
Fixup mixin linting

SuperQ:
Update various items to make the mixin linter happy.

Signed-off-by: SuperQ <superq@gmail.com>

### DIFF
--- a/mysqld-mixin/alerts/galera.yaml
+++ b/mysqld-mixin/alerts/galera.yaml
@@ -16,7 +16,7 @@ groups:
       severity: warning
     annotations:
       description: '{{$labels.job}} on {{$labels.instance}} is not ready.'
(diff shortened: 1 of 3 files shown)


tomwilkie (approved):
LGTM!

Looks like we're pulling in [mixtool:latest](https://github.com/prometheus/mysqld_exporter/blob/main/.circleci/config.yml#L47); its under pretty heavy development right now, do you want me to pin it?  I suspect we'll be in a place to do a 0.1 in a few months.

SuperQ:
Yes, probably good to switch to pinning eventually. I've had a lot of surprise cleanup due to blanket breaking policy changes in the dashboard linter. For example, requiring every panel to have a description. :-/

PR #649, opened 2022-07-27
Add support for single mysqld_exporter collect multiple MySQL instances

starsliao:
### Add support for single mysqld_exporter collect multiple MySQL instances

- Example：
```
curl 'http://localhost:9104/probe?dsn=mysql_user1:password1@tcp(196.168.1.1:3306)/'
curl 'http://localhost:9104/probe?dsn=mysql_user2:password2@tcp(196.168.2.2:3306)/'
```

### DIFF
--- a/collector/exporter.go
+++ b/collector/exporter.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"net/http"
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


SuperQ:
Supplying credentials in the URL params is not allowed as this is a security vulnerability.

Please take a look at https://github.com/prometheus-community/postgres_exporter/pull/618 as a pattern for which this can be implemented.

Issue #306, opened 2018-07-18
Change ENV configuration options


SuperQ:
# Problem
The current `DATA_SORUCE_NAME` is a single ENV var that controls the connection string.

Advantage:
* Single ENV var for config

Disadvantages:
* We have to parse and modify the connect string to add extra connection param.
* Having to template connection and auth in the same ENV is cumbersome.
* We would have to do a lot more code to validate input params.
* It's harder to include defaults, like the connection params separate from the auth.

# Proposal

Split the connection ENV into separate variables to allow easier validation and fewer bugs.

```bash
MYSQLD_EXPORTER_AUTH='user:pass'
MYSQLD_EXPORTER_CONNECTION='127.0.0.1:3306'
```

We can also automatically detect and configure a unix socket connection this way:
```bash
MYSQLD_EXPORTER_CONNECTION='/var/run/mysqld/mysqld.sock'
```

# Related
* [`?` in the password causes parsing errors](https://github.com/prometheus/mysqld_exporter/issues/303)
* [The upstream DSN documentation](https://github.com/go-sql-driver/mysql#dsn-data-source-name)

brian-brazil:
In line with the rest of the project and general security principles, we shouldn't be passing in information via the environment at all. I'd suggest that this be made to come only from a file. This also gives us one source of truth.

SuperQ:
I thought passing in credentials from an ENV was the correct security practice. We do this for several parts of the Prometheus config.

SuperQ:
For sure, we can move the connection option to a command line param, that would probably be better.

brian-brazil:
No part of the Prometheus configuration comes from the environment, it's all from flags and the config file, however some libraries we use happen to read the environment themselves (e.g. AWS libs) and our security model warns about this. Secrets should not go in the environment, as it was never designed with security in mind and it is common practice for example to dump the full environment as part of collecting debug logs. See http://movingfast.io/articles/environment-variables-considered-harmful/

> For sure, we can move the connection option to a command line param, that would probably be better.

You could do that, but not for the password as it'd be exposed to everyone on the same machine.

Given this exporter already reads connection information my.cnf, I'd suggest standardising on pulling connection information only from there (we usually use YAML for config files, but my.cnf seems a reasonable exception here).

You could do it all via flags too, with a flag pointing to a file for the password. Given it's best if config only comes from one place, I'm not personally sure this is worth losing my.cnf support for though.

[...]

PR #651, opened 2022-07-29
Support for scraping multiple mysqld hosts

MattiasAng:
This is a continuation on the work done by [samitpal](https://github.com/samitpal) in #504 to address PR comments and add `module` parameter support.


SuperQ:
Nice, thanks for picking this up.

SuperQ:
CC @syadmind who is working on similar support for the postgres_exporter.

We're using an INI style file here, rather than yaml, since it's more MySQL style.

MattiasAng:
I am wondering if having the `target` for default config is really necessary if users specifically need to enable support for exporting multiple instances? To me it feels like it only adds complexity with no real benefit.

sysadmind (commented):
I think that this PR is functionally correct. I would prefer to see this align with the `/probe` pattern as outlined by the prometheus docs, but otherwise, I think things look good in general.

MattiasAng:
@sysadmind Thanks, I will have a look at refactoring in to `/probe` endpoint but can't promise when yet as I have to prioritise other work.

MattiasAng:
@sysadmind  @SuperQ Probe endpoint is added now. Config could probably be handled a bit better as @sysadmind has done on postgres_exporter but due to time constraint I only refactored the parsing.

I have yet to update / clean up documentation so ignore that for now, would appreciate any feedback on the implementation.

SuperQ:
Nice, thanks. I'll take a look at this soon.

probe.go
```diff
@@ -0,0 +1,90 @@
...
+		registry.MustRegister(collector.New(ctx, dsn, metrics, filteredScrapers, logger))
+
+		if err != nil {
+			probeSuccessGauge.Set(0)
+			probeDurationGauge.Set(time.Since(start).Seconds())
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+
+		duration := time.Since(start).Seconds()
```
sysadmind, line 84: I'm not sure that the duration is accurate because this will be set before the actual metrics are collected below.

mysqld_exporter.go
```diff
@@ -104,8 +103,7 @@ var scrapers = map[collector.Scraper]bool{
 	collector.ScrapeReplicaHost{}:                         false,
 }
 
-func parseMycnf(config interface{}) (string, error) {
-	var dsn string
+func parseMycnf(config interface{}) (dsn string, err error) {
```
sysadmind, line 106: typically we try and stay away from named returns. I would remove the names here.

sysadmind (changes requested):

probe.go
MattiasAng: You are right, and also quite unnecessary since the collector handles timing as well.
Removed it.

MattiasAng:
Updated documentation and some overall cleanup from the previous implementation.

Do you want me to add anything to the changelog or do you handle that on release?

sysadmind (approved):
LGTM @SuperQ should be able to comment on the Changelog. I'm not familiar with the release process for this project.

SuperQ:
So I mentioned removing ENV var support in https://github.com/prometheus/mysqld_exporter/issues/306.

I'm wondering if we should include some basic command line / ENV replacement for this.

Maybe if we just had `MYSQLD_EXPORTER_PASSWORD`.

This could help issues like https://github.com/prometheus/mysqld_exporter/issues/458, we could properly escape passwords in the connection string to the MySQL driver.

I think we still want to support the sidecar use case easily, without a config file for credentials.

MattiasAng:
> Maybe if we just had `MYSQLD_EXPORTER_PASSWORD`.

If we want to support sidecar I would say we would probably need to have both `MYSQLD_EXPORTER_PASSWORD` and `MYSQLD_EXPORTER_USER`, since I think (or hope?) that most users have a separate user for the exporter.

My proposal if we want to have this replacement for `DATA_SOURCE_NAME` is that we only implement it for the `/metrics` endpoint and that the environment variables takes precedence if they are set.

SuperQ:
For configuration that's not in the file, we prefer command line arguments over env vars. The only exception needed is for secrets.

So rather than `MYSQLD_EXPORTER_USER`, we would want `--mysqld.username`.

README.md
```diff
@@ -112,6 +147,13 @@ web.listen-address                         | Address to listen on for web interf
 web.telemetry-path                         | Path under which to expose metrics.
 version                                    | Print the version information.
 
+### Environment Variables
+Name                                       | Description
+-------------------------------------------|--------------------------------------------------------------------------------------------------
+MYSQLD_EXPORTER_HOST                       | Hostname to be used for connecting to MySQL Server
+MYSQLD_EXPORTER_USER                       | Username to be used for connecting to MySQL Server
```
SuperQ, line 154: These two should be command line flags, not env vars.

MattiasAng:
> For configuration that's not in the file, we prefer command line arguments over env vars. The only exception needed is for secrets.
> 
> So rather than `MYSQLD_EXPORTER_USER`, we would want `--mysqld.username`.

Thanks for commenting so fast.

While I understand your line of thinking, I don't fully agree with going for that approach.

If a user still needs to set an environment variable (for the password) I feel that we are better off having all of these as environment variables. The only use-case where you don't need to set an environment variable in that case is for servers that do not require a password.

So in short, I would rather keep it as-is now, but if you insist it should be quite easy to fix by bringing in the arguments to `ReloadConfig`.

SuperQ:
I do insist, This is the require design pattern for exporters. Any non-secret options should be command line arguments to follow our design pattern.

mysqld_exporter.go
```diff
@@ -59,11 +54,21 @@ var (
 		"config.my-cnf",
 		"Path to .my.cnf file to read MySQL credentials from.",
 	).Default(path.Join(os.Getenv("HOME"), ".my.cnf")).String()
+	mysqldHost = kingpin.Flag(
+		"mysqld.host",
+		"Hostname to use for connecting to MySQL",
+	).Default("localhost").String()
```
SuperQ, line 60: Shouldn't this be `mysql.address`? With a default of `localhost:3306`? 

Then `ReloadConfig()` can use `net.SplitHostPort()` to get the host and port.

MattiasAng: Good point, for some reason I was thinking that I had added a `port` flag as well.

Decided against `net.SplitHostPort()` since it doesn't allow port to be optional, which would cause errors if a user specifies the `mysqld.address` without a port.

config/config.go
```diff
@@ -0,0 +1,237 @@
...
+		certPairs = append(certPairs, keypair)
+		tlsCfg.Certificates = certPairs
+	}
+	tlsCfg.InsecureSkipVerify = m.TlsInsecureSkipVerify
+	mysql.RegisterTLSConfig("custom", &tlsCfg)
+	return nil
+}
+
+func parseAddress(address string) (string, string) {
+	target := strings.Split(address, ":")
```
SuperQ, line 227: I'd prefer to use Go `net.SplitHostPort()`. This will break with IPv6 addresses.

https://pkg.go.dev/net#SplitHostPort

SuperQ (changes requested):

SuperQ (approved):
Awesome, thanks so much for all the work and requested changes.

Makefile
```diff
@@ -22,8 +22,8 @@ STATICCHECK_IGNORE =
 
 DOCKER_IMAGE_NAME ?= mysqld-exporter
 
-test-docker:
-	@echo ">> testing docker image"
+test-docker-single-exporter:
```
roidelapluie, line 25:
```suggestion
.PHONY: test-docker-single-exporter
test-docker-single-exporter:
```

README.md
```diff
@@ -112,6 +149,15 @@ web.listen-address                         | Address to listen on for web interf
 web.telemetry-path                         | Path under which to expose metrics.
 version                                    | Print the version information.
 
+### Environment Variables
+Name                                       | Description
+-------------------------------------------|--------------------------------------------------------------------------------------------------
+MYSQLD_EXPORTER_PASSWORD                   | Password to be used for connecting to MySQL Server
```
roidelapluie, line 155: we generally avoid passwords as env variables

config/config.go
```diff
@@ -0,0 +1,229 @@
+// Copyright 2018 The Prometheus Authors
```
roidelapluie, line 1:
```suggestion
// Copyright 2022 The Prometheus Authors
```

config/config.go
```diff
@@ -0,0 +1,229 @@
...
+	if cfg, err = ini.LoadSources(
+		opts,
+		[]byte("[client]\npassword = ${MYSQLD_EXPORTER_PASSWORD}\n"),
+		filename,
+	); err != nil {
+		return fmt.Errorf("failed to load %s", filename)
+	}
+
+	if host, port, err = net.SplitHostPort(mysqldAddress); err != nil {
+		return fmt.Errorf("failed to parse address: %s", err)
```
roidelapluie, line 106:
```suggestion
		return fmt.Errorf("failed to parse address: %w", err)
```

config/config.go
```diff
@@ -0,0 +1,229 @@
...
+			configReloadSeconds.SetToCurrentTime()
+		}
+	}()
+
+	if cfg, err = ini.LoadSources(
+		opts,
+		[]byte("[client]\npassword = ${MYSQLD_EXPORTER_PASSWORD}\n"),
+		filename,
+	); err != nil {
+		return fmt.Errorf("failed to load %s", filename)
```
roidelapluie, line 102: return the error as well

config/config.go
```diff
@@ -0,0 +1,229 @@
...
+		}
+	}
+
+	cfg.ValueMapper = os.ExpandEnv
+	config := &Config{}
+	m := make(map[string]MySqlConfig)
+	for _, sec := range cfg.Sections() {
+		sectionName := sec.Name()
+
+		if sectionName == "DEFAULT" {
```
roidelapluie, line 127: we should skip client section as well

config/config.go
```diff
@@ -0,0 +1,229 @@
...
+	} else {
+		if host, port, err = net.SplitHostPort(target); err != nil {
+			return dsn, fmt.Errorf("failed to parse target: %s", err)
+		}
+		dsn = fmt.Sprintf("%s:%s@tcp(%s:%s)/", user, password, host, port)
+	}
+
+	if m.SslCa != "" {
+		if tlsErr := m.CustomizeTLS(); tlsErr != nil {
+			tlsErr = fmt.Errorf("failed to register a custom TLS configuration for mysql dsn: %s", tlsErr)
```
roidelapluie, line 195:
```suggestion
			tlsErr = fmt.Errorf("failed to register a custom TLS configuration for mysql dsn: %w", tlsErr)
```

config/config.go
```diff
@@ -0,0 +1,229 @@
...
+		if socket != "" {
+			dsn = fmt.Sprintf("%s:%s@unix(%s)/", user, password, socket)
+		} else {
+			dsn = fmt.Sprintf("%s:%s@tcp(%s:%d)/", user, password, host, port)
+		}
+	} else {
+		if host, port, err = net.SplitHostPort(target); err != nil {
+			return dsn, fmt.Errorf("failed to parse target: %s", err)
+		}
+		dsn = fmt.Sprintf("%s:%s@tcp(%s:%s)/", user, password, host, port)
```
roidelapluie, line 190: what if a password contains : ? shouldn't we build this more carefully?

config/config.go
```diff
@@ -0,0 +1,229 @@
...
+	if ok := caBundle.AppendCertsFromPEM(pemCA); ok {
+		tlsCfg.RootCAs = caBundle
+	} else {
+		return fmt.Errorf("failed parse pem-encoded CA certificates from %s", m.SslCa)
+	}
+	if m.SslCert != "" && m.SslKey != "" {
+		certPairs := make([]tls.Certificate, 0, 1)
+		keypair, err := tls.LoadX509KeyPair(m.SslCert, m.SslKey)
+		if err != nil {
+			return fmt.Errorf("failed to parse pem-encoded SSL cert %s or SSL key %s: %s",
```
roidelapluie, line 220:
```suggestion
			return fmt.Errorf("failed to parse pem-encoded SSL cert %s or SSL key %s: %w",
```

config/config_test.go
```diff
@@ -0,0 +1,173 @@
+// Copyright 2018 The Prometheus Authors
```
roidelapluie, line 1:
```suggestion
// Copyright 2022 The Prometheus Authors
```

mysqld_exporter.go
```diff
@@ -59,11 +54,21 @@ var (
 		"config.my-cnf",
 		"Path to .my.cnf file to read MySQL credentials from.",
 	).Default(path.Join(os.Getenv("HOME"), ".my.cnf")).String()
+	mysqldAddress = kingpin.Flag(
+		"mysqld.address",
+		"Address to use for connecting to MySQL",
+	).Default("localhost:3306").String()
```
roidelapluie, line 60: should we just use localhost and connect via unix socket by default?

probe.go
```diff
@@ -0,0 +1,76 @@
+// Copyright 2018 The Prometheus Authors
```
roidelapluie, line 1:
```suggestion
// Copyright 2022 The Prometheus Authors
```

README.md
SuperQ: This is the one place where I am allowing this, as I want a way to avoid needing a config file for simple deployments.

config/config.go
SuperQ: This has been a long-standing issue with various Go SQL drivers. They all seem to suffer from this parsing problem.

There's https://pkg.go.dev/github.com/go-sql-driver/mysql?utm_source=godoc#Config which should probably be used here.

mysqld_exporter.go
SuperQ: The current default is localhost, I'd rather stick with that.

config/config.go
MattiasAng: That would break configuration for those that are currently using it with a config file rather than `DATA_SOURCE_NAME`, are we OK with that?

SuperQ: No, I think we need to keep this so it continues to work as-is.

config/config.go
MattiasAng: I am wondering if we couldn't leave this for another issue to rework? This PR is already adding quite a lot.

It will require quite a fair amount of work to pass that along to the Exporter and parse parameters, and my time is quite limited right now.

SuperQ (approved):

roidelapluie:
LGTM

SuperQ:
Let's do this!

config/config.go
SuperQ: That's fine. Let's work on it next.

yinshurman:
so how to configure multiple hosts and ports ? 
when i set host and port in different section of mysql.cnf , just the one in [client] makes effect.
and when i only set user and password in client ,and configure prometheus.yml file like this:
`
  - job_name: "mysql" # to get metrics  about mysql 
    params:
        # not required. default `client`
        auth_module: [client]
    static_configs:
      - targets: 
        # all mysql hostname to monitor 
        # - '10.203.4.249:10003'
        # - '10.203.4.249:10016'
        - '172.22.0.5:3306'
        - '172.22.0.11:3307'
        - '172.22.0.12:3308'
    relabel_configs:
      - source_labels: [__address__]
        target_label: __param_target
      - source_labels: [__param_target]
        target_label: instance 
      - target_label: __address__ 
        # the mysqld exporter host:port
        replacement: 172.22.0.6:9104
`
the mysqld exporter reports error information like this :
`ts=2022-10-14T02:05:48.270Z caller=exporter.go:149 level=error msg="Error pinging mysqld" err="dial tcp 127.0.0.1:3306: connect: connection refused"`


so , it seems must set host and port in some where other than prometheus.yml , if not ,the defautl  127.0.0.1:3306 is used .

any one can make it clear  where to set the different hosts and ports ? 
please help ! thanks very much !

Suomea:
@yinshurman looks like not release yet?

Suomea:
> The documentation is missing that `metrics_path` needs to be changed to `/probe`.
> 
> I.e.
> 
> ```
>   - job_name: "mysql" # to get metrics  about mysql 
>     metrics_path: /probe
> ```
I want to use multi-target to avoid starting multiple processes on machine. my situation  ->
mysqld_exporter and prometheus version
```
root@m1:~# ls -l
total 89184
-rw-r--r-- 1 root root  7450991 Oct 18 18:03 mysqld_exporter-0.14.0.linux-amd64.tar.gz
-rw-r--r-- 1 root root 83865783 Oct 18 18:06 prometheus-2.37.1.linux-amd64.tar.gz
```

mysqld_exporter config
```
# /srv/metrics/mysqld_exporter/my.cnf
[client]
user = root
password = 123456
```

prometheus config
```
scrape_configs:
  # The job name is added as a label `job=<job_name>` to any timeseries scraped from this config.
  - job_name: "prometheus"

    # metrics_path defaults to '/metrics'
    # scheme defaults to 'http'.

    static_configs:
      - targets: ["localhost:9090"]

  - job_name: "mysqld_exporter"
    metrics_path: /probe
    static_configs:
      - targets:
        - 172.27.145.40:3306
        - 172.27.145.41:3306
    relabel_configs:
      - source_labels: [__address__]
        target_label: __param_target
      - source_labels: [__param_target]
        target_label: instance
      - target_label: __address__
        # The mysqld_exporter host:port
        replacement: localhost:9104
```

The prb is the targets is down
image

I don't know why this is，i think maybe the exporter verions is wrong? or something else?

MattiasAng:
You need to compile it yourself, there is no release for the commits in this PR.

shubhamsetia05:
@MattiasAng Any plans to release this (support to monitor multiple mysql DB's) in helm? We also have a use case to monitor multiple AWS RDS instances.

PR #687, opened 2022-11-29
Fix infoSchemaInnodbMetricsEnabledColumnQuery

cezmunsta:
The query introduced with #523 performs poorly when there are sufficient numbers of rows in `information_schema.columns`.

* Added using the `table_schema` as a filter
* Added limit to the query

### DIFF
--- a/collector/info_schema_innodb_metrics.go
+++ b/collector/info_schema_innodb_metrics.go
@@ -31,8 +31,10 @@ const infoSchemaInnodbMetricsEnabledColumnQuery = `
 	SELECT
 	    column_name
 	  FROM information_schema.columns
-	  WHERE table_name = 'INNODB_METRICS'
+	  WHERE table_schema = 'information_schema'
+	    AND table_name = 'INNODB_METRICS'
 	    AND column_name IN ('status', 'enabled')
+	  LIMIT 1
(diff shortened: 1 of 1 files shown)


atymchuk:
> when there are sufficient numbers of columns in information_schema.columns

@cezmunsta I wonder, what would the definition of `sufficient` be?

cezmunsta:
> @cezmunsta I wonder, what would the definition of `sufficient` be?

@atymchuk other factors influence it, such as whether or not it needs to open tables (`information_schema` pre-8.0 MySQL). On servers where there are extremely high numbers of _tables_, even certain OS commands should be avoided for the MySQL datadir.
The PMM client includes a safety net for [`pmm-admin` for tablestats](https://docs.percona.com/percona-monitoring-and-management/setting-up/client/mysql.html#tablestats), with the default being 1000 _tables_. There is also a simple query performance example on the [PMM ticket](https://jira.percona.com/browse/PMM-11082): 
```sql
SELECT
column_name
FROM information_schema.columns
WHERE table_name = 'INNODB_METRICS';
```
That alone took `3.56 sec` for the user.

SuperQ (approved):
Makes sense, thanks

PR #691, opened 2022-12-05
Fix unclosed code block in README.md

ericyan:
This fixes a code block that not properly closed, causing the next two sections rendered as code.


ericyan:
@SuperQ could you review this?

SuperQ (approved):
Thanks

Issue #663, opened 2022-09-10
probe_success gauge is set 0 on success


PSSGCSim:
### Host operating system: output of `uname -a`

`Linux XXX 5.10.0-17-amd64 #1 SMP Debian 5.10.136-1 (2022-08-13) x86_64 GNU/Linux`

but using docker image from :main tag

### mysqld_exporter version: output of `mysqld_exporter --version`

```
mysqld_exporter, version 0.14.0 (branch: main, revision: 593b0095a5bd0bd852029dd2dfa5f204a15de946)
  build user:       root@846c6cdf111e
  build date:       20220901-14:48:58
  go version:       go1.18.5
  platform:         linux/amd64
```

### MySQL server version

`mysql_version_info{innodb_version="10.8.3",version="10.8.3-MariaDB-1:10.8.3+maria~jammy",version_comment="mariadb.org binary distribution"} 1`

### mysqld_exporter command line flags

`--config.my-cnf=/config/my.cnf`

### What did you do that produced an error?

Test exporter with:

`curl http://localhost:9104/probe?auth_module=client&target=server123.domain.tld:3306`

### What did you expect to see?

probe_success 1

### What did you see instead?

probe_success 0

### Comment

The console output of mysqld_exporter contains no errors and the exporter returns all metrics as expected except for probe_success. I think the problem is located at following lines https://github.com/prometheus/mysqld_exporter/blob/593b0095a5bd0bd852029dd2dfa5f204a15de946/probe.go#L67-L71 where the logic is opposite to blackbox_exporter which uses 0=failure 1=success.

MattiasAng:
I introduced it and I got it from a different project (no blaming here). Would argue that it is quite useless with current implementation, and needs a bit of refactoring. Would recommend that you rather rely on mysql status metric.

PR #714, opened 2023-03-07
Fix a mixture of error messages and metrics is returned from the multiple target endpoint.

yosida95:
This PR addresses #663 and the problem I mentioned in the description of #707.

Currently (8fa4a07777dd26bc29afb26c1ecdfc3c68ca9f69), the `/probe` endpoint emits both error messages and metrics if one of the following situations is met.

- config file does not contain the section which is requested by the `auth_module` parameter.
- network address which is requested by the `target` parameter is ill-formed.

```
Error parsing config section [non-existent]
# HELP mysql_exporter_last_scrape_error Whether the last scrape of metrics from MySQL resulted in an error (1 for error, 0 for success).
# TYPE mysql_exporter_last_scrape_error gauge
mysql_exporter_last_scrape_error 1
# HELP mysql_exporter_scrapes_total Total number of times MySQL was scraped for metrics.
# TYPE mysql_exporter_scrapes_total counter
mysql_exporter_scrapes_total 1
# HELP mysql_up Whether the MySQL server is up.
# TYPE mysql_up gauge
mysql_up 0
# HELP probe_success Displays whether or not the probe was a success
# TYPE probe_success gauge
probe_success 0
```

This PR removes metrics to emit only a single line of the error message.

It also removes `probe_success` metrics introduced in 593b0095a5bd0bd852029dd2dfa5f204a15de946 (not yet released) for the following reasons:

  - It will always be 1 (in case of #663 fixed; otherwise 0) even if failed to connect to the MySQL server.
  - Existing `mysql_up` parameter can be used instead.
  - `/metrics` endpoint does not contain it.

### DIFF
--- a/probe.go
+++ b/probe.go
@@ -26,9 +26,6 @@ import (
 
 func handleProbe(metrics collector.Metrics, scrapers []collector.Scraper, logger log.Logger) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
-		var dsn, authModule string
-		var err error
-
 		ctx := r.Context()
 		params := r.URL.Query()
 		target := params.Get("target")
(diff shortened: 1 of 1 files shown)


SuperQ (approved):
Makes sense, thanks

Issue #640, opened 2022-06-13
Missing grant for MariaDB


PhilipMay:
I am using MariaDB 10.7.4.

If I follow the README.md "[Required Grants](https://github.com/prometheus/mysqld_exporter#required-grants)" the exporter throws this log:

```
ts=2022-06-13T17:21:26.136Z caller=exporter.go:174 level=error msg="Error from scraper" scraper=slave_status err="Error 1227: Access denied; you need (at least one of) the SUPER, SLAVE MONITOR privilege(s) for this operation"
```

So my solution was: `GRANT SLAVE MONITOR ON *.* TO 'exporter';`

Maybe you want to add this to the doc?

PhilipMay:
@SuperQ what do you think?

mxsxs2:
I jus encountered this very same "Error". I think it should be added to the docs too!

OscarVanL:
Same problem here. The README should be updated to state this grant is also needed for MariaDB monitoring.

seth-lohr-sonic-com:
Hello there.  Adding `GRANT SLAVE MONITOR ON *.* TO 'exporter';` doesn't seem to work for me.

Running version `0.15.1-1.el9` of mysqld_exporter.x86_64
Running `Server version: 10.5.22-MariaDB MariaDB Server`

When giving the exporter the following to statements:

What the README.md says
`GRANT PROCESS, REPLICATION CLIENT, SELECT ON *.* TO 'exporter'@'localhost';`

What PhilipMay says to do:
`GRANT SLAVE MONITOR ON *.* TO 'exporter';`

I still don't see the right permissions in pt-show-grants

`-- Grants for 'exporter'@'localhost'
GRANT BINLOG MONITOR, PROCESS, SELECT ON *.* TO `exporter`@`localhost` IDENTIFIED BY PASSWORD '*1C2151F0C2AC2DBE50683B4AC0C32344967508A7' WITH MAX_USER_CONNECTIONS 3;
`

I think from the reading I've done in mariadb docs, there's *something changing in behavior with REPLICATION CLIENT to not be setting what this exporter is expecting properly, but I haven't been able to trace the root cause yet: https://mariadb.com/docs/server/ref/mdb/privileges/REPLICATION_CLIENT/

And https://mariadb.com/docs/server/ref/mdb/privileges/BINLOG_MONITOR/

Based on some of the tricky wordage here, I'm not sure if SUPER is actually needed or not?

[...]

PR #719, opened 2023-03-16
#640 updated documentation for GRANT SLAVE MONITOR

mmethner:
@SuperQ Updated documentation for newer mysql/mariadb versions for the message "Error 1227: Access denied; you need (at least one of) the SUPER, SLAVE MONITOR privilege(s) for this operation".


SuperQ:

mmethner:
Hi @SuperQ I am not too experienced with github, so this may is a foolish question for you. I signed the commit and pushed, but it is signed with another email that I use to login to github. Must the e-mail be the same in the commit message than the username here?

PR #723, opened 2023-04-11
Refactor mysql_up metric

SuperQ:
Use a return value on the scrape function to only send the `mysql_up` metric down the channel once.


SuperQ:
Had this idea after I merged #722. CC @LeoQuote

Nexucis (approved):
yeah that looks cleaner to me !

PR #727, opened 2023-04-25
use the target for dsn building in colector

tinpotnick:
It looks like that in the latest release; there is a provision for multiple targets.

> To use the multi-target functionality, send an http request to the endpoint /probe?target=foo:3306 where target is set to the DSN of the MySQL instance to scrape metrics from.

I used this mechanism but the scraper only tried to scrape itself (127.0.0.1). When you set the mysqld.address flag - then it scrapes that host - for all targets.

This fixes that by using the target in the query string as part of the DSN building.


tinpotnick:
Correct PR.

PR #728, opened 2023-04-25
Use target in DSN builder.

tinpotnick:
When I configured mysqld_exporter for multi targets as per the main README:

> To use the multi-target functionality, send an http request to the endpoint /probe?target=foo:3306 where target is set to the DSN of the MySQL instance to scrape metrics from.

If you use this - regardless of the target it uses 127.0.0.1 as the MySQL host. If you use the mysqld.addressflag - then for all targets it uses the mysqld.address host. 

Signed-off-by: Nick Knight <nick@babblevoice.com>


SuperQ (approved):
Nice catch

PR #729, opened 2023-04-25
Create a multi stage Dockerfile build file.

tinpotnick:
The original Dockerfile didn't have a mechanism for building the project. This modifies the Dockerfile to a multistage target for building and release.

Works with buildx for cross-platform building.

Signed-off-by: Nick Knight <nick@babblevoice.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,19 @@
-ARG ARCH="amd64"
-ARG OS="linux"
-FROM quay.io/prometheus/busybox-${OS}-${ARCH}:latest
+
+FROM alpine:3.17 as builder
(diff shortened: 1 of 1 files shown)


SuperQ:
The Prometheus project doesn't use multi-stage builds to keep the size small.

tinpotnick:
Hi,

Thank you for your feedback. Image size is also important to me.

prom/mysqld-exporter              latest                  121b8a7cd052   14 months ago    17.8MB

vs my build

<none>                            <none>                  e6c031e9863c   42 seconds ago   13.3MB

The multi-stage Dockerfile wins on size.

SuperQ:
The size difference is because you are not statically building, and you need the intermediate alpine layers as well. The Prometheus base image is Busybox.

And again, not how we build our project.

PR #734, opened 2023-05-15
Add config.my-cnf hot reload `/-/reload`

winfredwz:
Add config.my-cnf hot reload `/-/reload`


winfredwz:
Add sign-off

mysqld_exporter.go
```diff
@@ -252,7 +252,10 @@ func main() {
 		http.Handle("/", landingPage)
 	}
 	http.HandleFunc("/probe", handleProbe(enabledScrapers, logger))
-
+	http.HandleFunc("/-/reload", func(w http.ResponseWriter, r *http.Request) {
+		c.ReloadConfig(*configMycnf, *mysqldAddress, *mysqldUser, *tlsInsecureSkipVerify, logger)
```
SuperQ, line 256: `ReloadConfig` returns an error, we need to check this error and return an appropriate HTTP status code.

SuperQ (changes requested):
Please add correct error handling of the reload endpoint.

mysqld_exporter.go
```diff
@@ -252,7 +252,13 @@ func main() {
 		http.Handle("/", landingPage)
 	}
 	http.HandleFunc("/probe", handleProbe(enabledScrapers, logger))
-
+	http.HandleFunc("/-/reload", func(w http.ResponseWriter, r *http.Request) {
+		if error := c.ReloadConfig(*configMycnf, *mysqldAddress, *mysqldUser, *tlsInsecureSkipVerify, logger); error != nil {
+			level.Info(logger).Log("msg", "Error reloading host config", "file", *configMycnf, "err", error)
```
SuperQ, line 257: `error` is a Go type, so it can be ambiguous. We generally use `err`. This should also be at least a Warning level error.

```suggestion
		if err = c.ReloadConfig(*configMycnf, *mysqldAddress, *mysqldUser, *tlsInsecureSkipVerify, logger); err != nil {
			level.Warn(logger).Log("msg", "Error reloading host config", "file", *configMycnf, "error", err)
```

SuperQ (approved):

Issue #686, opened 2022-11-29
option to support empty password in cnf


hueyvle:
### Host operating system: output of `uname -a`
uname -a
Linux icmmaster 3.10.0-1160.76.1.el7.x86_64 #1 SMP Wed Aug 10 16:21:17 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

### mysqld_exporter version: output of `mysqld_exporter --version`
node_exporter --version
node_exporter, version 1.1.2 (branch: HEAD, revision: b597c1244d7bef49e6f3359c87a56dd7707f6719)
  build user:       root@f07de8ca602a
  build date:       20210305-09:29:10
  go version:       go1.15.8
  platform:         linux/amd64

### MySQL server version
5.5.68-MariaDB

### mysqld_exporter command line flags

mysqld_exporter --config.my-cnf /home/user/monitoring/conf/.mysql_exporter.cnf --collect.global_status --collect.info_schema.innodb_metrics --collect.auto_increment.columns --collect.info_schema.processlist --collect.binlog_size --collect.info_schema.tablestats --collect.global_variables --collect.info_schema.query_response_time --collect.info_schema.userstats --collect.info_schema.tables --collect.perf_schema.tablelocks --collect.perf_schema.file_events --collect.perf_schema.eventswaits --collect.perf_schema.indexiowaits --collect.perf_schema.tableiowaits --collect.slave_status --web.listen-address=0.0.0.0:9104

cat /home/user/monitoring/conf/.mysql_exporter.cnf 
[client]
user=metricsexporter
password=

### What did you do that produced an error?
1. create user 'metricsexporter'@'%' with no password
2. grant PROCESS, REPLICATION CLIENT, SELECT to user 'metricsexporter'@'%' 
3. I can log in with "mysql -u metricexporter --skip-password"

### What did you expect to see?
mysqld_exporter to run with skip password option when password is set to null

### What did you see instead?
Nov 28 21:56:45 myhost mysqld_exporter: level=info ts=2022-11-29T05:56:45.246Z caller=mysqld_exporter.go:284 msg="Error parsing my.cnf" file=/home/icmanage/monitoring/conf/.mysql_exporter.cnf err="password or ssl-key should be specified under [client] in /home/icmanage/monitoring/conf/.mysql_exporter.cnf"

hueyvle:
I'm looking for an option to use mysqld_exporter with no password. (Yes I know the security risk, it is not under my control)

[...]

PR #742, opened 2023-06-15
Allow empty passwords

SuperQ:
Empty passwords are allowed for UNIX socket connections and when using TLS auth.

Fixes: https://github.com/prometheus/mysqld_exporter/issues/686


SuperQ:
CC @strafer

matthiasr (approved):
I vaguely remember MySQL making a distinction between "no password" and "empty password" in some circumstances. Which one will this end up using on the wire?

SuperQ:
I don't know, but from the perspective of the Go mysql driver, there's no difference. The [DSN Config struct](https://pkg.go.dev/github.com/go-sql-driver/mysql?#Config) defines pasword as a `string`, not a `*string`, so the default is empty string.

PR #745, opened 2023-06-19
Don't use HOME env in the my-cnf config

SuperQ:
Use a relative file as the default for the .my.cnf style configuration.

Fixes: https://github.com/prometheus/mysqld_exporter/issues/633

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,11 @@
 
 BREAKING CHANGES:
 
+Don't prepend the `.my.cnf` file with the `HOME` environment.
+
 Changes:
+
+* [CHANGE] Don't use HOME env in the my-cnf config path. #745
 
 * [CHANGE]
 * [FEATURE]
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,8 +2,12 @@
 
 BREAKING CHANGES:
 
+Don't prepend the `.my.cnf` file with the `HOME` environment.
```
matthiasr, line 5: As a user, this description is not helpful to me. Does this mean I should not use `$HOME` when specifying the flag? What happens differently when I specify some file name?

I _think_ what actually changes from a user perspective is

```suggestion
The default client configuration file is now `.my.cnf` in the process working directory. Use `--config.my-cnf="$HOME/.my.cnf"` to retain the previous default.
```

matthiasr (commented):
The README also states the default and needs an update

SuperQ:
Updated README.md

matthiasr (approved):
✨

Issue #746, opened 2023-06-27
Limit mysqld_exporter to at most 1 set of pending queries running per Prometheus scrape interval (merge fix from PMM)


ColinDKelley:
### Host operating system: output of `uname -a`
```
Linux <pod> 4.19.0-18-cloud-amd64 #1 SMP Debian 4.19.208-1 (2021-09-29) x86_64 GNU/Linux
```
### mysqld_exporter version: output of `mysqld_exporter --version`

0.12.1

### MySQL server version
5.7

### What did you do that produced an error?

We've periodically had a problem in production where the `mysqld_exporter` contributes to database overload. Here are the steps for how that happened:

1. Our production database got bogged down running a very slow query.
2. Prometheus hit mysqld_exporter at a scrape interval.
3. The mysqld_exporter query took more than the scrape interval to run. So before it could complete, we looped back to (2). Repeat.

Note: We submitted a fix for this 2 years ago on the Percona (PMM) fork. That has been running successfully ever since. Here is [that issue](https://jira.percona.com/browse/PMM-4557) and [the PR](https://github.com/percona/mysqld_exporter/pull/56).

Also note, [this ticket is related](https://github.com/prometheus/mysqld_exporter/issues/725), although not a duplicate, since there are many possible causes of query slowdown.

### What did you expect to see?

We expected the first mysqld_exporter query to return results as soon as possible and later attempts to run the same query to return a 429 "Too Many Requests" error.

### What did you see instead?

When the query time exceeded the scrape interval, we saw an unbounded stack of mysqld_exporter queries all running at the same time. These queries snowballed until the `mysqld_exporter` ran out of memory and was OOM-killed.

PR #747, opened 2023-06-27
issue #746: return 429 Too Many Requests if requests stack up

ColinDKelley:
This is a PR for issue #746.

It is a direct port from https://github.com/percona/mysqld_exporter/pull/56.

### DIFF
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -132,7 +132,16 @@ func init() {
 }
 
 func newHandler(scrapers []collector.Scraper, logger log.Logger) http.HandlerFunc {
+	processing := false
 	return func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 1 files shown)


ArthurSens:
Hey 👋, sorry for the super-delayed review. I think the idea is pretty good, but since the codebase has evolved a lot over the past 3 years, it might be worth starting the PR from scratch if you still want to contribute.

There are a few things that we will need to keep in mind:
* We now have another endpoint called `/probe` that might need the same implementation, probably a good idea to implement somthing re-usable between /probe and /metrics
* Doesn't 503 fit better than 429 here? For 429, I would expect clients to implement retry mechanisms, and it doesn't sound like a good fit here.
* We definitely need tests


That said, I'm closing the PR. A new one would be very welcome!

PR #758, opened 2023-07-24
Mixin: Switch to timeseries and improve stat panels.

v-zhuravlev:
- Switch to timeseries
- Use 'last' in stat panels instead of mean (like, for uptime).
- Chain job and instance variables

image


ArthurSens:
Hi, sorry for the super delayed review here. The mixin has evolved quite a lot in the last few years, so it's probably a better idea to restart the work on top of latest main. Apologies for that.

I'll be closing the PR for now, but feel free to open a new one if you're still willing to contribute to the mixins

PR #760, opened 2023-08-12
Fix race condition in ReloadConfig

ajmaidak:
Calling ReloadConfig was not thread safe in spite of the mutex.  One thread could overwrite the package level cfg pointer after it was mutated by another thread.


SuperQ:

SuperQ (approved):
I'm going to ignore the mixin error, it's unrelated to this change.

Issue #759, opened 2023-07-26
Feature request: support collector configuration via config file


maxenglander:
I would like the ability to dynamically change active collectors, so that when a database is having issues I can turn on more collectors, and turn them back off after the issue is resolved.

What I have in mind to make this work:

 1. Support configuring collectors via config file (this issue).
 2. Support reloading collector configuration from config file with `/-/reload` endpoint is touched.

SuperQ:
We have no plans to create a configuration for this.

maxenglander:
@SuperQ would you be open to a contribution?

PR #764, opened 2023-08-22
Support collector config via file

maxenglander:
I would like to have the ability to `/-/reload` the scraper configuration. Currently scrapers are only configurable via `--collect.<name>[.arg]` and cannot be reconfigured at runtime, so this proposal also entails allowing collectors to be configured via a file, and then to have the ability to reconfigure scrapers when the contents of that file change and `/-/reload` is called.

I opened a GH issue https://github.com/prometheus/mysqld_exporter/issues/759. It was closed by Ben saying that the maintainers had no plans to implement, but didn't say whether they would be open to a contribution, hoping that they are.


SuperQ:
We already support dynamic collection via URL params. This is not necessary.

PR #765, opened 2023-08-26
Heartbeat collector query using palceholders

malderete:
Improves how heartbeat generates the SQL query. Reduce the type of errors produced by "wrong" arguments.

Use argument's placeholders and [QueryContext](https://pkg.go.dev/database/sql#Conn.QueryContext) instead of using [ftm.Sprintf](https://pkg.go.dev/fmt#Sprintf) to generate the query.

@SuperQ, can you review it please?

### DIFF
--- a/collector/heartbeat.go
+++ b/collector/heartbeat.go
@@ -18,7 +18,6 @@ package collector
 import (
 	"context"
 	"database/sql"
-	"fmt"
 	"strconv"
 
 	"github.com/alecthomas/kingpin/v2"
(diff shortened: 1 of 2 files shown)


collector/heartbeat.go
```diff
@@ -33,7 +32,7 @@ const (
 	// timestamps. %s will be replaced by the database and table name.
 	// The second column allows gets the server timestamp at the exact same
 	// time the query is run.
-	heartbeatQuery = "SELECT UNIX_TIMESTAMP(ts), UNIX_TIMESTAMP(%s), server_id from `%s`.`%s`"
+	heartbeatQuery = "SELECT UNIX_TIMESTAMP(ts), UNIX_TIMESTAMP(?), server_id from ?.?"
```
ArthurSens, line 35: Using `?` for the timestamp could be a good change, but using it in the FROM statement doesn't work; we need to revert that

ArthurSens (commented):
Hey, sorry for the super delayed review here. I have a comment, but a rebase will also be needed

PR #807, opened 2024-01-25
Add graceful shutdown

wanghaowei0107:
When I send kill signal to mysqld_exporter, I meet "Exit Code: 2, Exit Message: "Docker container exited with non-zero exit code: 2". It make my job-watchdog can't know the mysqld_exporter is it a normal exit.

### DIFF
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -15,12 +15,15 @@ package main
 
 import (
 	"context"
+	"errors"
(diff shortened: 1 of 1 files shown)


SuperQ (changes requested):
This should be handled in the exporter-toolkit web package.

wanghaowei0107:
I think mysqld-exporter service need to graceful shutdown. The exporter-toolkit web package don't need it. How do you think?

PR #826, opened 2024-03-07
FEATURE: add statement lock time and cpu time

taylorchu:


collector/perf_schema_events_statements.go
```diff
@@ -97,6 +101,16 @@ var (
 		"The total time of events statements by digest.",
 		[]string{"schema", "digest", "digest_text"}, nil,
 	)
+	performanceSchemaEventsStatementsLockTimeDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, performanceSchema, "events_statements_lock_time_seconds_total"),
+		"The total lock time of events statements by digest.",
+		[]string{"schema", "digest", "digest_text"}, nil,
+	)
+	performanceSchemaEventsStatementsCPUTimeDesc = prometheus.NewDesc(
```
SuperQ, line 109: I think we intentionally left out CPU time to reduce cardinality. Can you show more about why this is needed?

Maybe we should add a flag like `--collect.perf_schema.eventsstatements.extra-timing-metrics`?

taylorchu: why is cpu time having higher cardinality than others in the file? we are using it to track what query is slow and why. so we have wait time, which is total time. these 2 will help us drill down load due to lock, cpu, and io (i assume it is total - lock - cpu, or you know there is a column for this. )

SuperQ: Because it's more metrics. Every additional metric adds cardinality cost. For things like events statements, it's multiplied by every unique statement.

I think we may have had these metrics previously, but we removed them to reduce the load on the monitoring platform(s). We found that `mysql_perf_schema_events_statements_seconds_total` was sufficient for most cases.

I agree, the additional metrics would be useful to do additional drill down, but I think we should put them behind an opt-in feature flag.

SuperQ:

taylorchu:
I am going to close this and add another PR with only lock time

PR #828, opened 2024-03-17
Support for prometheus scrape timeout in probe endpoint

mmontes11:
Hey @SuperQ ! First of all, I wanted to say thank you for the great work done in this exporter and for the effort maintaining it 🙏🏻 

In [mariadb-operator](https://github.com/mariadb-operator/mariadb-operator), we are currently provisioning `mysqld_exporter` as a source of metrics for our `MariaDB` resources. The recent introduction of the [multi-target support](https://github.com/prometheus/mysqld_exporter?tab=readme-ov-file#multi-target-support) was a massive improvement for us, as we are able to provision a single `mysqld_exporter`  `Deployment` that exposes the metrics of all the `MariaDB` replicas. This implies that we no longer need  `mysqld_exporter`  sidecars in our `MariaDB` `StatefulSet`, and more importantly, we can upgrade the exporters without affecting the `MariaDB` availability.

Driven by the curiosity of how the `/probe` endpoint was implemented, I've realized that it doesn't support configuring the Prometheus scrape timeout as the `/metrics` endpoint currently does. This PR separates this functionality into a reusable function, applicable to both `/metrics` and `/probe` endpoints. It is my very first PR here, so let me know if I'm missing something!

Exposing metrics in [mariadb-operator](https://github.com/mariadb-operator/mariadb-operator) without `mysqld_exporter`  would have been way more difficult, so again I wanted to express my gratitude, let me know if there is any way I can keep collaborating with this project!

### DIFF
--- a/mysqld_exporter.go
+++ b/mysqld_exporter.go
@@ -15,6 +15,7 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"os"
 	"strconv"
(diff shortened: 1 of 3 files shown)


SuperQ (approved):
Thanks, the refactoring in the `getScrapeTimeoutSeconds()` function is a lot easier to read.

mmontes11:
@SuperQ no problem! Thanks for the review 🙏🏻

PR #837, opened 2024-05-15
Support MySQL 8.4 replicas syntax

mtanda:
#### Summary

In MySQL 8.4, several "Replication SQL syntax" commands have been deprecated, requiring the use of new syntax. During testing, it was confirmed that errors occur with the old syntax. This PR updates the code to be compatible with the new syntax. Could you please review this PR?

#### Details

The changes address the syntax errors that arise due to the deprecated commands. Below are the specific errors encountered:

```
mysqld-exporter  | ts=2024-05-15T02:39:40.760Z caller=exporter.go:173 level=error msg="Error from scraper" scraper=slave_status target=mysql_replica:3306 err="Error 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'SLAVE STATUS' at line 1"
mysqld-exporter  | ts=2024-05-15T02:43:26.471Z caller=exporter.go:173 level=error msg="Error from scraper" scraper=slave_hosts target=mysql_replica:3306 err="Error 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'SLAVE HOSTS' at line 1"
```

For reference, you can find the MySQL 8.4 release notes here: [MySQL 8.4 Release Notes](https://dev.mysql.com/doc/refman/8.4/en/mysql-nutshell.html#mysql-nutshell-removals).

Thank you for reviewing this PR!

### DIFF
--- a/collector/slave_hosts.go
+++ b/collector/slave_hosts.go
@@ -31,12 +31,12 @@ const (
 	// timestamps. %s will be replaced by the database and table name.
 	// The second column allows gets the server timestamp at the exact same
 	// time the query is run.
-	slaveHostsQuery = "SHOW SLAVE HOSTS"
 )
 
 // Metric descriptors.
 var (
-	SlaveHostsInfo = prometheus.NewDesc(
+	slaveHostsQueries = [2]string{"SHOW SLAVE HOSTS", "SHOW REPLICAS"}
+	SlaveHostsInfo    = prometheus.NewDesc(
 		prometheus.BuildFQName(namespace, heartbeat, "mysql_slave_hosts_info"),
 		"Information about running slaves",
 		[]string{"server_id", "slave_host", "port", "master_id", "slave_uuid"}, nil,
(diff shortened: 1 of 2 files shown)


collector/slave_hosts.go
```diff
@@ -63,7 +63,17 @@ func (ScrapeSlaveHosts) Version() float64 {
 
 // Scrape collects data from database connection and sends it over channel as prometheus metric.
 func (ScrapeSlaveHosts) Scrape(ctx context.Context, db *sql.DB, ch chan<- prometheus.Metric, logger log.Logger) error {
-	slaveHostsRows, err := db.QueryContext(ctx, slaveHostsQuery)
+	var (
+		slaveHostsRows *sql.Rows
+		err            error
+	)
+	// Try the both syntax for MySQL 8.0 and MySQL 8.4
+	for _, query := range slaveHostsQueries {
```
SuperQ, line 71: I'm not sure I like this for loop/break pattern. I would prefer to do a simple linear `if` drop through.

collector/slave_hosts.go
```diff
@@ -31,12 +31,12 @@ const (
 	// timestamps. %s will be replaced by the database and table name.
 	// The second column allows gets the server timestamp at the exact same
 	// time the query is run.
-	slaveHostsQuery = "SHOW SLAVE HOSTS"
 )
 
 // Metric descriptors.
 var (
-	SlaveHostsInfo = prometheus.NewDesc(
+	slaveHostsQueries = [2]string{"SHOW SLAVE HOSTS", "SHOW REPLICAS"}
```
SuperQ, line 38: I would rather keep the const and add a new `showReplicasQuery` const for the new version.

SuperQ (changes requested):

mtanda:
@SuperQ Thank you for the review! I have addressed the comments and made the necessary changes. Could you please review it again?

SuperQ (approved):
Thanks!

SuperQ:
I suppose we should add MySQL 8.4 to our CI pipeline.

mtanda:
Thanks for merge!!

happywzy:
Which version of docker solves this problem?

Issue #790, opened 2023-12-07
Docker: my.cnf => 'No such file or directory'


SK-StYleZ:
### Host operating system: output of `uname -a`
Linux home 6.1.21-v8+ #1642 SMP PREEMPT Mon Apr  3 17:24:16 BST 2023 aarch64 GNU/Linux

### mysqld_exporter version: output of `mysqld_exporter --version`
latest

### MySQL server version
latest

### mysqld_exporter command line flags
docker run -d \
  -p 9104:9104 \
  prom/mysqld-exporter:latest
  --config.my-cnf=./my.cnf
d7c0d75852982e7182b260df7569be11b71bcde238b99f2e6a7f8d8031fa8cdc
-bash: --config.my-cnf=./my.cnf: No such file or directory

the "my.cnf" is in the current working directory
the format of "my.cnf" is the same as in the documentation:
`
    [client]
    user = foo
    password = foo123
    [client.servers]
    user = bar
    password = bar123`

### What did you expect to see?
running exporter

### What did you see instead?
an error

can someone please provide an working docker // docker-compose.yaml example?

SK-StYleZ:
please update the docker documentaion, it's incorrect ...

kavishan-thathsara:
Having same issue. How did this fix?

imamhadid:
https://github.com/prometheus/mysqld_exporter/blob/main/README.md#single-exporter-mode
Running using ```.my.cnf``` from the current directory


try using ```.my.cnf``` not ```my.cnf```

SK-StYleZ:
Thanks  for the respond.
I'm using  mysqld_exporter **with docker**.
So here is the updated command promt:
```
docker run -d
-p 9104:9104
prom/mysqld-exporter:latest
--config.my-cnf=.my.cnf
"docker run" requires at least 1 argument.
See 'docker run --help'.

Usage:  docker run [OPTIONS] IMAGE [COMMAND] [ARG...]

Create and run a new container from an image
-bash: -p: command not found
-bash: prom/mysqld-exporter:latest: No such file or directory
-bash: --config.my-cnf=.my.cnf: command not found
```

This is the official documentation for docker: https://github.com/prometheus/mysqld_exporter/blob/main/README.md#using-docker

I've tried this:
```
docker run -d \
  -p 9104:9104 \
  --network my-mysql-network  \
  prom/mysqld-exporter
  --config.my-cnf=./my.cnf
[...]

PR #843, opened 2024-05-30
Fix Docker run command and update documentation for cnf file handling

jcy0308:
@SuperQ 

When running docker run as currently specified in the README, it returns -bash: --config.my-cnf=./my.cnf: No such file or directory. If you try using .my.cnf as suggested by someone else in the issue #790 , it returns -bash: --config.my-cnf=.my.cnf: command not found.

It would be better to revise the documentation to have users create the cnf file themselves and use volume mounting instead of using an option.

The updated command is as follows:
```bash
docker run -d \
  -p 9104:9104 \
  -v /home/user/user_my.cnf:/.my.cnf \
  --network my-mysql-network \
  prom/mysqld-exporter
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,9 +196,9 @@ docker pull prom/mysqld-exporter
 
 docker run -d \
   -p 9104:9104 \
+  -v /home/user/user_my.cnf:/.my.cnf \
   --network my-mysql-network  \
   prom/mysqld-exporter
-  --config.my-cnf=<path_to_cnf>
 ```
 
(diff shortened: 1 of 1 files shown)


SuperQ:
I don't understand the "command not found" error. I am unable to reproduce this. The `ENTRYPOINT` is the exporter, not bash. So there must be something wrong with the environment in the linked issue.

However, I think this is a good documentation update since the volume mount documentation is good idea.

PR #859, opened 2024-08-12
Add the instance struct to handle connections

s10:
The intent is to use the instance struct to hold the connection to the database as well as metadata about the instance:
- version
- flavor (mariadb or mysql)

Change is similar to prometheus-community/postgres_exporter#785

### DIFF
--- a/collector/instance.go
+++ b/collector/instance.go
@@ -1,0 +1,114 @@
+// Copyright 2024 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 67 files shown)


collector/instance.go
```diff
@@ -0,0 +1,114 @@
...
+	"regexp"
+	"strconv"
+	"strings"
+
+	"github.com/blang/semver/v4"
+)
+
+const (
+	MYSQL   = "mysql"
+	MARIADB = "mariadb"
```
SuperQ, line 28: Let's go with the upstream naming.

```suggestion
	MySQL   = "mysql"
	MariaDB = "mariadb"
```

SuperQ (changes requested):

SuperQ (approved):

PR #862, opened 2024-08-26
Fetch lock time and cpu time from performance schema

cristiangreco:
Extend the query on `events_statements_summary_by_digest` to fetch `SUM_LOCK_TIME` and `SUM_CPU_TIME` and return them as counters. These are useful to report time spent running on the cpu (e.g. to sort results or run computations) vs waiting on locks (e.g. to acquire write lock on a table or row), by query digest.

### DIFF
--- a/collector/perf_schema_events_statements.go
+++ b/collector/perf_schema_events_statements.go
@@ -31,6 +31,8 @@ const perfEventsStatementsQuery = `
 	    LEFT(DIGEST_TEXT, %d) as DIGEST_TEXT,
 	    COUNT_STAR,
 	    SUM_TIMER_WAIT,
+	    SUM_LOCK_TIME,
+	    SUM_CPU_TIME,
 	    SUM_ERRORS,
 	    SUM_WARNINGS,
 	    SUM_ROWS_AFFECTED,
@@ -54,6 +56,8 @@ const perfEventsStatementsQuery = `
 	    Q.DIGEST_TEXT,
 	    Q.COUNT_STAR,
 	    Q.SUM_TIMER_WAIT,
+	    Q.SUM_LOCK_TIME,
+	    Q.SUM_CPU_TIME,
 	    Q.SUM_ERRORS,
 	    Q.SUM_WARNINGS,
 	    Q.SUM_ROWS_AFFECTED,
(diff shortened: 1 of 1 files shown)


SuperQ (commented):
I think we originally intentionally left these out for cardinality reasons. Would you mind documenting how they they are additionally useful in the PR description? That way we can have better history of "why" these are added by default.

cristiangreco:
> I think we originally intentionally left these out for cardinality reasons. Would you mind documenting how they they are additionally useful in the PR description? That way we can have better history of "why" these are added by default.

Hey @SuperQ thanks for having a look! These metrics would have the same cardinality (in terms of labels values) as the other sibling metrics created out of the query `events_statements_summary_by_digest`. Or are we worried about adding new metrics in general? (I wanted to submit a followup PR to additionally expose `QUANTILE_{95,99,999}` so it'd be good to understand what is the approach to take)

SuperQ:
It's about adding new metrics in general.

Exposing the quantiles is less useful, as they can't be aggregated between MySQL servers.

cristiangreco:
I've added a short note on the PR description.

SuperQ:
Needs a rebase.

cristiangreco:
> Needs a rebase.

Thanks so much! Just rebased.

SuperQ:
I had a think about the history of this. I wish we had written down the queries we experimented with in production at the time.

I think we tried adding CPU time originally, but we didn't really find it useful compared to total query time.

Lock time is probably a good one to have tho.

cristiangreco:
> I think we tried adding CPU time originally, but we didn't really find it useful compared to total query time.

@SuperQ fair enough, if you prefer I can close this PR and open a new one adding just lock time?

SuperQ:
I'm mostly looking for up-to-date expertise in MySQL here. It's been a while since I looked deeply at this data. I assume you've got data on the uselessness of these metrics.

cristiangreco:
> I'm mostly looking for up-to-date expertise in MySQL here. It's been a while since I looked deeply at this data. I assume you've got data on the uselessness of these metrics.

So I'm not sure that `cpu time = total time - lock time` is true but can't disprove it either. I thought it was cheap enough that we could export both metrics anyway.

SuperQ:
Ok, I guess this is fine as-is. If people complain about the additional cardinality we can look it over again.

dbakit:
https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-28.html
----
As indicated in the release notes, `SUM_LOCK_TIME` and `SUM_CPU_TIME` are only supported starting from version 8.0.28. These fields do not exist in versions 5.7 and below or in versions prior to 8.0.28. It is recommended to differentiate the scrap SQL based on the `mysql_version`.

PR #871, opened 2024-09-10
Add new collector

silverlee425:
Collect the duration of the longest-running long transaction.

### DIFF
--- a/collector/info_schema_innodb_trx_long.go
+++ b/collector/info_schema_innodb_trx_long.go
@@ -1,0 +1,105 @@
+// Copyright 2018 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Scrape `information_schema.innodb_trx`.
+
+package collector
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/alecthomas/kingpin/v2"
+	"github.com/go-kit/log"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const infoSchemaInnodbTrxLongQuery = `
+		  SELECT (unix_timestamp(now()) - unix_timestamp(trx_started)) AS trx_seconds
+			  FROM information_schema.innodb_trx tx
+		   WHERE tx.trx_state IN ('RUNNING','LOCK WAIT','ROLLING BACK','COMMITTING')
+  			 AND (unix_timestamp(now()) - unix_timestamp(tx.trx_started)) > %d
+		   ORDER BY trx_seconds DESC
+		   LIMIT 1
+	`
+
+// Tunable flags.
+var (
+	trxMinTime = kingpin.Flag(
+		"collect.info_schema.innodb_trx.min_seconds",
+		"Minimum time a thread must be in each state to be counted",
+	).Default("0").Int()
+)
+
+// Metric descriptors.
+var (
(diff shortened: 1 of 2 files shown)


collector/info_schema_innodb_trx_long.go
```diff
@@ -0,0 +1,105 @@
...
+// Scrape `information_schema.innodb_trx`.
+
+package collector
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/alecthomas/kingpin/v2"
+	"github.com/go-kit/log"
```
ArthurSens, line 23: we're using a new logging library now 😬

collector/info_schema_innodb_trx_long.go
```diff
@@ -0,0 +1,105 @@
...
+		"Minimum time a thread must be in each state to be counted",
+	).Default("0").Int()
+)
+
+// Metric descriptors.
+var (
+	trxTimeDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, informationSchema, "innodb_trx_long"),
+		"Execution threshold time of long transactions",
+		nil, nil)
```
ArthurSens: wouldn't it be useful to keep the `state` as a label?

collector/info_schema_innodb_trx_long.go
```diff
@@ -0,0 +1,105 @@
...
+
+	"github.com/alecthomas/kingpin/v2"
+	"github.com/go-kit/log"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const infoSchemaInnodbTrxLongQuery = `
+		  SELECT (unix_timestamp(now()) - unix_timestamp(trx_started)) AS trx_seconds
+			  FROM information_schema.innodb_trx tx
+		   WHERE tx.trx_state IN ('RUNNING','LOCK WAIT','ROLLING BACK','COMMITTING')
```
ArthurSens, line 30: this WHERE isn't doing anything, since we're listing all possible states, right? Couldn't we drop it?

collector/info_schema_innodb_trx_long.go
```diff
@@ -0,0 +1,105 @@
...
+  			 AND (unix_timestamp(now()) - unix_timestamp(tx.trx_started)) > %d
+		   ORDER BY trx_seconds DESC
+		   LIMIT 1
```
ArthurSens, line 33:
```suggestion
	SELECT
		trx_state,
		COUNT(*) AS transactions,
		COALESCE(MAX(TIMESTAMPDIFF(SECOND, trx_started, NOW())), 0) AS oldest_transaction_seconds
	FROM information_schema.innodb_trx
	GROUP BY trx_state
```

I think a query like this could help us keep the oldest active transaction while also having the number of transactions per state. We can drop the configured age filter, since PromQL is better for choosing alerting thresholds than for collection-side logic, like in exporters.

collector/info_schema_innodb_trx_long.go
```diff
@@ -0,0 +1,105 @@
...
+	`
+
+// Tunable flags.
+var (
+	trxMinTime = kingpin.Flag(
+		"collect.info_schema.innodb_trx.min_seconds",
+		"Minimum time a thread must be in each state to be counted",
+	).Default("0").Int()
+)
```
ArthurSens, line 42: We can probably drop this, let's always capture the oldest active transaction, and let PromQL queries decide if something is old enough to trigger an alert

collector/info_schema_innodb_trx_long.go
```diff
@@ -0,0 +1,105 @@
...
+
+// Metric descriptors.
+var (
+	trxTimeDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, informationSchema, "innodb_trx_long"),
+		"Execution threshold time of long transactions",
+		nil, nil)
+)
```
ArthurSens, line 50: Thinking of how the transaction age can be used in monitoring use cases, let's rethink this metric to expose age of oldest active transaction.

PromQL can decide thresholds better than mysql exporter

ArthurSens (commented):
Hi, sorry for the very delayed review. I've put some comments below!

We also need a rebase, and it would be awesome if the new code is properly tested :)

PR #873, opened 2024-09-19
Bump exporter-toolkit and fix associated test

dswarbrick:
Bump exporter-toolkit dependency to v0.12.0 and fix a (IMHO overly brittle) test that breaks due to a cosmetic change in the landing page.

And yes, I am aware that exporter-toolkit v0.13.0 is available, but that introduces a backwards-incompatible change that will require more extensive changes (go-kit/log -> log/slog migration), best saved for a separate PR.


SuperQ (approved):
Thanks. Yes, I will take care of the slog change soon.

PR #879, opened 2024-10-02
info_schema_tables: do not collect the `sys` schema

frivoire:
This `sys` schema is a mariadb-specific "system" schema, introduced in MariaDB 10.6
cf: https://mariadb.com/kb/en/sys-schema/

And it's not useful to collect metrics for it (it's even counter-productive, cf cost of storing the metrics afterwards)
So, let's remove this schema, just like the other "system" schemas (ex: performance_schema) are already excluded today

Note:
It's my first PR here, so I'm not sure about the CI/tests

### DIFF
--- a/collector/info_schema_tables.go
+++ b/collector/info_schema_tables.go
@@ -46,7 +46,7 @@ const (
 		SELECT
 		    SCHEMA_NAME
 		  FROM information_schema.schemata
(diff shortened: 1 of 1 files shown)


SuperQ (approved):
Agreed, thanks, if we do want to collect anything from sys schema, we can add a separate collector function.

frivoire:
Thanks !!

Issue #883, opened 2024-10-15
Add feature to get MySQL password from Vault, instead of config file.


Dmitry-Eremeev:
It would be nice to implement security feature to get some credentials from Vault, e.g. mysql password.

PR #875, opened 2024-09-29
chore!: adopt log/slog, drop go-kit/log

tjhop:
Requires: prometheus/common#697

This PR includes:
- linter updates to enable `sloglint` linter
- Go dep updates for prometheus/{client_golang,common,exporter-toolkit} libs
- refactorings to adopt log/slog in favor of go-kit/log

The bulk of this PR was automated by the following script which is being used to aid in converting the various exporters/projects to use slog:

https://gist.github.com/tjhop/49f96fb7ebbe55b12deee0b0312d8434

Builds and passes tests locally with go workspaces and up-to-date main branch of prometheus/common.

PR #885, opened 2024-10-15
Added feature to get password from Vault

Dmitry-Eremeev:
To use the feature we need to set the following options in config section 'client':

use_vault = true
vault_address = http://<IP_ADDRESS:<IP_PORT>
vault_role_id = <ROLE_ID>
vault_secret_id = <SECRET_ID>
vault_secret_path = <deployments/<UNIT>/dev/user/passwords_yml> vault_secret_mount_path = <secret_v2>
credential_name_in_vault_secret = <prometheus_mysql_exporter_database_password>

https://github.com/prometheus/mysqld_exporter/issues/883


SuperQ (commented):
Sorry, but I don't think we want to support this. There are far too many secrets managers out there, the code would be unmaintainable.

Dmitry-Eremeev:
@cristiangreco 
You put my PR in RN:
* [CHANGE] Replace logging library go-kit/log with slog https://github.com/prometheus/mysqld_exporter/pull/885

I believe it's typo.

cristiangreco:
@Dmitry-Eremeev yes sorry, I initially referenced #885 instead of #875, but then I fixed it.

PR #896, opened 2024-11-21
Add command line option to explicitly disable lock_wait_timeout (adds possibility to opt out of breaking changes with older versions)

festeveira:
This PR adds a command line option to the exporter allowing the user to disable the lock_wait_timeout MySQL connection parameter making the exporter work again with older versions of mysql, while still behaving the same way by default.

Was really helpful for me, I need to monitor several instances in multi target mode and most of them are unsupported versions. Unless you see a problem with this I think it would be nice to have it merged.


SuperQ:
I'm not generally a fan of "magic values" like this. I prefer more explicit settings.

festeveira:
Hello, thank you for your feedback! I have added a command line parameter to explicitly disable the option. What do you think?

README.md
```diff
@@ -145,6 +145,7 @@ mysqld.username                            | Username to be used for connecting
 config.my-cnf                              | Path to .my.cnf file to read MySQL credentials from. (default: `~/.my.cnf`)
 log.level                                  | Logging verbosity (default: info)
 exporter.lock_wait_timeout                 | Set a lock_wait_timeout (in seconds) on the connection to avoid long metadata locking. (default: 2)
+exporter.disable_lock_wait_timeout         | Disable the lock_wait_timeout connection parameter. Makes the exporter compatible with older versions of MySQL. (default: false)
```
SuperQ, line 148: I prefer "positive" flags, so that we don't have double negative behavior.

The disable would be `--no-exporter.enable_lock_wait_timeout`.

```suggestion
exporter.enable_lock_wait_timeout         | Disable the lock_wait_timeout connection parameter. Makes the exporter compatible with older versions of MySQL. (default: true)
```

SuperQ (changes requested):

SuperQ:
Yes, as a boolean flag is better. Thanks.

SuperQ (approved):
Thanks!

PR #908, opened 2025-01-16
Fix database quoting problem in collector 'info_schema.tables'

oblitorum:
In `collector/info_schema_tables.go`, the SQL it uses to retrieves the table meta data is

```
SELECT
	TABLE_SCHEMA,
	...
FROM information_schema.tables
WHERE TABLE_SCHEMA = '%s'
```

Which could cause a problem if the schema contains the single quote character `'` (It's indeed a vaild character in database/table name), we should use the placeholder `?` and pass the schema as an argument to `db.QueryContext(...)` to avoid this kind of problem. Also to avoid the SQL injection problem as user can pass anything to `--collect.info_schema.tables.databases` in reality (accidentally or on purpose).


cristiangreco (approved):
Thanks! I've just encountered something similar myself.

PR #909, opened 2025-01-18
Creating DSN using database name.

BeniaminK:
This patch introduces the ability to specify a database name in the Data Source Name (DSN) configuration for the MySQL exporter.

Previously, the exporter would connect to the default database if no database name was provided.

With this change, users can define a specific database name in the configuration.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -61,6 +61,7 @@ type MySqlConfig struct {
 	User                  string `ini:"user"`
 	Password              string `ini:"password"`
 	Host                  string `ini:"host"`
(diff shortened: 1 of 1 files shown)


ArthurSens (changes requested):
Hey, sorry for the super delayed review. It's a good change to enable configuring a default database, but as it is the chosen database won't filter all metric collectors to chose the configured database.

This information needs to be passed down to metric collectors so the filter applies everywhere.

It would also be awsome to have tests around the behavior we want to introduce here :)

PR #911, opened 2025-01-28
Update Mixin dashboard to editable: false

grooverdan:
This was reported by the mixtool lint:
  mixtool lint mixin.libsonnet
  [uneditable-dashboard] 'MySQL': Dashboard 'MySQL' is editable, it should be set to 'editable: false'

This is quick attempt at fixing the mixin CI error going by its error description. I haven't looked further on the implications of this.

### DIFF
--- a/mysqld-mixin/dashboards/mysql-overview.json
+++ b/mysqld-mixin/dashboards/mysql-overview.json
@@ -13,7 +13,7 @@
     ]
   },
   "description": "",
-  "editable": true,
+  "editable": false,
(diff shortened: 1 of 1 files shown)


rgeyer (approved):
LGTM. This is mostly a cosmetic/continuity rule. The result is that when it is imported into Grafana, the dashboard will be read-only by default.

It can, of course, be overwritten by a user with adequate permissions.

cristiangreco (approved):

PR #917, opened 2025-02-18
Prepare release 0.17.0

cristiangreco:
Update changelog for 0.17.0


CHANGELOG.md
```diff
@@ -9,6 +9,15 @@ Changes:
 * [ENHANCEMENT]
 * [BUGFIX]
 
+## 0.16.1 / 2025-02-19
```
SuperQ, line 12: Patch releases should be bugfix-only.

Let's release v0.17.0.

dswarbrick:
How about pulling in the updated exporter-toolkit v0.14.0? Note that it will require tweaks to the landing page test, similar to #873.

cristiangreco:
> How about pulling in the updated exporter-toolkit v0.14.0? Note that it will require tweaks to the landing page test, similar to #873.

Hey @dswarbrick, would you have time to prepare a PR for that? 🙏

cristiangreco:
@SuperQ should we go ahead with the release if #919 is going to wait some more time?

SuperQ (changes requested):
We also need to update the VERSION file.

PR #964, opened 2025-07-16
Add gtid_transaction counter taken from @gtid_executed

sjmudd:
This PR add a global transaction counter for MySQL when using GTIDs which is basically the default for any modern version of MySQL.
- This should work from MySQL 5.6 when GTIDs were added
- A single transaction counter is generated from the GTID executed value
- gaps in transactions and also the new 8.4+ tagged GTIDs format is handled.
- The metric is exposed as `mysql_gtid_transactions` and is most useful when checking rate of change e.g. transactions per second.

I am not aware of any other transaction counter provided by MySQL explicitly.


sjmudd:
I wasn't aware of sign off requirements. I'll drop this PR and redo it after signing off next time.

PR #965, opened 2025-07-16
Add a gtid transaction counter based on the @@gtid_executed global variable

sjmudd:
This PR provides a counter of GTID executions derived from the `@@gtid_executed` global variable. Strictly speaking this counts GTID events rather than MySQL transactions, but in practical terms the two are equivalent, and this can be used to track the transaction rate on a MySQL server at least for monitoring purposes.

- The full GTID set is combined into a single execution count
- UUIDs are ignored
- Any GTID gaps are taken into account
- works on MySQL 5.6+
- works with MySQL 8.4+ tagged gtid format


sjmudd:
Hello,

It is not very clear to me how long I should expect to wait for a response to my PR, whether positive or negative. If you could provide some sort of feedback here or in the [README](https://github.com/prometheus/mysqld_exporter/blob/main/README.md) that would be most helpful.

sjmudd:
2 months have passed. Any chance of getting this PR accepted? If you need something more from me can you tell me what it is?

collector/mysql_gtid.go
```diff
@@ -0,0 +1,87 @@
...
+	prometheusName      = "transactions"
+	// gtidTransactionCountQuery is the query used to fetch gtid_executed.
+	// With this value we can convert it to an incremental transaction counter.
+	gtidTransactionCountQuery = "SELECT @@gtid_executed"
+)
+
+var (
+	// Metric descriptors.
+	GtidTransactionCounterDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, prometheusSubsystem, prometheusName),
```
ArthurSens, line 38:
```suggestion
		prometheus.BuildFQName(namespace, prometheusSubsystem,"executed_total"),
```

Following [Prometheus naming conventions](https://prometheus.io/docs/practices/naming/), counters should be suffixed with `_total`. Besides that, `gtid_executed` seems more accurate than `gtid_transactions` 🤔

collector/mysql_gtid_test.go
```diff
@@ -0,0 +1,90 @@
...
+		name     string
+		gtidSet  string
+		expected float64
+	}{
+		{"empty_set", "", 0},
+		{"single_uuid_and_range", `uuid1:1-1000`, 1000},
+		{"multiple_uuid_single_range", `uuid1:1-1000,
+	uuid1:1001-2000`, 2000},
+		{"single_uuid_with_ranges", `uuid1:1-1000,2001-4000`, 3000},
+	}
```
ArthurSens, line 40: Looking at docs

- [GTID format, sets, intervals, and tags](https://dev.mysql.com/doc/refman/8.4/en/replication-gtids-concepts.html)
- [gtid_executed system-variable semantics](https://dev.mysql.com/doc/refman/8.4/en/replication-options-gtids.html)
- [GTID lifecycle, gaps, filtering, and resets](https://dev.mysql.com/doc/refman/8.4/en/replication-gtids-lifecycle.html)

The test cases seems incorrect 🤔 

I'd add something like


```suggestion
	tests := []struct {
		name      string
		gtidSet   string
		expected  float64
		expectErr bool
	}{
		{"empty_set", "", 0, false},
		{"single_transaction", `3E11FA47-71CA-11E1-9E33-C80AA9429562:1`, 1, false},
		{"single_uuid_and_range", `3E11FA47-71CA-11E1-9E33-C80AA9429562:1-1000`, 1000, false},
		{"single_uuid_with_ranges", `3E11FA47-71CA-11E1-9E33-C80AA9429562:1-3:11:47-49`, 7, false},
		{"multiple_uuids", `3E11FA47-71CA-11E1-9E33-C80AA9429562:1-3,
24BC7856-9C4A-11E1-9D41-80C16E4A511C:1-10`, 13, false},
		{"multiple_uuids_single_line", `3E11FA47-71CA-11E1-9E33-C80AA9429562:1-3,24BC7856-9C4A-11E1-9D41-80C16E4A511C:1-10`, 13, false},
		{"tagged_gtid_with_ranges", `3E11FA47-71CA-11E1-9E33-C80AA9429562:Domain_1:1-3:11:47-49`, 7, false},
		{"multiple_tags_for_same_uuid", `3E11FA47-71CA-11E1-9E33-C80AA9429562:Domain_1:1-3:15-21,
3E11FA47-71CA-11E1-9E33-C80AA9429562:Domain_2:8-52`, 55, false},
		{"mixed_tagged_and_untagged_sets", `3E11FA47-71CA-11E1-9E33-C80AA9429562:1-5,
24BC7856-9C4A-11E1-9D41-80C16E4A511C:Analytics_1:10-14`, 10, false},
		{"missing_uuid", `1-10`, 0, true},
		{"comma_between_intervals", `3E11FA47-71CA-11E1-9E33-C80AA9429562:1-3,11-20`, 0, true},
		{"reversed_range", `3E11FA47-71CA-11E1-9E33-C80AA9429562:20-10`, 0, true},
		{"zero_transaction_id", `3E11FA47-71CA-11E1-9E33-C80AA9429562:0`, 0, true},
		{"missing_tagged_interval", `3E11FA47-71CA-11E1-9E33-C80AA9429562:Domain_1`, 0, true},
		{"invalid_tag", `3E11FA47-71CA-11E1-9E33-C80AA9429562:123tag:1-10`, 0, true},
		{"transaction_id_overflow", `3E11FA47-71CA-11E1-9E33-C80AA9429562:1-9223372036854775808`, 0, true},
	}```

File: go.mod [hunk 3]
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/prometheus/client_model v0.6.2
 	github.com/prometheus/common v0.65.0
 	github.com/prometheus/exporter-toolkit v0.13.2
+	github.com/sjmudd/mysqlgtid v0.1.0
```
ArthurSens, line 16: This dependency seems to parse GTID incorrectly; is there a more accurate one? It's been a few years since the PR was opened, so maybe there are newer, more accurate versions?

collector/mysql_gtid.go
```diff
@@ -0,0 +1,87 @@
...
+		[]string{}, nil,
+	)
+)
+
+// ScrapeGtidExecuted scrapes transaction count from @@gtid_executed.
+type ScrapeGtidExecuted struct{}
+
+// Name of the Scraper. Should be unique.
+func (ScrapeGtidExecuted) Name() string {
+	return "gtid_transactions"
```
ArthurSens, line 49:
```suggestion
	return "gtid_executions"
```

collector/mysql_gtid.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// Help describes the role of the Scraper.
+func (ScrapeGtidExecuted) Help() string {
+	return "Number of GTID transactions"
```
ArthurSens, line 54:
```suggestion
	return "Number of GTID executions"
```

ArthurSens (commented):
Hey, sorry for the super-delayed review. We are understaffed for exporter maintainers, but we're trying to catch up with our backlog 😓 

I've left a few comments below. Also, we might want to rewrite the code to clarify that it's about GTID executions, not transactions, since people might get confused with regular database transactions. For example, `SET gtid_purged` is not a transaction at all, but counts as a GTID execution

collector/mysql_gtid.go
sjmudd: ok, this is picking hairs but if you need it to continue I'm fine with that.  It may be true that the absolutely exact definition of a transaction and the GTID changes is not 1:1 but the intention of this "metric" is to provide you a reference transactions and thus transactions per second counter which is not available by any other mains. The noise in the difference there on anything except a toy server is so small I'd say it's 0.

However, you are right to point out this difference. I assumed most people would be aware, but some may not recognise the difference.

I can not apply the suggested change through the github CI as for some reason github shows it with my former employee's email address and it would be wrong to sign this off as I no longer work there. The contribution was from my personal account and done at a completely personal level. So I'll apply the change in my branch and push which I think should solve this.

go.mod
sjmudd: I will double check as you are correct that this PR has been sitting idle for quite some time.

README.md
```diff
@@ -113,6 +113,7 @@ collect.engine_innodb_status                                 | 5.1           | C
 collect.engine_tokudb_status                                 | 5.6           | Collect from SHOW ENGINE TOKUDB STATUS.
 collect.global_status                                        | 5.1           | Collect from SHOW GLOBAL STATUS (Enabled by default)
 collect.global_variables                                     | 5.1           | Collect from SHOW GLOBAL VARIABLES (Enabled by default)
+collect.gtid_executions                                      | 5.6           | Collect a transaction count from @gtid_executed.
```
ArthurSens, line 116:
```suggestion
collect.gtid_executions                                      | 5.6           | Collect a execution count from @gtid_executed.
```

collector/mysql_gtid.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/sjmudd/mysqlgtid"
+)
+
+const (
+	// prometheusSubsystem is the Metric subsystem we use.
+	prometheusSubsystem = "gtid"
+	prometheusName      = "executed_total"
+	// gtidTransactionCountQuery is the query used to fetch gtid_executed.
+	// With this value we can convert it to an incremental transaction counter.
+	gtidTransactionCountQuery = "SELECT @@gtid_executed"
```
ArthurSens, line 32:
```suggestion
	// gtidExecutionCountQuery is the query used to fetch gtid_executed.
	// With this value we can convert it to an incremental execution counter.
	gtidExecutionCountQuery = "SELECT @@gtid_executed"
```

collector/mysql_gtid.go
```diff
@@ -0,0 +1,88 @@
+// Copyright 2026 The Prometheus Authors
```
ArthurSens, line 1:
```suggestion
// Copyright The Prometheus Authors
```

CI is failing because of this

collector/mysql_gtid_test.go
```diff
@@ -0,0 +1,109 @@
+// Copyright 2026 The Prometheus Authors
```
ArthurSens, line 1:
```suggestion
// Copyright The Prometheus Authors
```
ditto

go.mod
```diff
@@ -1,18 +1,19 @@
...
 	github.com/blang/semver/v4 v4.0.0
-	github.com/go-sql-driver/mysql v1.10.0
+	github.com/go-sql-driver/mysql v1.10.1
 	github.com/google/go-cmp v0.7.0
 	github.com/google/uuid v1.6.0
 	github.com/prometheus/client_golang v1.24.1
-	github.com/prometheus/client_model v0.6.2
-	github.com/prometheus/common v0.70.1
+	github.com/prometheus/client_model v0.6.3
+	github.com/prometheus/common v0.71.0
```
ArthurSens, line 14: are these changes needed for this PR?

collector/mysql_gtid.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// Version of MySQL from which scraper is available.
+func (ScrapeGtidExecuted) Version() float64 {
+	return 5.6
+}
+
+// Scrape collects data from database connection and sends it over channel as prometheus metric.
+func (ScrapeGtidExecuted) Scrape(ctx context.Context, instance *instance, ch chan<- prometheus.Metric, logger *slog.Logger) error {
+	var gtidExecuted string
```
ArthurSens, line 65: let's add a guard to not run this on MariaDB, something similar to

```go
if instance.flavor != FlavorMySQL {
		return nil
	}
```

If we don't do that, this scraper will return errors on every scrape. A correspondind test would also be nice :)

ArthurSens (commented):
Thanks for picking this up even after such a long time!

We're very close to a mergable state, there are some CI failures and I found a couple extra things that might need fixing :)

PR #967, opened 2025-07-23
refactor: move exporter flags to main package

cristiangreco:
This refactoring makes flags handling more similar to postgres_exporter, moving `exporter.` flags to the main package and using an "opts" pattern for building the Exporter instance.


SuperQ (approved):
Thanks!

Issue #749, opened 2023-06-28
Optionally use IAM authentication


mcanevet:
It would be great to be able to use IAM authentication when the exporter runs on AWS: https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/UsingWithRDS.IAMDBAuth.Connecting.Go.html

tejavegesna-lb:
Agree with @mcanevet We need a way to use IAM Authentication for metrics

lasith-kg:
Agree! Support for dynamic credentials (of the likes of RDS IAM Auth) would be great from a security posture

PR #983, opened 2025-10-17
Add IAM Auth support

bvalente:
Addressing https://github.com/prometheus/mysqld_exporter/issues/749

We're adding the option to authenticate to AWS RDS with IAM, as described in this guide: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.Connecting.Go.html

To test, I used the following config file (with redacted values):
```ini
[client]
user=my-user
host=mydatabase.eu-central-1.rds.amazonaws.com
port=3306
aws-iam-auth=true
aws-region=eu-central-1
enable-cleartext-plugin=true
ssl-ca=global-bundle.pem
```

@SuperQ 🙂


ArthurSens (commented):
Hey, sorry for the delayed review here. I believe this is a good feature, but we might need to coordinate this a bit better with the rest of the community!

I'd love to have a consistent implementation with what @matt-gp is working on for https://github.com/prometheus-community/postgres_exporter/issues/1386.

Let's discuss this as a group in the issue above and then move forward, this way we make sure we have similar experiences throughout all our exporters that could benefit from this feature :)

PR #985, opened 2025-10-17
sys.user_summary: Fix handling of negative current_memory values

maleki-it:
MySQL's sys.user_summary.current_memory column can contain negative values
when memory is freed faster than allocated. This is valid MySQL behavior.

Changes:
- Changed current_memory type from uint64 to int64
- Added test case with negative current_memory value (-16360)

Fixes scan error: 'converting driver.Value type []uint8 to uint64: invalid syntax'

Signed-off-by: Ali Maleki <alimaleki.wit@gmail.com>


SuperQ:
If you are going to use LLMs to write PRs, please include this information in the PR description.

PR #1005, opened 2026-02-08
fix authentication failure when the password contains the $ symbol

morawskim:
This pr should resolve this issue https://github.com/prometheus/mysqld_exporter/issues/1004

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -144,7 +144,14 @@ func (ch *MySqlConfigHandler) ReloadConfig(filename string, mysqldAddress string
 		}
 	}
 
-	cfg.ValueMapper = os.ExpandEnv
+	cfg.ValueMapper = func(val string) string {
+		return os.Expand(val, func(name string) string {
+			if name == "$" {
+				return "$"
+			}
+			return os.Getenv(name)
+		})
+	}
 	config := &Config{}
 	m := make(map[string]MySqlConfig)
 	for _, sec := range cfg.Sections() {
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ This exporter supports the multi-target pattern. This allows running a single in
 To use the multi-target functionality, send an http request to the endpoint `/probe?target=foo:3306` where target is set to the DSN of the MySQL instance to scrape metrics from.
 
 To avoid putting sensitive information like username and password in the URL, you can have multiple configurations in `config.my-cnf` file and match it by adding `&auth_module=<section>` to the request.
- 
+
 Sample config file for multiple configurations
 
         [client]
@@ -68,7 +68,13 @@ Sample config file for multiple configurations
         password = foo123
         [client.servers]
         user = bar
-        password = bar123
+        password = ${PASSWORD_FROM_ENV}
+        [client.servers2]
+        user = baz
+        password = MY_$$ECRET
+
+If your password contains a dollar sign ($) and you don't want it to be interpreted as an environment variable,
+escape it by doubling it ($$). In the example above, the password for `client.servers2` is `MY_$ECRET`.
 
 On the prometheus side you can set a scrape config as follows
 
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -214,6 +214,57 @@ func TestValidateConfig(t *testing.T) {
 			fmt.Errorf("tls-max-version=TLSv-something is not allowed, use one of: TLSv1.0, TLSv1.1, TLSv1.2, TLSv1.3"),
 		)
 	})
+
+	convey.Convey("Expand variables", t, func() {
+		c := MySqlConfigHandler{
+			Config: &Config{},
+		}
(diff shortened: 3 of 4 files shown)


config/config.go
```diff
@@ -123,7 +123,22 @@ func (ch *MySqlConfigHandler) ReloadConfig(filename string, mysqldAddress string
...
+			i++
+		}
+
+		if i == 0 {
+			return val
+		}
+
+		if i%2 == 0 {
+			return strings.Repeat("$", i/2) + val[i:]
+		}
```
ArthurSens: This doesn't sound like a good heuristic for escaping dollar signs; ${}, $$${}, and $$$$${} would all be the same thing after parsing 🤔

ArthurSens (commented):
Since MYSQLD_EXPORTER_PASSWORD is the only env var that can be expanded in the config, couldn't we just try to parse ${MYSQLD_EXPORTER_PASSWORD} in addition to what is already possible?

The current implementation looks a bit over-engineered

config/config.go
morawskim: Yes, I don't like this too.

I did this for backward-compatibility.
The latest mysqld_exporter is 0.20.0 so base on semantic versioning, the next version can introduce breaking, backward-incompatible changes. 
If for maintainers this is OK, for me too :).

IMHO we can assume if a first character is `$` , then we expand this env variable.
Otherwise we assume that this is password - no more $$ for escaping dollar sign.
I will adjust this PR in this week.
Thanks for feedback.

morawskim: If the password starts with `$`, it will be processed through `os.ExpandEnv`.

If the password does not start with `$`, it is treated as a literal string. Any `$` symbols within the password (e.g., `P@$$word`) will be preserved as-is and will not trigger environment variable expansion.

ArthurSens: > IMHO we can assume if a first character is $ , then we expand this env variable.

We can't make this assumption, I believe; passwords can start with `$`

ArthurSens:
I think I keep my opinion from the previous comment. We don't need to expand every Env variable, since `MYSQLD_EXPORTER_PASSWORD` is the only one we accept. Let's try to expand `$MYSQLD_EXPORTER_PASSWORD` and `${MYSQLD_EXPORTER_PASSWORD}` and that's it.

morawskim:
@ArthurSens 

I describe my use case, because I am afraid that `MYSQLD_EXPORTER_PASSWORD` is not only one which is accepted.
I am using one mysqld_exporter to monitor many independent mysqld servers.
If this is not correct way of using mysqld_exporter, I will update this PR to fit the project requirements.

Let's say I have docker-compose.yml

```
services:
  mysql1:
    image: mysql:8.0
    command: --default-authentication-plugin=mysql_native_password --character-set-server=utf8mb4 --collation-server=utf8mb4_unicode_ci
    environment:
      MYSQL_ROOT_PASSWORD: password
      MYSQL_DATABASE: dbname
      MYSQL_USER: dbname
      MYSQL_PASSWORD: userpassword
    ports:
      - "127.0.0.1:3307:3306"
  mysql2:
    image: mysql:5.7
    environment:
      MYSQL_ROOT_PASSWORD: password2
      MYSQL_DATABASE: dbname2
      MYSQL_USER: dbname2
      MYSQL_PASSWORD: userpassword2
    ports:
      - "127.0.0.1:3308:3306"
```

And I have config file `.my.cnf` for mysqld_exporter:

```
[client.server1]
user = root
password = $MYSQL1_ROOT_PASS
[client.server2]
user = root
password = $MYSQL2_ROOT_PASS
```

I start exporter - `MYSQL1_ROOT_PASS=password MYSQL2_ROOT_PASS=password2 ./mysqld_exporter`
When I run `curl '127.0.0.1:9104/probe?auth_module=client.server1&target=127.0.0.1%3A3307'` I get metrics for mysql1.
When I run `curl '127.0.0.1:9104/probe?auth_module=client.server2&target=127.0.0.1%3A3308'` I get metrics for mysql2.

IMHO at the moment mysqld_exporter allow for many variables not only MYSQLD_EXPORTER_PASSWORD.
In README there is mention about this variable, but [source code allow for more](https://github.com/prometheus/mysqld_exporter/blob/main/config/config.go#L147)

ArthurSens:
Oh, sorry about the confusion. I see your point. Ok, let's expand any env var.

The assumption that passwords don't start with `$` doesn't seem realistic, though, so we need to consider other approaches.

Maybe we could do something like
```go
cfg.ValueMapper = func(value string) string {
	return os.Expand(value, func(name string) string {
		if name == "$" {
			return "$"
		}
		return os.Getenv(name)
	})
}
```

This way:

- `$PASSWORD` → environment expansion
- `${PASSWORD}` → environment expansion
- `$$secret` → literal $secret
- `abc$$foo` → literal abc$foo

What do you think?

morawskim:
@ArthurSens 

For me - :heart_eyes: 
I adjusted PR.

ArthurSens (approved):
LGTM

PR #684, opened 2022-11-19
Change processlist query to support ONLY_FULL_GROUP_BY sql_mode

darxriggs:
Otherwise `Error 1055: 'information_schema.processlist.HOST' isn't in GROUP BY` is raised.

fixes #610 (for details see [#issuecomment-1320577266](https://github.com/prometheus/mysqld_exporter/issues/610#issuecomment-1320577266))

@SuperQ 
CC @peterloeffler

PR #1006, opened 2026-02-13
Resolve ambiguous column reference in processlist query

tanji:
Replace bare column names with COALESCE expressions in GROUP BY clause to match SELECT clause, eliminating MySQL warning 1052 for ambiguous 'state' column

### DIFF
--- a/collector/info_schema_processlist.go
+++ b/collector/info_schema_processlist.go
@@ -38,7 +38,7 @@ const infoSchemaProcesslistQuery = `
 		  FROM information_schema.processlist
 		  WHERE ID != connection_id()
 		    AND TIME >= %d
-		  GROUP BY user, host, command, state
+		  GROUP BY user, SUBSTRING_INDEX(host, ':', 1), COALESCE(command, ''), COALESCE(state, '')
 	`
 
 // Tunable flags.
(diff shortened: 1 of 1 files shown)


ArthurSens (changes requested):
This change would re-introduce the bug fixed in #684 

I don't know how, but it would be awesome if we had tests with MariaDB to make sure we don't re-introduce bugs like this 🤔

tanji:
@ArthurSens I found a solution if you are willing to review it. Wrapping the query into a subquery solves the issue and works with ONLY_FULL_GROUP_BY (tested both ways)

ArthurSens:
Awesome! I think the only thing we're missing is integration tests to verify that we're no longer seeing the warnings.

If you rebase your branch, you'll find a few integration tests already. They all start with:

```go
connDSN := os.Getenv("TEST_MYSQL_DSN")
if connDSN == "" {
	t.Skip("TEST_MYSQL_DSN is not set")
}
//more code
conn, err := instance.getDB().Conn(ctx)
if err != nil {
	t.Fatal(err)
}
defer conn.Close()
```

and for testing warnings I'm imagining something like

```go
	warnings, err := conn.QueryContext(ctx, "SHOW WARNINGS")
	if err != nil {
		t.Fatalf("show warnings: %v", err)
	}
	defer warnings.Close()
	for warnings.Next() {
		// assertions here
	}
```

I haven't tested this at all, so might need to refinements

ArthurSens:
A rebase also would not be a bad idea :)

PR #1015, opened 2026-03-27
chore: modernize WaitGroup usage

cristiangreco:
In this PR:
- use `wg.Go()` instead of `wg.Add(1)` and `defer wg.Done()`
- remove the `sortedMapKeys` function and replace it with modern `maps.Keys` and `slices.Sorted`
- remove loop variable capture in tests


SuperQ (approved):
Nice

PR #1025, opened 2026-05-05
feat: per-scraper query timeout to prevent slow scraper from cascading-failing the rest

AadiSanghani:
## Summary

  A single slow scraper currently causes every other scraper to fail on the same `/metrics` request. Two coupled changes fix this:

  1. **Per-scraper context timeout** — each scraper gets its own `context.WithTimeout(parent, --exporter.query_timeout)` instead of sharing the request context. A new `--exporter.query_timeout` flag controls
  the timeout. Default is `0` (disabled), so existing deployments see no behavior change.

  2. **Connection pool capacity** — `MaxOpenConns` raised from `1` to `2` and `ConnMaxLifetime` set to `1m`. With only one connection, scrapers serialize behind a slow query and timeout while waiting for the
  connection rather than while running, which the per-scraper timeout cannot catch on its own.

  The `instance.Ping()` call also takes a context now and uses `db.PingContext` so the ping respects the same timeout.

  ## Why both changes are needed

  Per-scraper timeout alone is not sufficient. With `MaxOpenConns(1)`, fast scrapers queue behind a slow one waiting for the connection, then hit their per-scraper deadline while still in the queue and fail
  with `context deadline exceeded`. Bumping the pool to 2 lets the fast scrapers run on a second connection while the slow one is in flight, so the per-scraper timeout fires only on the genuinely slow
  scraper.

  ## Reproduction

  A scraper that runs `SELECT SLEEP(15)` was added locally to reproduce the failure mode against MySQL 8.4 with a 5s scrape budget (`X-Prometheus-Scrape-Timeout-Seconds: 5`).

  | Configuration | Scrape time | Failed scrapers | Notes |
  |---|---|---|---|
  | `MaxOpenConns=1`, no per-scraper timeout (pre-fix) | 4.8s | 3 of 7 | slow + 2 queued behind it |
  | `MaxOpenConns=2` + `--exporter.query_timeout=3` (post-fix) | 3.0s | 1 of 7 | only the slow scraper |

  Pre-fix uses the entire scrape budget regardless of what's actually slow; post-fix uses exactly the per-scraper timeout that's configured.

Also full credit to @joshbroughton for this PR.

### DIFF
--- a/collector/exporter.go
+++ b/collector/exporter.go
@@ -74,6 +74,7 @@ type Exporter struct {
 	enableLockWaitTimeout bool
 	lockWaitTimeout       int
 	slowLogFilter         bool
+	queryTimeout          time.Duration
 }
 
 type ExporterOpt func(*Exporter)
@@ -93,6 +94,14 @@ func SetLockWaitTimeout(timeout int) ExporterOpt {
 func SetSlowLogFilter(b bool) ExporterOpt {
 	return func(e *Exporter) {
 		e.slowLogFilter = b
+	}
+}
+
+// SetQueryTimeout sets a per-scraper query timeout. Zero disables the timeout
+// and falls back to the parent (request) context.
+func SetQueryTimeout(timeout time.Duration) ExporterOpt {
+	return func(e *Exporter) {
+		e.queryTimeout = timeout
 	}
 }
 
@@ -157,7 +166,13 @@ func (e *Exporter) scrape(ctx context.Context, ch chan<- prometheus.Metric) floa
 	defer instance.Close()
 	e.instance = instance
 
-	if err := instance.Ping(); err != nil {
+	pingCtx := ctx
+	if e.queryTimeout > 0 {
+		var pingCancel context.CancelFunc
+		pingCtx, pingCancel = context.WithTimeout(ctx, e.queryTimeout)
+		defer pingCancel()
+	}
+	if err := instance.Ping(pingCtx); err != nil {
 		e.logger.Error("Error pinging mysqld", "err", err)
 		return 0.0
 	}
@@ -177,7 +192,13 @@ func (e *Exporter) scrape(ctx context.Context, ch chan<- prometheus.Metric) floa
 			label := "collect." + scraper.Name()
 			scrapeTime := time.Now()
 			collectorSuccess := 1.0
-			if err := scraper.Scrape(ctx, instance, ch, e.logger.With("scraper", scraper.Name())); err != nil {
+			scrapeCtx := ctx
+			if e.queryTimeout > 0 {
+				var cancel context.CancelFunc
+				scrapeCtx, cancel = context.WithTimeout(ctx, e.queryTimeout)
+				defer cancel()
+			}
+			if err := scraper.Scrape(scrapeCtx, instance, ch, e.logger.With("scraper", scraper.Name())); err != nil {
 				e.logger.Error("Error from scraper", "scraper", scraper.Name(), "target", e.getTargetFromDsn(), "err", err)
 				collectorSuccess = 0.0
 			}
--- a/collector/exporter.go
+++ b/collector/exporter.go
@@ -105,6 +105,16 @@ func SetQueryTimeout(timeout time.Duration) ExporterOpt {
 	}
 }
 
+// withQueryTimeout derives a context bounded by the configured query timeout.
+// When the timeout is disabled (0), it returns the parent context and a no-op
+// cancel so callers can unconditionally `defer cancel()`.
+func (e *Exporter) withQueryTimeout(ctx context.Context) (context.Context, context.CancelFunc) {
+	if e.queryTimeout > 0 {
+		return context.WithTimeout(ctx, e.queryTimeout)
+	}
+	return ctx, func() {}
+}
+
 // New returns a new MySQL exporter for the provided DSN.
 func New(ctx context.Context, dsn string, scrapers []Scraper, logger *slog.Logger, opts ...ExporterOpt) *Exporter {
 	e := &Exporter{
@@ -166,12 +176,8 @@ func (e *Exporter) scrape(ctx context.Context, ch chan<- prometheus.Metric) floa
 	defer instance.Close()
 	e.instance = instance
 
-	pingCtx := ctx
-	if e.queryTimeout > 0 {
-		var pingCancel context.CancelFunc
-		pingCtx, pingCancel = context.WithTimeout(ctx, e.queryTimeout)
-		defer pingCancel()
-	}
+	pingCtx, pingCancel := e.withQueryTimeout(ctx)
+	defer pingCancel()
 	if err := instance.Ping(pingCtx); err != nil {
 		e.logger.Error("Error pinging mysqld", "err", err)
 		return 0.0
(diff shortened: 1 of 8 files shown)


AadiSanghani:
cc: @SuperQ

collector/instance.go
```diff
@@ -41,8 +43,9 @@ func newInstance(dsn string) (*instance, error) {
 	if err != nil {
 		return nil, err
 	}
-	db.SetMaxOpenConns(1)
+	db.SetMaxOpenConns(2)
 	db.SetMaxIdleConns(1)
+	db.SetConnMaxLifetime(1 * time.Minute)
```
nicolastakashi: I'm wondering if these changes are related and required to the main idea behind this PR.

collector/exporter.go
```diff
@@ -157,7 +166,13 @@ func (e *Exporter) scrape(ctx context.Context, ch chan<- prometheus.Metric) floa
 	defer instance.Close()
 	e.instance = instance
 
-	if err := instance.Ping(); err != nil {
+	pingCtx := ctx
+	if e.queryTimeout > 0 {
```
nicolastakashi: I'm wondering if we can consolidate this check queryTimeout > 0 and thinking if we really need a context for ping and another for scrape, since they are using the same timeout value.

AadiSanghani: Fair! I moved them out to its method now.

collector/instance.go
AadiSanghani: Removed the `db.SetConnMaxLifetime(1 * time.Minute)` as its not needed for this PR.

nicolastakashi:
@AadiSanghani can you check DCO 🙏🏽

collector/exporter.go
```diff
@@ -96,6 +97,24 @@ func SetSlowLogFilter(b bool) ExporterOpt {
...
+func SetQueryTimeout(timeout time.Duration) ExporterOpt {
+	return func(e *Exporter) {
+		e.queryTimeout = timeout
+	}
+}
+
+// withQueryTimeout derives a context bounded by the configured query timeout.
+// When the timeout is disabled (0), it returns the parent context and a no-op
+// cancel so callers can unconditionally `defer cancel()`.
+func (e *Exporter) withQueryTimeout(ctx context.Context) (context.Context, context.CancelFunc) {
```
nicolastakashi: Maybe? 

```suggestion
func (e *Exporter) withQueryTimeoutContext(ctx context.Context) (context.Context, context.CancelFunc) {
```

nicolastakashi (commented):
LGTM, waiting final approval.

nicolastakashi (approved):

ArthurSens:
Well, do I understand correctly that if `MaxOpenConns` is now 2, 2 slow queries would be sufficient for the same behavior we had before with only 1 slow query?


If that's the case, would it make sense to also make `MaxOpenConns` configurable?

ArthurSens:
Something very very weird is happening in this PR too... CI did not run lol

Only the DCO check ran so far.

The current code doesn't compile, which would have been caught by CI.

collector/exporter_test.go
```diff
@@ -146,3 +148,59 @@ func TestExporterWithOpts(t *testing.T) {
...
+		return s.validate(ctx)
+	}
+	return nil
+}
+
+func TestScrapeContextTimeout(t *testing.T) {
+	connDSN := os.Getenv("TEST_MYSQL_DSN")
+	if connDSN == "" {
+		t.Skip("TEST_MYSQL_DSN is not set")
+	}
```
ArthurSens, line 173: I'm not sure if I'm missing something, but does that mean we currently don't run this test in our CI?

If that's true, could we find a way to test that a really slow query doesn't block other queries?

ArthurSens (commented):
I'm not sure if we have precedence in this repository (I'm not a maintainer of this exporter), but it would be nice to have integration tests that run a real mysql database (maybe testcontainers?).

I'd love to see a test with a real slow query and an assertion that it doesn't block a fast query

collector/exporter_test.go
SuperQ: We have an integration test workflow that populates this env var.

SuperQ:
I think the reason tests are missing is the fork branch is behind main and we recently migrated from CircleCI to GitHub Actions here.

@AadiSanghani would you please rebase your fork against main?

AadiSanghani:
>Well, do I understand correctly that if MaxOpenConns is now 2, 2 slow queries would be sufficient for the same behavior we had before with only 1 slow query?

>If that's the case, would it make sense to also make MaxOpenConns configurable?

@ArthurSens Yea I agree with you on this. I made it so that it defaults to 2, but is configurable.

AadiSanghani:
>I think the reason tests are missing is the fork branch is behind main and we recently migrated from CircleCI to GitHub Actions here.

>@AadiSanghani would you please rebase your fork against main?

@SuperQ I rebased our fork and this branch, still no CI. It does "3 workflows awaiting approval" so maybe once I get an approval it will run, not sure?

mysqld_exporter.go
```diff
@@ -256,6 +266,11 @@ func main() {
 	logger.Info("Starting mysqld_exporter", "version", version.Info())
 	logger.Info("Build context", "build_context", version.BuildContext())
 
+	if *exporterMaxOpenConns < 1 {
+		logger.Error("Invalid value for --exporter.max_open_connections, must be >= 1", "value", *exporterMaxOpenConns)
+		os.Exit(1)
+	}
```
ArthurSens: Could you add similar validation for query_timeout? It cannot be negative.

collector/instance.go
ArthurSens: we should use QueryRowContext here, right?

ArthurSens (commented):
I'm also missing a test that actually test the problem we want to avoid: a slow query won't block a sequential fast query.

I asked some LLM to write this for me and I got

```go
func TestSlowScraperDoesNotBlockFastScraper(t *testing.T) {
	connDSN := os.Getenv("TEST_MYSQL_DSN")
	if connDSN == "" {
		t.Skip("TEST_MYSQL_DSN is not set")
	}

	const queryTimeout = 2 * time.Second
	testCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
	defer cancel()

	slowReady := make(chan struct{})
	slowResult := make(chan error, 1)
	fastResult := make(chan error, 1)

	slowScraper := &mockScraper{
		name: "slow",
		validate: func(ctx context.Context, instance *instance) error {
			conn, err := instance.getDB().Conn(ctx)
			if err != nil {
				close(slowReady)
				slowResult <- err
				return err
			}
			defer conn.Close()

			// Signal only after reserving one of the pool's connections.
			close(slowReady)

			var result int
			err = conn.QueryRowContext(ctx, "SELECT SLEEP(10)").Scan(&result)
			slowResult <- err
			return err
		},
	}

	fastScraper := &mockScraper{
		name: "fast",
		validate: func(ctx context.Context, instance *instance) error {
			select {
			case <-slowReady:
			case <-ctx.Done():
				fastResult <- ctx.Err()
				return ctx.Err()
			}

			var result int
			err := instance.getDB().QueryRowContext(ctx, "SELECT 1").Scan(&result)
			fastResult <- err
			return err
		},
	}

	exporter := New(
		testCtx,
		connDSN,
		[]Scraper{slowScraper, fastScraper},
		promslog.NewNopLogger(),
		SetQueryTimeout(queryTimeout),
		SetMaxOpenConns(2),
	)

	ch := make(chan prometheus.Metric)
	go func() {
		exporter.Collect(ch)
		close(ch)
	}()
	for range ch {
	}

	select {
	case err := <-fastResult:
		if err != nil {
			t.Fatalf("fast query failed behind slow query: %v", err)
		}
	case <-testCtx.Done():
		t.Fatal("timed out waiting for fast query result")
	}

	select {
	case err := <-slowResult:
		if !errors.Is(err, context.DeadlineExceeded) {
			t.Fatalf("slow query error = %v, want context deadline exceeded", err)
		}
	case <-testCtx.Done():
		t.Fatal("timed out waiting for slow query result")
	}
}
```

From first glance the test looks correct to me, but feel free to double check and verify the correctness

mysqld_exporter.go
AadiSanghani: Added

collector/instance.go
AadiSanghani: Yea I think that makes sense. Updated.

ArthurSens:
@AadiSanghani , looks like you messed up your rebase 😅 There are commits that have already been merged to main in here

ArthurSens (approved):
Thanks!

PR #1027, opened 2026-05-25
Add TLS version config options

vyzigold:
Add tls-min-version and tls-max-version config options. These allow restrictions on which TLS version will be used for communicating with the database server. The values follow the same pattern as the mysql client, so "TLSv1.2, TLSv1.3, etc"

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -18,8 +18,10 @@ import (
 	"crypto/x509"
 	"fmt"
 	"log/slog"
+	"maps"
 	"net"
 	"os"
+	"slices"
 	"strconv"
 	"strings"
 	"sync"
@@ -55,6 +57,14 @@ var (
 	}
 
 	err error
+
+	// Use the same pattern of TLS version strings as the mysql client binary.
+	tlsVersions = map[string]uint16{
+		"TLSv1.0": tls.VersionTLS10,
+		"TLSv1.1": tls.VersionTLS11,
+		"TLSv1.2": tls.VersionTLS12,
+		"TLSv1.3": tls.VersionTLS13,
+	}
 )
 
 type Config struct {
@@ -73,6 +83,8 @@ type MySqlConfig struct {
 	SslKey                string `ini:"ssl-key"`
 	TlsInsecureSkipVerify bool   `ini:"ssl-skip-verfication"` //nolint:misspell
 	Tls                   string `ini:"tls"`
+	TlsMinVersion         string `ini:"tls-min-version"`
+	TlsMaxVersion         string `ini:"tls-max-version"`
 }
 
 type MySqlConfigHandler struct {
@@ -172,6 +184,19 @@ func (m MySqlConfig) validateConfig() error {
 		return fmt.Errorf("no user specified in section or parent")
 	}
 
+	allowedTLSVersions := strings.Join(slices.Sorted(maps.Keys(tlsVersions)), ", ")
+	if _, ok := tlsVersions[m.TlsMinVersion]; !ok && m.TlsMinVersion != "" {
+		return fmt.Errorf("tls-min-version=%s is not allowed, use one of: %s", m.TlsMinVersion, allowedTLSVersions)
+	}
+	if _, ok := tlsVersions[m.TlsMaxVersion]; !ok && m.TlsMaxVersion != "" {
+		return fmt.Errorf("tls-max-version=%s is not allowed, use one of: %s", m.TlsMaxVersion, allowedTLSVersions)
+	}
+	if m.TlsMinVersion != "" && m.TlsMaxVersion != "" {
+		if tlsVersions[m.TlsMinVersion] > tlsVersions[m.TlsMaxVersion] {
+			return fmt.Errorf("tls-min-version must not be higher than tls-max-version: %s > %s", m.TlsMinVersion, m.TlsMaxVersion)
+		}
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


vyzigold:
Original content of the CustomizeTLS is unchanged, it's just moved inside the `if m.SslCa != ""`  condition.

vyzigold:
@SuperQ @roidelapluie what do you think?

SuperQ:
Oh, I like the idea. To start @tjhop can give this a quick review.

config/config.go
```diff
@@ -55,6 +57,14 @@ var (
 	}
 
 	err error
+
+	// Use the same pattern of TLS version strings as the mysql client binary
```
tjhop: Prom code convention is comments should end in a full stop (period). This is something the linter should catch, but looks like our golangci-lint configs in this exporter repo are pretty minimal currently

config/config.go
```diff
@@ -172,6 +184,19 @@ func (m MySqlConfig) validateConfig() error {
...
+	}
+	if m.TlsMaxVersion != "" && !slices.Contains(allowedTLSVersions, m.TlsMaxVersion) {
+		return fmt.Errorf("tls-max-version=%s is not allowed, use one of: %s", m.TlsMaxVersion, strings.Join(allowedTLSVersions, ", "))
+	}
+	if m.TlsMinVersion != "" && m.TlsMaxVersion != "" {
+		if tlsVersions[m.TlsMinVersion] > tlsVersions[m.TlsMaxVersion] {
+			return fmt.Errorf("tls-min-version must not be higher than tls-max-version: %s > %s", m.TlsMinVersion, m.TlsMaxVersion)
+		}
+	}
+
```
tjhop: The logic here reads a bit convoluted to me.

I can understand the desire to build a list of versions from the map keys, but we just re-parse it for the conditionals and still duplicate the `strings.Join()` calls. 

Why not just check the map directly in the conditional for the existence check? Ie, `if minVer, ok := tlsVersions[m.TlsMinVersion]; !ok { ... }`

That'd also allow you to hoist the `strings.Join()` call to here:
`allowedTLSVersions := strings.Join(slices.Sorted(maps.Keys(tlsVersions)), ", ")`

config/config_test.go
```diff
@@ -291,5 +317,14 @@ func TestFormDSNWithCustomTls(t *testing.T) {
...
 
+		convey.Convey("Target tls custom", func() {
+			cfg := c.GetConfig()
+			section := cfg.Sections["client_tls_custom"]
+			if dsn, err = section.FormDSN(""); err != nil {
+				t.Error(err)
+			}
+			convey.So(dsn, convey.ShouldEqual, "usr:pwd@tcp(server3:3306)/?tls=custom")
+		})
+
```
tjhop: This only checks that we're using a custom TLS config, the core functionality we're adding is untested. 

Fortunately, it looks like the parsed DSN provides this for us in a structured way. Maybe something closer to this? 

```go
  parsed, _ := mysql.ParseDSN(dsn)
  convey.So(parsed.TLS.MinVersion, convey.ShouldEqual, uint16(tls.VersionTLS12))
  convey.So(parsed.TLS.MaxVersion, convey.ShouldEqual, uint16(tls.VersionTLS13))
```

config/config_test.go
```diff
@@ -169,6 +169,32 @@ func TestValidateConfig(t *testing.T) {
...
+	convey.Convey("Client with incorrect TLS version configuration", t, func() {
+		c := MySqlConfigHandler{
+			Config: &Config{},
+		}
+		os.Clearenv()
+		err := c.ReloadConfig("testdata/incorrect_tls_versions.cnf", "localhost:3306", "", true, promslog.NewNopLogger())
+		convey.So(
+			err,
+			convey.ShouldResemble,
+			fmt.Errorf("no configuration found"),
```
tjhop: Why use `ReloadConfig()` here? We're asserting on the `no config found` error, which isn't what we care about testing here.

We made changes to the `validateConfig()`, ideally that's what should be tested. It looks like this is just an extension of existing patterns, but the `ReloadConfig()` function logs validation errors and carries on:

https://github.com/prometheus/mysqld_exporter/pull/1027/changes#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17R165-R168

...so I'm pretty sure a good chunk of these tests are tautologically inaccurate (cc: @SuperQ, please tell me if I'm off base)

tjhop (changes requested):
Overall I like the idea, thanks! 

Requesting changes because I think the config loading can be simplified, and the tests need to be re-worked.

config/config.go
vyzigold: Done :+1:

config/config.go
vyzigold: I remember I had a few versions of this code, while trying to find the best set of condition while having some sort of nice error message. Your suggestion is good. Thank you!

config/config_test.go
vyzigold: I added the version tests on the parsed DSN as suggested. Thanks.

config/config_test.go
vyzigold: You're right, that this was an extension of existing patterns. I thought to extend the existing tests of similar functionality that were already there (the User validation) and keep the code consistent. But looking at it now, I agree, testing the validateConfig directly seems like a better idea.

tjhop (approved):
I think this look and reads much better now, and thanks very much for the test improvements. I'm on board!

SuperQ (approved):
Thanks! Also thanks @tjhop for the review help.

PR #1030, opened 2026-05-28
build: add distroless image variant

roidelapluie:
The distroless image runs as nonroot (UID 65532) without a shell, reducing the attack surface compared to the busybox-based image.


tjhop (approved):
Aligns with how prom does it, seems reasonable to me 👍

SuperQ (approved):

PR #1041, opened 2026-06-18
docs: fix typos in clientstats/userstats comments

s3onghyun:
Fixes "cient" -> "client" and "other then" -> "other than" in the info_schema collector comments.


SuperQ (approved):
Thanks!

PR #1043, opened 2026-06-29
Refactor mysqld_exporter to make it re-usable as a Go library

ArthurSens:
Refactors the mysqld_exporter, following the guidelines documented [here](https://github.com/prometheus/prometheus-opentelemetry-collector/blob/main/docs/embeddable-exporters.md). 

The goal is to facilitate the use case where mysql_exporter runs as part of other applications, using mysqld_exporter as a Go library.

### DIFF
--- a/collector/runtime.go
+++ b/collector/runtime.go
@@ -1,0 +1,119 @@
+// Copyright 2026 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package collector
+
+import (
+	"context"
+	"errors"
+	"log/slog"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/mysqld_exporter/config"
+)
+
+type Runtime struct {
+	exporter *Exporter
+}
+
+func NewRuntime(cfg *config.Config, logger *slog.Logger) (*Runtime, error) {
+	return NewRuntimeWithContext(context.Background(), cfg, logger)
+}
+
+func NewRuntimeWithContext(ctx context.Context, cfg *config.Config, logger *slog.Logger) (*Runtime, error) {
+	if cfg == nil {
+		return nil, errors.New("config is required")
+	}
+	if !cfg.Validated() {
+		return nil, errors.New("config has not been validated; call cfg.Validate before NewRuntime")
+	}
+	if ctx == nil {
+		ctx = context.Background()
+	}
+	if logger == nil {
+		logger = slog.Default()
+	}
+
+	return &Runtime{
+		exporter: New(
+			ctx,
+			cfg.DataSourceName,
+			EnabledScrapers(*cfg),
+			logger,
+			EnableLockWaitTimeout(cfg.EnableExporterLockWaitTimeout),
+			SetLockWaitTimeout(cfg.ExporterLockWaitTimeout),
+			SetSlowLogFilter(cfg.SlowLogFilter),
+		),
+	}, nil
+}
+
+func (r *Runtime) Collectors() []prometheus.Collector {
+	if r == nil || r.exporter == nil {
+		return nil
+	}
+	return []prometheus.Collector{r.exporter}
+}
+
+func EnabledScrapers(cfg config.Config) []Scraper {
+	scrapers := AllScrapers(cfg)
+	enabled := make([]Scraper, 0, len(scrapers))
+	for _, scraper := range scrapers {
+		if cfg.Collectors[scraper.Name()] {
+			enabled = append(enabled, scraper)
+		}
+	}
+	return enabled
+}
+
+func AllScrapers(cfg config.Config) []Scraper {
+	emptyConfig := config.EmptyConfig{}
+
+	return []Scraper{
+		ScrapeGlobalStatus(emptyConfig),
+		ScrapeGlobalVariables(emptyConfig),
+		ScrapeSlaveStatus(emptyConfig),
+		ScrapeProcesslist(cfg.InfoSchemaProcesslist),
+		ScrapeUser(cfg.MysqlUser),
+		ScrapeTableSchema(cfg.InfoSchemaTables),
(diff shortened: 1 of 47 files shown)


collector/runtime.go
```diff
@@ -0,0 +1,119 @@
...
+func (r *Runtime) Collectors() []prometheus.Collector {
+	if r == nil || r.exporter == nil {
+		return nil
+	}
+	return []prometheus.Collector{r.exporter}
+}
+
+func EnabledScrapers(cfg config.Config) []Scraper {
+	scrapers := AllScrapers(cfg)
+	enabled := make([]Scraper, 0, len(scrapers))
```
macabu, line 79: Nit: I don't think we can pre-allocate here because the size of enabled depends on `if cfg.Collectors[scraper.Name()] {`

collector/runtime.go
```diff
@@ -0,0 +1,119 @@
...
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/mysqld_exporter/config"
+)
+
+type Runtime struct {
+	exporter *Exporter
+}
+
+func NewRuntime(cfg *config.Config, logger *slog.Logger) (*Runtime, error) {
```
macabu: Nit: Is this ever going to be called outside of tests? If not, for tests you could use `t.Context()` for example.

collector/runtime.go
```diff
@@ -0,0 +1,119 @@
...
+		return nil, errors.New("config is required")
+	}
+	if !cfg.Validated() {
+		return nil, errors.New("config has not been validated; call cfg.Validate before NewRuntime")
+	}
+	if ctx == nil {
+		ctx = context.Background()
+	}
+	if logger == nil {
+		logger = slog.Default()
```
macabu: Nit: Is `slog` the default logger we use throughout? I'm unsure if these scenarios of `ctx == nil` or `logger == nil` ever practically happen, and if they ever do, wouldn't we want to error out instead? Letting the caller pick the context and logger implementation could document things better than overriding it in the implementation details of `Runtime`?

config/config.go
```diff
@@ -57,7 +46,219 @@ var (
 	err error
 )
 
+const (
+	DefaultTimeoutOffset             = 0.25
+	DefaultExporterLockWaitTimeout   = 2
```
macabu: Nit on these: could they have a unit: minute, seconds, ms?

config/config.go
```diff
@@ -75,27 +276,47 @@ type MySqlConfig struct {
...
+			Name:      "config_last_reload_successful",
+			Help:      "Mysqld exporter config loaded successfully.",
+		}),
+		configReloadSeconds: prometheus.NewGauge(prometheus.GaugeOpts{
+			Namespace: "mysqld_exporter",
+			Name:      "config_last_reload_success_timestamp_seconds",
+			Help:      "Timestamp of the last successful configuration reload.",
+		}),
+	}
+	registerer.MustRegister(ch.configReloadSuccess, ch.configReloadSeconds)
```
macabu: Nit: since we return errors here couldn't we use `registerer.Register`?

config/config.go
```diff
@@ -57,7 +46,219 @@ var (
...
+		if _, ok := DefaultCollectorConfig()[name]; !ok {
+			return fmt.Errorf("unknown collector %q", name)
+		}
+	}
+
+	c.validated = true
+	return nil
+}
+
+func (c Config) Validated() bool {
```
macabu: Big nit, not for here, but I have been experimenting with the idea of "Parse don't validate" since Go has been getting better generic support, with a pattern that looks like:
```go
type Unparsed struct{}
type Parsed struct{}

type Config[P Unparsed | Parsed] struct {
	DataSourceName string
}

func (c *Config[Unparsed]) Parse() (*Config[Parsed], error) {
	if c.DataSourceName == "" {
		return nil, fmt.Errorf("data source name must not be empty")
	}
	return (*Config[Parsed])(c), nil
}
```
That way you wouldn't need to check if the config has been validated or not in runtime, even if the config can change during runtime, you can re-parse it.

mysqld_exporter.go
```diff
@@ -167,11 +185,69 @@ func getScrapeTimeoutSeconds(r *http.Request, offset float64) (float64, error) {
 	return timeoutSeconds, nil
 }
 
+func configForCollectParams(cfg config.Config, collectParams []string) config.Config {
```
macabu, line 198: I wonder if the collector config should be something separate `config.CollectorConfig`/`config.RuntimeConfig` than the global `config.Config`, maybe I misunderstood, but only the collector config can be changed through HTTP via runtime right?

macabu (commented):
Obviously not familiar with the codebase so take everything with a big grain of salt 😃

config/config.go
ArthurSens: Good point, fixed!

config/config.go
ArthurSens: For this part specifically I'm trying to stick with what is documented at https://github.com/prometheus/prometheus-opentelemetry-collector/blob/main/docs/embeddable-exporters.md

If we follow a standard, then all exporter's codebase would look pretty much the same and it's a lot easier to maintain. I would love to better understand this idea though, if this is something we'd like to pursue we probably want to update the guidance doc.

collector/runtime.go
ArthurSens: I'm not sure I understand... this is used by downstream go applications that use Mysqld_exporter as a library 🤔

collector/runtime.go
ArthurSens: I think the pre allocation to the maximum possible is still good. This means we don't need to allocate everytime we add something to the array, it saves a couple of CPU cycles :)

collector/runtime.go
ArthurSens: fixed!

config/config.go
ArthurSens: good point as well, I've renamed the variables to include the unit

mysqld_exporter.go
ArthurSens: Hmmmm, I think you caught something here...

If we're running mysqld_exporter as a library, it shouldn't carry HTTP specific configurations... I think we need to split those 🤔

ArthurSens: I moved it out of collector config

ArthurSens:
Thanks for the review @macabu , I believe I've addressed most of the comments!

collector/runtime.go
macabu: Oh I just meant this specific constructor without an explicit `ctx` :) - but it's fine!

config/config.go
macabu: Got it, I agree with following the standard, wasn't aware that this was part of it, so no worries!

macabu (approved):
Awesome!

ArthurSens:
@SuperQ, wanna take a look? 

Things looks mostly ready to be merged :)

PR #1044, opened 2026-07-01
Add a tls-server-name config option and fix TLS config race

vyzigold:
This allows specifying a hostname for TLS verification of the server certificates when accessing mysql server through its IP. This is the same as the server_name config option used in prometheus configuration.

Also a pre-existing race condition in TLS configuration was discovered during review. This would heavily impact the new tls-server-name config option. So the race condition is fixed as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Changes:
 * [BUGFIX]
 
 * [FEATURE] Add tls-min-version and tls-max-version config options
+* [FEATURE] Add an optional tls-server-name config option used to verify the server hostname
 
 ## 0.19.0 / 2026-03-18
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Changes:
 
 * [FEATURE] Add tls-min-version and tls-max-version config options
 * [FEATURE] Add an optional tls-server-name config option used to verify the server hostname
+* [BUGFIX] Fix a race condition due to which targets could use custom CA from different config sections
 
 ## 0.19.0 / 2026-03-18
 
--- a/config/config.go
+++ b/config/config.go
@@ -85,6 +85,7 @@ type MySqlConfig struct {
 	Tls                   string `ini:"tls"`
 	TlsMinVersion         string `ini:"tls-min-version"`
 	TlsMaxVersion         string `ini:"tls-max-version"`
+	TlsServerName         string `ini:"tls-server-name"`
 }
 
 type MySqlConfigHandler struct {
@@ -234,7 +235,13 @@ func (m MySqlConfig) FormDSN(target string) (string, error) {
 		config.TLSConfig = "skip-verify"
 	} else {
 		config.TLSConfig = m.Tls
-		if m.SslCa != "" || m.TlsMinVersion != "" || m.TlsMaxVersion != "" {
+
+		hasCustomTLS := m.SslCa != "" ||
+			m.TlsMinVersion != "" ||
+			m.TlsMaxVersion != "" ||
+			m.TlsServerName != ""
+
+		if hasCustomTLS {
 			if err := m.CustomizeTLS(); err != nil {
 				err = fmt.Errorf("failed to register a custom TLS configuration for mysql dsn: %w", err)
 				return "", err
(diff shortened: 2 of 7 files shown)


tjhop:
I'd suggest parking this until after #1027 -- it's touching the same hunks and there will be conflicts

SuperQ (dismissed):
Needs a rebase.

config/config.go
```diff
@@ -234,7 +235,13 @@ func (m MySqlConfig) FormDSN(target string) (string, error) {
...
 	} else {
 		config.TLSConfig = m.Tls
-		if m.SslCa != "" || m.TlsMinVersion != "" || m.TlsMaxVersion != "" {
+
+		hasCustomTLS := m.SslCa != "" ||
+			m.TlsMinVersion != "" ||
+			m.TlsMaxVersion != "" ||
+			m.TlsServerName != ""
+
+		if hasCustomTLS {
```
tjhop, line 244: given that this is only used once and immediately after declaration, I'm not sure it really improves readability all that much. I'd be fine with the 4 inline alternations, personally.

tjhop (approved):
👍

config/config.go
vyzigold: Having it on one line was getting quite long to my liking especially if this would ever be extended further. When leaving the alternations inside the if, but making the condition multiline as it is right now, gofmt would indent the later lines to the same level as the current L245, which was quite unreadable on first look.

This was the solution that looked the most readable and possibly extensible to me. I can of course change it to whatever you'd prefer.

tjhop: It's functionally fine, I won't block on this

vyzigold:
Just rebased on top of latest main.

config/config.go
ArthurSens: I'm not sure about this, but it feels like this could have some racing conditions in the multi-target pattern...?

I can imagine multiple probes with different auth_module sections all writing to the same place

ArthurSens (commented):
It might be worth updating README, more specifically the section `## Customizing Configuration for a SSL Connection`.

config/config.go
```diff
@@ -280,6 +287,9 @@ func (m MySqlConfig) CustomizeTLS() error {
 	if m.TlsMaxVersion != "" {
 		tlsCfg.MaxVersion = tlsVersions[m.TlsMaxVersion]
 	}
+	if m.TlsServerName != "" {
+		tlsCfg.ServerName = m.TlsServerName
+	}
 	tlsCfg.InsecureSkipVerify = m.TlsInsecureSkipVerify
```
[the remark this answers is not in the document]
vyzigold: While this is a good find, I don't agree with this comment being made on this PR. Here is why:

- This is a pre-existing issue. Even without this PR, when a config has the TLS version config (added by me recently) or ssl-ca (which is there for some while already), it will hit the same issue.
- A config, where the user configures a whole bunch of TLS options and then sets tls=skip-verify doesn't make much sense and it may make sense to even completely reject it.

I can add the suggestion if the reviewers wish, it's not a big deal, but I'd personally rather stay focused on the problem at hand, rather than solving unrelated, pre-existing issues.

config/config.go
vyzigold: I agree.

While this is a pre-existing issue, which could deserve to be addressed separately (the same race would occur with ssl-ca and the tls versions I think), it seems pretty crucial to me to have it fixed for the server name to work correctly. Because while the same race occurs with the other options, it probably doesn't matter for them in most cases (I assume the CA, client cert/key and TLS versions will very likely work even across config sections), this can't be said about server name, which I assume will always be different for each section, so the race would be really noticable.

I think I see how this could be fixed. We can propagate the config section name here and into CustomizeTLS and then use the `<config_section_name>` instead of `custom`. I'll try to get to it tomorrow.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ Changes:
 * [BUGFIX]
 
 * [FEATURE] Add tls-min-version and tls-max-version config options
+* [FEATURE] Add an optional tls-server-name config option used to verify the server hostname
```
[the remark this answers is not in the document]
vyzigold: done

config/config.go
vyzigold: See the second commit of the PR for the fix.

ArthurSens (approved):
Thanks!!

PR #1064, opened 2026-07-22
fix: preserve permission errors in slave_hosts scraper

### LINKED ISSUE #970: Using --collect.slave_hosts with MariaDB and user with inadequate permissions results in a Syntax Errors instead of a Permission Denied error.
Using --collect.slave_hosts with MariaDB which does not yet support `SHOW REPLICAS` and a user who does not have the required grants to run the  `SHOW SLAVE HOSTS` command causes a syntax error to be logged. 

This seems to originate from commit f6a64d768c6d0e182ab70733c07f6d8781d4fa0c (https://github.com/prometheus/mysqld_exporter/pull/837) which added support for `SHOW REPLICAS` results but the implementation assumes that the first error is due to the old syntax not being supported and overwrites the error with the result of `SHOW REPLICAS` which will result in a syntax error for everything other than MySQL 8.4 for users with incorrect grants.

## Steps to reproduce

1. Install any versio of MariaDB or MySQL 8.0
2. Create a user without the [REPLICATION SLAVE](https://dev.mysql.com/doc/refman/8.4/en/privileges-provided.html#priv_replication-slave) privilege
3. Run mysqld_exporter with --collect.slave_hosts

The result should be that a syntax error is logged instead of the correct error that shows the user doesn't have the correct grants.

locker95:
## Summary

When `--collect.slave_hosts` ran against MariaDB (or MySQL without `SHOW REPLICAS`) with a user missing the required grants, the scraper always fell back from `SHOW SLAVE HOSTS` to `SHOW REPLICAS` on *any* error. The secondary statement then failed with a syntax error and masked the real permission-denied failure.

Only fall back to `SHOW REPLICAS` when `SHOW SLAVE HOSTS` fails with MySQL error 1064 (parse/syntax error), which is the case on MySQL 8.4+ where that statement was removed. Permission and other errors are returned unchanged.

## Changes

- `collector/slave_hosts.go`: gate the `SHOW REPLICAS` fallback on MySQL error number 1064
- `collector/slave_hosts_test.go`: cover permission-denied (no fallback) and syntax-error (fallback succeeds)

Fixes #970

## Test plan

- [x] `go test ./collector/ -run TestScrapeSlaveHosts`
- [ ] Reproduce with MariaDB + user without `REPLICATION SLAVE`: log should show access denied, not a syntax error from `SHOW REPLICAS`
- [ ] Confirm MySQL 8.4 still scrapes via `SHOW REPLICAS` fallback

### DIFF
--- a/collector/slave_hosts.go
+++ b/collector/slave_hosts.go
@@ -20,6 +20,7 @@ import (
 	"database/sql"
 	"log/slog"
 
+	"github.com/go-sql-driver/mysql"
 	"github.com/google/uuid"
 	"github.com/prometheus/client_golang/prometheus"
 )
@@ -69,8 +70,16 @@ func (ScrapeSlaveHosts) Scrape(ctx context.Context, instance *instance, ch chan<
 		err            error
 	)
 	db := instance.getDB()
-	// Try the both syntax for MySQL 8.0 and MySQL 8.4
+	// Try SHOW SLAVE HOSTS first (MySQL < 8.4 / MariaDB). On MySQL 8.4+ that
+	// statement was removed, so fall back to SHOW REPLICAS only when the server
+	// reports a syntax/parse error. Other failures (e.g. permission denied)
+	// must be returned as-is so they are not masked by a secondary syntax error
+	// from SHOW REPLICAS on servers that do not support it.
 	if slaveHostsRows, err = db.QueryContext(ctx, slaveHostsQuery); err != nil {
+		mysqlErr, ok := err.(*mysql.MySQLError)
+		if !ok || mysqlErr.Number != 1064 {
+			return err
+		}
 		if slaveHostsRows, err = db.QueryContext(ctx, showReplicasQuery); err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


collector/slave_hosts.go
```diff
@@ -69,8 +70,16 @@ func (ScrapeSlaveHosts) Scrape(ctx context.Context, instance *instance, ch chan<
...
 	db := instance.getDB()
-	// Try the both syntax for MySQL 8.0 and MySQL 8.4
+	// Try SHOW SLAVE HOSTS first (MySQL < 8.4 / MariaDB). On MySQL 8.4+ that
+	// statement was removed, so fall back to SHOW REPLICAS only when the server
+	// reports a syntax/parse error. Other failures (e.g. permission denied)
+	// must be returned as-is so they are not masked by a secondary syntax error
+	// from SHOW REPLICAS on servers that do not support it.
 	if slaveHostsRows, err = db.QueryContext(ctx, slaveHostsQuery); err != nil {
+		mysqlErr, ok := err.(*mysql.MySQLError)
+		if !ok || mysqlErr.Number != 1064 {
```
ArthurSens, line 82: Could we add a link to the documentation for error 1064?

locker95:
added a link to the MySQL ER_PARSE_ERROR (1064) docs next to the fallback.

ArthurSens (commented):
Thank you!

ArthurSens (approved):

PR #1065, opened 2026-07-22
feat: make max open DB connections configurable

### LINKED ISSUE #1037: Configurable number of database connections SetMaxOpenConns
Hello!

This is more of a question rather than an issue. I see that `SetMaxOpenConns` is hardcoded to 1. Is there a specific reason why it is not configurable?

My understanding is that collectors fan-out but they all share the same database connection. A single slow collector will block the others, right?  

I saw that there was a issue and PR created many years ago https://github.com/prometheus/mysqld_exporter/issues/330 but it also involved some more changes.

locker95:
## Summary

`database/sql.DB.SetMaxOpenConns` was hardcoded to `1`, so all scrapers in a scrape shared a single connection. A slow collector can block the others even though they fan out concurrently.

Add a CLI flag to configure the pool size while keeping the historical default of 1.

## Changes

- New flag: `--exporter.max_open_conns` (default `1`)
- Plumb value through `collector.SetMaxOpenConns` into `newInstance`
- Apply the option from both `/metrics` and `/probe` handlers
- Document the flag in `README.md`

Fixes #1037

## Test plan

- [x] `go test ./...`
- [ ] Default (no flag): behavior unchanged, single open connection
- [ ] `--exporter.max_open_conns=N` with multiple collectors: concurrent scrapes can open up to N connections

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,7 @@ log.level                                  | Logging verbosity (default: info)
 exporter.lock_wait_timeout                 | Set a lock_wait_timeout (in seconds) on the connection to avoid long metadata locking. (default: 2)
 exporter.enable_lock_wait_timeout          | Enable the lock_wait_timeout connection parameter. Makes the exporter compatible with older versions of MySQL. (default: true)
 exporter.log_slow_filter                   | Add a log_slow_filter to avoid slow query logging of scrapes.  NOTE: Not supported by Oracle MySQL.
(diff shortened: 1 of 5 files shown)


ArthurSens:
This one will be solved by https://github.com/prometheus/mysqld_exporter/pull/1025, so I'm closing. I hope that's ok :)

PR #1068, opened 2026-08-03
Add OpenContainer image labels to Dockerfile

### LINKED ISSUE #979: Add opencontainer labels to docker image
Tools like renovate, dependabot and socket will use [opencontainer](https://specs.opencontainers.org/image-spec/annotations/#pre-defined-annotation-keys) docker labels to inline changelogs in PRs/comments when updating

Othertools probably use the same standard labels

locker95:
Other exporters (blackbox etc.) already ship OCI labels; mysqld_exporter was still just the old maintainer label.

Added the usual org.opencontainers.image.* set so renovate/dependabot and friends can pick up source/docs metadata.

Fixes #979

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,9 +3,9 @@ ARG OS="linux"
 FROM quay.io/prometheus/busybox-${OS}-${ARCH}:latest
 LABEL maintainer="The Prometheus Authors <prometheus-developers@googlegroups.com>"
 
-LABEL org.opencontainers.image.authors="The Prometheus Authors <prometheus-developers@googlegroups.com>"
+LABEL org.opencontainers.image.authors="The Prometheus Authors"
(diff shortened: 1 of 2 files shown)


ArthurSens (changes requested):
Looks good, but it seems like you're adding different values for a few metadata labels that are already present in `Dockerfile.distroless`. Could we make this the exact same as the distroless one, just changing the `io.prometheus.image.variant` label?

ArthurSens (commented):
Actually, `Apache-2.0` is the OCI/SPDX-compliant license value, so we might need to update this one in the distroless image instead

locker95:
aligned the busybox Dockerfile labels with distroless (only variant differs), and switched distroless to SPDX Apache-2.0 as well.

ArthurSens (approved):
Thanks!

PR #1082, opened 2026-09-09
Add AGENTS.md

ArthurSens:
After reviewing more than 40 PRs in the last month, I'm encoding the not-so-obvious things I had to look up the most into `AGENTS.md` for proper PR reviews.

I hope this file helps both new reviewers, and contributors making contributions through LLMs.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,57 @@
+# Contributor Guidance for AI Agents
+
+Keep changes focused and validate claims with repository evidence. Read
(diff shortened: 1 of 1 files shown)


SuperQ (approved):
I still would like to see some kind of sync / template of best practices from prometheus/prometheus so we don't have to hand-maintain this across all repos.

ArthurSens:
Yeah, I'm not sure if Prometheus/prometheus is the best place for this since its exporter specific, but I agree that some stuff like naming conventions and collectors package details could be maintained in a central place

